PR #2, opened 2020-05-08
Fix k8s middleware to deduplicate lines when metadata is absent

CJP10:
When a metadata lookup returns None deduplication was not occuring.


bin/Cargo.toml
```diff
@@ -10,7 +10,7 @@
 
 [package]
 name = "logdna-agent"
-version = "2.1.8"
+version = "2.1.9-rc1"
```
jakedipity, line 13: We've decided to stick with beta as the format for prerelease builds, omitting the number on the first beta. So this should be `2.1.9-beta`

CJP10: :+1:

PR #4, opened 2020-05-20
Ensure paths that resolve to event are valid for the configured watch window

jakedipity:
This change places a stricter requirement for resolved paths for an event to fall inside of the initial watched directory. This stops the issue where Kubernetes environments have duplicate lines for multiple files due to the usage of symlinks.

This change also excludes `/var/log/pods/**/*` so the only valid path to get logs for pods will be at `/var/log/containers` fixing an issue where the symlink at `/var/log/pods/` would start generating logs before the symlink at `/var/log/containers` was created thus leading to missing metadata information (found in the `/var/log/containers/` path name).

Since we've eliminated duplicate logs we can axe the deduplication logic we were previously using.


CJP10:
@jakedipity I'm not sure if this will work. Excluding `/var/log/pods/**/*` would make the backing dirs unreachable, in this case (`/var/lib/docker/...` and `/var/data/kubeletlogs/...`).

jakedipity:
@CJP10 We have three functions for testing path validity

1. `is_initial_dir_target` - tests whether a path falls within a initial directory
2. `is_symlink_target` - tests whether a path is protected because a symlink points to it
3. `passes` - tests for `is_initial_dir_target || is_symlink_target`.

For all entry adding we use `passes`. The change introduces a `is_initial_dir_target` on all paths resolved for events (so it doesn't affect entries that are created). This should mean we won't interfere with exclusion rules blocking symlinks.

CJP10:
Master rules get applied on both `is_initial_dir_target ` and `is_symlink_target`.
Putting `/var/log/pods/**/*` in the master rules makes `passes` always return false.
That will short-circuit `fs::insert` (https://github.com/logdna/logdna-agent-v2/blob/watch-window/common/fs/src/cache/mod.rs#L380).
`/var/log/pods` is the only entry into the backing dirs since `/var/log/containers` doesn't link to the backing dirs but just `/var/log/pods`.

jakedipity:
`is_symlink_target` only checks the inclusion rules of the master rules, not the exclusion rules. However I think we should remove any dependency on master rules from that function, I just can't remember why we added that stipulation originally.

But my understanding (and testing verifies) is that this will work for what we want. Anything in `/var/log/pods` that follows the inclusion rule (`*.log,!(*.*)`) should not breaking inserting of entries, but will be excluded from the resolved paths

jakedipity:
@CJP10 We're going to move ahead since the path validation seemingly works as expected. Path validation works for `/var/log/pods/**/*` because of the symlinks created in `/var/log/containers`. This doesn't invalidate our exclusion rules because the resolved paths will remove `/var/log/pods/**/*` when we generate a log event

PR #5, opened 2020-05-28
Separate namespace into it's own yaml and create a sane installation flow

jakedipity:
Limiting ourselves to trying to install the new agent with just two steps causes a janky installation because of the introduction of the `logdna-agent` namespace. To make sure the secret gets created in the namespace, we moved it to be the second step, but this causes the agent to be created without a valid secret which leads to pods in an error state. In some environments this is kind of okay and the pods will start once the secret is created, in other environments the pods need to be restarted - in all environments this is kind of janky.

This change fixes that by moving the namespace into its own yaml and making that the first step in installation.


docs/README.md
```diff
@@ -66,8 +66,9 @@ The agent is compatible with Kubernetes clusters running `v1.9` or greater. It c
 
 
 ```
-kubectl create -f https://raw.githubusercontent.com/logdna/logdna-agent-v2/master/k8s/logdna-agent.yaml
+kubectl create -f https://raw.githubusercontent.com/logdna/logdna-agent-v2/master/k8s/logdna-agent-ns.yaml
 kubectl create secret generic logdna-agent-key -n logdna-agent --from-literal=logdna-agent-key=<YOUR LOGDNA INGESTION KEY>
```
smusali, line 70: is `-n` in this command line an option for `namespace`? Like the shortened version of `--namespace` and basically, for creating and adding a `secret` into `logdna-agent` namespace, right?

smusali:
What about regular yamls inside the `k8s` folder? Aren't you gonna remove or replace them with the newer ones you created in this PR?

docs/README.md
jakedipity: Yup, `--namespace` is the long form for `-n`.

jakedipity:
> What about regular yamls inside the `k8s` folder? Aren't you gonna remove or replace them with the newer ones you created in this PR?

By regular yamls I assume you mean `logdna-agent.yaml` and `logdna-agent-st.yaml`. Our in app documentation won't be updated for a while (probably next week or 2 weeks from now at the latest) so we need to make sure the old documentation still works as expected. Once our in app documentation is updated across all environments we can remove those yamls.

PR #6, opened 2020-06-05
[LOG-5131] Switch the base image in the docker file from ubuntu to Red Hat's UBI

jakedipity:
Internal Ref: LOG-5131

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ COPY . .
 RUN cargo build --release && strip ./target/release/logdna-agent
 
 # Use ubuntu as the final base image
-FROM ubuntu:18.04
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2
 
 ARG REPO
 ARG BUILD_TIMESTAMP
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -44,16 +44,12 @@ RUN cargo build --release
...
 
 # Use ubuntu as the final base image
-FROM ubuntu:18.04
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2
 
 ENV DEBIAN_FRONTEND=noninteractive
 ENV _RJEM_MALLOC_CONF="narenas:1,tcache:false,dirty_decay_ms:0,muzzy_decay_ms:0"
 ENV JEMALLOC_SYS_WITH_MALLOC_CONF="narenas:1,tcache:false,dirty_decay_ms:0,muzzy_decay_ms:0"
 
-RUN apt update -y && \
```
gregswift, line 53: You'll want to replace this with
```
RUN microdnf update -y && rm -rf /var/cache/yum
```

jakedipity: Added

jakedipity:
Bump @gregswift

gregswift (dismissed):

mdeltito (commented):
lgtm, will approve once commit reference is updated

jakedipity:
@mdeltito Still not 100% sure how I want to handle private tickets on a public repo like this. Created a new issue which tracks against the internal ticket without going into the business / private details and referenced the issue in the commit and PR.

Any thoughts on that?

mdeltito:
@jakedipity yeah good question. IMO the PR description should be written to provide enough context for external folks to understand what/why changes were made without a separate issue on GH to track it. The PR description really serves the same purpose from the perspective of someone following the repo.

GH issues to me are for external use - reporting bugs or feature requests - which we then bring internally into our systems as-needed (rather than the reverse). For our own tracking purposes tacking the `ref` onto the commit is enough to link it internally to those systems.

jakedipity:
@mdeltito Updated the commit and PR description to reference the internal ticket.

I think keeping our ticketing system internal and private will limit the ability for contributors to really engage with the long-term planning and discussion of agent features / goals.

Don't really need to figure it out now, but would like to start an internal discussion on it.

PR #11, opened 2020-07-13
Setup the Makefile to rely on docker provided Rust and update CI to use this

gregswift:
* using docker to execute the cargo steps reduces local version management of rust 
* Remove make calls from in Dockerfile (Makefile wasnt being copied in anyways) to avoid inception
* Add docker image metadata


Dockerfile
```diff
@@ -40,12 +40,30 @@ RUN grep -aL "github.com" target/release/deps/* | xargs rm \
...
+ARG VCS_REF
+ARG VCS_URL
+ARG BUILD_VERSION
+
+LABEL org.label-schema.schema-version="1.0"
+LABEL org.label-schema.url="https://logdna.com"
+LABEL org.label-schema.build-date="${BUILD_DATE}"
+LABEL org.label-schema.maintainer="LogDNA <support@logdna.com>"
+LABEL org.label-schema.name="answerbook/${REPO}"
+LABEL org.label-schema.description="LogDNA Agent V2"
```
jakedipity, line 60: Can we use the description we use on GitHub: `The blazingly fast, resource efficient log collection client`

Dockerfile
```diff
@@ -40,12 +40,30 @@ RUN grep -aL "github.com" target/release/deps/* | xargs rm \
...
+ARG BUILD_DATE
+ARG VCS_REF
+ARG VCS_URL
+ARG BUILD_VERSION
+
+LABEL org.label-schema.schema-version="1.0"
+LABEL org.label-schema.url="https://logdna.com"
+LABEL org.label-schema.build-date="${BUILD_DATE}"
+LABEL org.label-schema.maintainer="LogDNA <support@logdna.com>"
+LABEL org.label-schema.name="answerbook/${REPO}"
```
jakedipity, line 59: For the name can we use `LogDNA Agent`

Makefile
```diff
@@ -8,6 +18,17 @@ else
 	CARGO_COMPILE_OPTS = --release
 endif
 
+BUILD_DATE=$(shell date -u +'%Y%m%dT%H%M%SZ')
```
jakedipity, line 21: We should modify the format string to be RFC 3339 as defined in the label schema: `%Y-%m-%dT%H:%M:%SZ`

Makefile
```diff
@@ -33,8 +54,32 @@ test-deps:	## Install dependencies needed for the test target
...
+	$(DOCKER) build . -t $(REPO):$(BUILD_TAG)) \
+		--pull --no-cache=true \
+		--build-arg BUILD_DATE=$(BUILD_DATE) \
+		--build-arg BUILD_VERSION=$(BUILD_VERSION) \
+		--build-arg REPO=$(REPO) \
+		--build-arg VCS_REF=$(VCS_REF) \
+		--build-arg VCS_URL=$(VCS_URL)
+	$(DOCKER) tag $(REPO):$(BUILD_TAG) $(REPO):$(BUILD_VERSION)
+
+.PHONY:publish
```
jakedipity, line 70: I believe the `.PHONY` needs to be `publish-build`

Makefile
```diff
@@ -33,8 +54,32 @@ test-deps:	## Install dependencies needed for the test target
...
+publish-build:    ## Publish a build version of the docker image to our private registry
+	for version in $(BUILD_VERSION) $(VCS_BRANCH); do \
+		$(DOCKER) tag $(REPO):$(BUILD_TAG) $(DOCKER_PRIVATE_IMAGE):$${version}; \
+		$(DOCKER) push $${image}:$${version}; \
+	done
 
-.PHONY:docker
-docker:		## Build a docker image as specified in the Dockerfile
-	$(DOCKER) build .
+.PHONY:publish
```
jakedipity, line 77: Same issue with `.PHONY`

Makefile
```diff
@@ -33,8 +54,32 @@ test-deps:	## Install dependencies needed for the test target
...
+		--pull --no-cache=true \
+		--build-arg BUILD_DATE=$(BUILD_DATE) \
+		--build-arg BUILD_VERSION=$(BUILD_VERSION) \
+		--build-arg REPO=$(REPO) \
+		--build-arg VCS_REF=$(VCS_REF) \
+		--build-arg VCS_URL=$(VCS_URL)
+	$(DOCKER) tag $(REPO):$(BUILD_TAG) $(REPO):$(BUILD_VERSION)
+
+.PHONY:publish
+publish-build:    ## Publish a build version of the docker image to our private registry
```
jakedipity, line 71: This target should depend on `build-image`

Makefile
```diff
@@ -33,8 +54,32 @@ test-deps:	## Install dependencies needed for the test target
...
+	for version in $(BUILD_VERSION) $(VCS_BRANCH); do \
+		$(DOCKER) tag $(REPO):$(BUILD_TAG) $(DOCKER_PRIVATE_IMAGE):$${version}; \
+		$(DOCKER) push $${image}:$${version}; \
+	done
 
-.PHONY:docker
-docker:		## Build a docker image as specified in the Dockerfile
-	$(DOCKER) build .
+.PHONY:publish
+publish-release:    ## Publish SemVer compliant releases and 'latest' tag
```
jakedipity, line 78: This target should depend on `build-image`

Makefile
```diff
@@ -33,8 +40,25 @@ test-deps:	## Install dependencies needed for the test target
 
 .PHONY:help
 help:		## Prints out a helpful description of each possible target
-	@awk 'BEGIN {FS = ":.*?## "}; /^.+: .*?## / && !/awk/ {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' ${MAKEFILE_LIST}
+	@awk 'BEGIN {FS = ":.*?## "}; /^.+: .*?## / && !/awk/ {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+
+.PHONY:build-image
+BUILD_DATE=$(shell date -u +'%Y%m%dT%H%M%SZ')
```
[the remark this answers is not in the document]
gregswift: yeah, that would be great..

Dockerfile
```diff
@@ -40,12 +40,30 @@ RUN grep -aL "github.com" target/release/deps/* | xargs rm \
...
 # Use ubuntu as the final base image
 FROM ubuntu:18.04
 
+ARG REPO
+ARG BUILD_DATE
+ARG VCS_REF
+ARG VCS_URL
+ARG BUILD_VERSION
+
+LABEL org.label-schema.schema-version="1.0"
```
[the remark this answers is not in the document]
gregswift: i learned of that change after copying this from our older images.. 100% for the change

Dockerfile
gregswift: sure

Dockerfile
gregswift: :+1:

gregswift:
> I don't think we should migrate our open source projects to Jenkins. It isn't publicly accessible. For prospective contributors, this would cut them out of a big part of the feedback loop.

I'm definitely a fan of keeping our public repo's feedback public to our customers.  We should definitely firm up a plan around this.

jakedipity:
@esatterwhite @gregswift Just adding this here to document it since I believe you're both already aware: we're going to look into making the jenkin build steps and output publicly available so it won't be hidden from contributors.

Makefile
gregswift: @jakedipity so i'm playing with this over in the tooling-elasticsearch repository.

RFC3339 doesnt work inside a docker tag because of the colons, and on the next few line we make the build tag using this.

What I've landed on is:

```
BUILD_DATE=$(shell date -u "'%Y-%m-%d")
BUILD_TIME=$(shell date -u "%H:%M:%SZ")
BUILD_CREATED=$(BUILD_DATE)T$(BUILD_TIME)
BUILD_TAG=$(BUILD_DATE).$(BUILD_VERSION)
```
And then using `BUILD_CREATED` in the Dockerfile, which aligns it with teh new OCI tag better.

PR #12, opened 2020-07-18
WIP: Expand on the original jenkins-docker PR.

jakedipity:
* Rectify some of the issues originally outlines
* Separate build makefile from docker makefile


Dockerfile
```diff
@@ -40,29 +46,32 @@ RUN grep -aL "github.com" target/release/deps/* | xargs rm \
...
-LABEL org.label-schema.build-date="${BUILD_DATE}"
-LABEL org.label-schema.maintainer="LogDNA <support@logdna.com>"
-LABEL org.label-schema.name="answerbook/${REPO}"
-LABEL org.label-schema.description="LogDNA Agent V2"
-LABEL org.label-schema.vcs-url="${VCS_URL}"
-LABEL org.label-schema.vcs-ref="${VCS_REF}"
-LABEL org.label-schema.vendor="LogDNA Inc."
-LABEL org.label-schema.version="${BUILD_VERSION}"
-LABEL org.label-schema.docker.cmd="docker run logdna-agent-v2"
+LABEL com.logdna.stage="${REPO}-final"
```
jakedipity, line 63: This is also used for identifying final images for the same purpose of removal.

Dockerfile
```diff
@@ -1,5 +1,12 @@
...
-FROM rust:1.41 as build
+ARG BUILD_IMAGE
+
+FROM ${BUILD_IMAGE} as build
+
+ARG REPO
+ARG VCS_REF
+
+LABEL com.logdna.stage="${REPO}-build"
+LABEL com.logdna.tag="${VCS_REF}"
```
jakedipity, line 9: This label allows the cleaning to be more specific so we can safely have concurrent jenkins builds

Makefile.docker
```diff
@@ -0,0 +1,95 @@
+REPO := logdna-agent-v2
```
jakedipity, line 1: Separation of concerns, I moved the docker related stuff into it's own Makefile

Makefile.docker
```diff
@@ -0,0 +1,95 @@
...
+IMAGE_REPO ?= docker.io/rust
+IMAGE_TAG ?= 1.42
+IMAGE ?= $(IMAGE_REPO):$(IMAGE_TAG)
+IMAGE := $(IMAGE)
+
+VCS_REF := $(shell git rev-parse --short HEAD)
+VCS_URL := https://github.com/logdna/logdna-agent-v2
+BUILD_TIMESTAMP := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+BUILD_VERSION := $(shell sed -nE "s/^version = \"(.+)\"\$$/\1/p" bin/Cargo.toml)
+BUILD_TAG := $(VCS_REF)
```
jakedipity, line 19: I opted to use only the VCS_REF as the build tag instead of the ref+timestamp.

There's edge cases with a timestamp when the Makefile is called on multiple times where each call might get a different timestamp. The image should also have a timestamp metadata baked into it if we need it.

Makefile.docker
```diff
@@ -0,0 +1,95 @@
...
+# `make -f Makefile.docker build IMAGE=docker.io/rust:1.42.0`
+IMAGE_REPO ?= docker.io/rust
+IMAGE_TAG ?= 1.42
+IMAGE ?= $(IMAGE_REPO):$(IMAGE_TAG)
+IMAGE := $(IMAGE)
+
+VCS_REF := $(shell git rev-parse --short HEAD)
+VCS_URL := https://github.com/logdna/logdna-agent-v2
+BUILD_TIMESTAMP := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+BUILD_VERSION := $(shell sed -nE "s/^version = \"(.+)\"\$$/\1/p" bin/Cargo.toml)
```
jakedipity, line 18: We're grabbing the version as it is defined within the code as opposed to what Git tags might say.

Makefile.docker
```diff
@@ -0,0 +1,95 @@
...
+
+# The image repo and tag can be modified e.g.
+# `make -f Makefile.docker build IMAGE=docker.io/rust:1.42.0`
+IMAGE_REPO ?= docker.io/rust
+IMAGE_TAG ?= 1.42
+IMAGE ?= $(IMAGE_REPO):$(IMAGE_TAG)
+IMAGE := $(IMAGE)
+
+VCS_REF := $(shell git rev-parse --short HEAD)
+VCS_URL := https://github.com/logdna/logdna-agent-v2
```
jakedipity, line 16: Wanted to use a web friendly link instead of the github link which in my cases was a git ssh link

Makefile.docker
```diff
@@ -0,0 +1,95 @@
...
+	done;
+
+.PHONY:publish-public
+publish-public: ## Publish SemVer compliant releases to our public registries
+	@# TODO: Have a boolean that prevents this unless forced or run by Jenkins (which can force it)
+	for image in $(DOCKER_PUBLIC_IMAGE) $(DOCKER_IBM_IMAGE); do \
+		for version in $(MAJOR_VERSION) $(MINOR_VERSION) $(PATCH_VERSION) $(BUILD_VERSION) latest; do \
+			if [[ $${version} != $(BUILD_VERSION) ]] && [[ $(PRERELEASE_VERSION) = beta* ]]; then \
+				continue; \
+			fi; \
```
jakedipity, line 87: If the version is a beta, we won't update the major, minor, or patch versions of the images. E.g. for `2.1.9-beta.2`, the only thing published to the public is the tag `2.1.9-beta.2`

common/fs/src/cache/mod.rs
```diff
@@ -502,9 +502,7 @@ impl<T: Default> FileSystem<T> {
         entries.retain(|other| *other != entry_ptr);
         if entries.is_empty() {
             watch_descriptors.remove(&wd);
-            match self.watcher.unwatch(wd) {
-                _ => {}
-            }; // TODO: Handle this error case
+            let _ = self.watcher.unwatch(wd); // TODO: Handle this error case
```
jakedipity, line 505: These are fixes for issues raised in the new tests

common/fs/src/cache/mod.rs
```diff
@@ -607,7 +605,7 @@ impl<T: Default> FileSystem<T> {
                 .unwrap()
         };
 
-        entry.set_parent(new_parent.clone());
+        entry.set_parent(new_parent);
```
jakedipity, line 608: Ditto for above

Jenkinsfile
```diff
@@ -1,42 +1,107 @@
...
+                anyOf {
+                    expression {
+                        return changesMatch('^Makefile$')
                     }
-                }
-                stage('Build Agent') {
-                    steps {
-                        sh 'make build'
+                    expression {
+                        return changesMatch('^rust-image/')
```
gregswift, line 27: What are you trying to accomplish here?

Makefile.docker
gregswift: You shouldnt need to buiuld i nthe extra logic. the pipeline in jenkins should be configured to push it only if its being promoted.  such as [here](https://github.com/answerbook/tooling-elasticsearch/blob/master/Jenkinsfile#L17-L38)

Makefile.docker
```diff
@@ -0,0 +1,128 @@
...
+publish-rust: ## Publish superset rust image (see target rust-image)
+	$(DOCKER) tag $(RUST_REPO):$(BUILD_TAG) docker.pkg.github.com/answerbook/docker-images/$(RUST_REPO):latest
+	$(DOCKER) tag $(RUST_REPO):$(BUILD_TAG) us.gcr.io/logdna-k8s/$(RUST_REPO):latest
+	$(DOCKER) push docker.pkg.github.com/answerbook/docker-images/$(RUST_REPO):latest
+	$(DOCKER) push us.gcr.io/logdna-k8s/$(RUST_REPO):latest
+
+.PHONY:clean-rust-image
+clean-rust-image:
+	$(call CLEAN_DOCKER_IMAGES,--filter label=com.logdna.stage=$(RUST_REPO)-build $(CLEAN_ADDTL_FILTERS))
+	$(call CLEAN_DOCKER_IMAGES,--filter label=com.logdna.stage=$(RUST_REPO)-build $(CLEAN_ADDTL_FILTERS))
```
gregswift, line 124: If we are going to maintain a base rust image, this should really be external to this repository in its own.

rust-image/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+FROM ${BUILD_IMAGE}-slim as build
+
+ARG REPO
+ARG VCS_REF
+
+LABEL com.logdna.stage="${REPO}-build"
+LABEL com.logdna.tag="${VCS_REF}"
+
+ARG REPO
+ARG VCS_REF
```
gregswift, line 12: This is duplicative

Makefile.docker
jakedipity: I was on the fence about how to handle this. I was originally going to add the logic for building the logdna-agent rust base image to https://github.com/answerbook/docker-images.

I decided to go with the current approach for the following reasons:
1. The base image is tightly coupled with the agent itself; breaking them apart doesn't remove the coupling but instead hides it
2. Kills some of the automagic. Changes on build or test dependencies would need the base image to be updated which would need to be manually done since it wouldn't be able to detect changes from a different repository.

The current approach does the following when a Jenkins build goes for the logdna-agent-v2:
1. Check for changes to the base image
1.a If there are changes build a new base image and use the local version for future steps
1.b Otherwise, use the remote image for future steps
2. Test / build the agent with whatever base image assigned previously
3. If successful and there's a new base image, push that image to the remote repo

It does admittedly make the Jenkins build script a bit more complicated. Thoughts?

Jenkinsfile
jakedipity: I explained the base image build process in a different comment, but this is to detect changes for the files used to build the rust base image. We only want to build a local version if it's changed. Additionally we want to see the entire difference for the entire branch/PR hence the use of `changesMatch` instead of Jenkins `changeSet`

Dockerfile
```diff
@@ -1,5 +1,10 @@
-# select build image
-FROM rust:1.41 as build
+ARG BUILD_IMAGE
+
+FROM ${BUILD_IMAGE} as build
+
+ARG REPO
+
+LABEL com.logdna.stage="${REPO}-build"
```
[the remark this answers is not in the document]
jakedipity: Also, you can shorten `docker images $(REPO):* --format "{{.ID}}"` to just `docker images $(REPO):* -q`

Makefile.docker
```diff
@@ -0,0 +1,95 @@
...
+	CLEAN_ADDTL_FILTERS :=
+else
+	CLEAN_ADDTL_FILTERS := --filter label=com.logdna.tag=$(BUILD_TAG)
+endif
+
+CLEAN_DOCKER_IMAGES = if [[ ! -z "docker images -q --filter $(1) $(CLEAN_ADDTL_FILTERS)" ]]; then docker images -q --filter $(1) $(CLEAN_ADDTL_FILTERS) | head -n 1 | xargs -I {} docker rmi -f {}; fi
+
+.PHONY:build
+build: ## (Runs in a docker container) Build the agent. Set RELEASE=1 to build a release image - defaults to 0
+	$(DOCKER_RUN) /bin/sh -c "make build RELEASE=$(RELEASE) && $(CHOWN)"
```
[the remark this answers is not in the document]
gregswift: > I don't think it makes sense to make Docker a hard requirement for building this project e.g. the Makefile supports targets for building locally. 

So the point here was to actually make it so docker was the only requirement, and do this across all of our repos.  It actually decreases the potential for variance and should facilitate variable types of builds more readily.

If you don't want to follow that build process, in any of the repos, the normal language build steps like `cargo build` would still work.

Makefile.docker
gregswift: So my first question is... why are we building this rust base image ?

If the answer is that we absolutely have to, then i do think it should be in a separate repository. i would do a repository specifically for rust, because the problem you are describing is explicitly an issue we have to resolve regardless.  Over in `docker-images` we'll need to be updating these images regularly for security reasons, and triggering builds on things that haven't gotten a recent release (really it will just be blanket) so that we have updates available.

Increasing the complexity in this repository either way, to me, seems like the wrong route.

jakedipity: I removed this and moved it to our private repo docker-images

Makefile.docker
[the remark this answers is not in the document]
jakedipity: So I settled on moving the cargo build system into a system made for cargo, cargo-make. This solves the recursive issue and keeps the Makefile for docker building and the Makefile.toml (cargo make) for cargo building.

PR #16, opened 2020-08-27
Makes logdna-agent-v2 compatible with LogDNA's Jenkins pipeline

jakedipity:
This change is to add CI/CD to logdna-agent-v2. It also includes a number of fixes that have been found due to the tightened up lint testing


docker/entrypoint.sh
```diff
@@ -0,0 +1,31 @@
...
+	    groupadd -g "$2" logdna
+	fi
+
+	if case "$USER" in *"no such user") true ;; *) false ;; esac; then
+	    USER=logdna
+	    useradd -g "$GROUP" -u "$1" -m logdna
+	fi
+
+	chown -R "$USER":"$GROUP" "$4"
+	runuser -u "$USER" -g "$GROUP" -- /bin/sh -c "$5"
```
gregswift, line 26: Why are we doing this at runtime of the image?

docs/README.md
```diff
@@ -88,7 +88,7 @@ The compiled binary will be built to `./target/release/logdna-agent`.
 To build a Docker image of the agent, ensure docker is installed properly, verify the docker engine is running, and then run the following command:
 
 ```
-docker build .
+make docker-build-image
```
gregswift, line 91: this would be `make build-image` now, ya?

Makefile
```diff
@@ -0,0 +1,194 @@
...
+	$(DOCKER_RUN) $(RUST_IMAGE) $(ENTRYPOINT) "cargo test"
+
+.PHONY:clean
+clean: ## Clean all artifacts from the build process
+	$(DOCKER_RUN) $(RUST_IMAGE) $(ENTRYPOINT) "cargo clean"
+
+.PHONY:clean-docker
+clean-docker: ## Cleans the intermediate and final agent images left over from the build-image target
+	@# Clean any agent images, left over from the multi-stage build
+	$(call CLEAN_DOCKER_IMAGES,$(REPO):$(CLEAN_TAG))
```
gregswift, line 75: this really could just be

```
	-docker images -q $(1) | xargs docker rmi -f
```

Makefile
```diff
@@ -0,0 +1,194 @@
...
+	PULL_OPTS :=
+endif
+
+CHANGE_VERSION = awk '{sub(/^version = ".+"$$/, "version = \"$(1)\"")}1' bin/Cargo.toml >> bin/Cargo.toml.tmp && mv bin/Cargo.toml.tmp bin/Cargo.toml
+
+REMOTE_BRANCH := $(shell git branch -vv | awk '/^\*/{split(substr($$4, 2, length($$4)-2), arr, "/"); print arr[2]}')
+
+.PHONY:build
+build: ## Build the agent
+	$(DOCKER_RUN) $(RUST_IMAGE) $(ENTRYPOINT) "cargo build"
```
gregswift, line 58: All of these could be simplfied into a single variable up top and then only that variable used everywhere.

```
CARGO_COMMAND := $(DOCKER_RUN) $(RUST_IMAGE) $(ENTRYPOINT)
...
	$(CARGO_COMMAND) "cargo build"
```

docker/entrypoint.sh
gregswift: I see what you are doing.. i'm still at a loss as to why you need to do this :/

jakedipity: I am trying to ensure two things:
1. The guest machine should have permissions to modify existing files; needed for `make clean`.
2. The host machine has correct ownership of files created by the guest machine; byproduct of `make build`, `make test`, and `make lint`.

Docker on the Mac platform gives us this ability out of the box:
1. The guest machine runs as root
2. Despite writing as root on the guest machine, Mac's filesystem correctly writes the files on the host machine with the permissions of the calling user/group.

Linux; however, does not. The filesystem will copy the permissions from the guest machine to the host machine so if the guest machine is running as root then the files are created as root on the host machine. To get around this, we need to run the guest machine as the user on the guest machine.

We can't simply pass the user into the docker run command because Docker always mounts the host directory as `1000:1000` so if the host machine user/group does not share this id it won't have access. The entrypoint script allows us to start the machine as root, chown the mounted directory to the correct user/group, and then run as that user/group with `runuser`.

Finally `runuser` won't work with ids so we need to ensure a user and group exists with the passed in ids. If they don't we need to create them.

This is ultimately to allow commands like `make build`, `make clean`, and `make test` to work both for developers (Mac & Linux) and our CI/CD system (Linux).

docs/README.md
jakedipity: Good catch

docker/entrypoint.sh
esatterwhite: I'm pretty sure it mounts the directory with the permissions the directory on the host machine is created as.
if the first step is `rm -r target && mkdir -p target` I think the directory permissions will be preserved.

This all might be a bit of over kill
I personally would just run `cargo build` 🤷 or  chown the artifact this myself anyway. 
But maybe you don't need a container for this. can you just use a shell script on the host and chown it as the user?

If the goal is to provide an artifact so people don't have to build it, could we just host pre-built binaries on jenkins or github?

esatterwhite: Second question - Does this build actually work on Mac? Can you build a rust binary in a linux container and Have it run on a mac laptop?

jakedipity: @esatterwhite Ah while that seems to be true for Linux, it's not for Mac. We could change the script to run on the host and issue the correct docker run command:
```console
# Mac:
docker run --rm -w /build -v /Users/jake/logdna/logdna-agent-v2:/build:Z docker.io/rust:1.42 make build

# Linux:
docker run --rm -u $(id -u):$(id -g) -w /build -v /Users/jake/logdna/logdna-agent-v2:/build:Z docker.io/rust:1.42 make build
```

The purpose for using Docker, afaik, is to allow anyone to come along and build/test/lint the agent without needing to install cargo. This also ensures everyone is running on the same version and on the same platform to remove any inconsistencies between environments.

The build artifacts (the final binary as well as the compilation units) are, afaik, not compatible with Mac. The agent can't be built against Mac either because it relies on `inotify` which Mac doesn't provide. The benefits of keeping the build artifacts (namely the compilation units) is the ability to re use them for future builds/tests/lints.

Makefile
jakedipity: I can move the `CLEAN_DOCKER_IMAGES` statement down to this target and drop the `$(call ...)` syntax, but I think I still need to keep the
```bash
if [[ ! -z "$(shell docker images -q $(1))" ]]; then
```

because on my Linux box it complains when the image is not there. The behavior is different on Mac (seems like the pipe silently exits on no data?) . I suppose it's a matter of taste, but I wouldn't want the clean command to fail because there's nothing to clean.

.gitignore
```diff
@@ -1,4 +1,7 @@
 /target
+/common/*/target
+/bin/target
 **/*.rs.bk
 /.idea
-.idea/
\ No newline at end of file
+.idea/
```
gregswift, line 6: this 2nd one should cover line 5 i believe

Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+        ansiColor 'xterm'
+    }
+    triggers {
+        cron(env.BRANCH_NAME ==~ /\d\.\d/ ? 'H H 1,15 * *' : '')
+    }
+    stages {
+        stage('Test') {
+            steps {
+                sh """
+                    make lint RUST_IMAGE_REPO=${RUST_IMAGE_REPO}
```
gregswift, line 18: i'm not sure why we dont just make this the default in the makefile and avoid this?  Why is the makefile's image different?

Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+                                    sh '''
+                                        echo "${DOCKERHUB_ACCESS_TOKEN}" | docker login -u logdna --password-stdin
+                                        echo "${ICR_IAMAPIKEY}" | docker login -u iamapikey --password-stdin icr.io
+                                        make publish
+                                    '''
+                                }
+                            }
+                            post {
+                                always {
+                                    sh '''
```
gregswift, line 56: This is interesting.  Why are we logging out ?

Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+                    }
+                    stages {
+                        stage('Publish Images') {
+                            input {
+                                message "Should we publish the versioned image?"
+                                ok "Publish image"
+                            }
+                            steps {
+                                withCredentials([string(credentialsId: 'icr-iamapikey', variable: 'ICR_IAMAPIKEY'), string(credentialsId: 'dockerhub-access-token', variable: 'DOCKERHUB_ACCESS_TOKEN')]) {
+                                    sh '''
```
gregswift, line 47: we really should build this into the jenkins instance like our other creds, so that this kind of thing doesnt have to exist in multiple places

Jenkinsfile
jakedipity: This is still a public repository and it's a bit awkward if the default tooling doesn't work. Ideally we could make the rust image publicly available.

Jenkinsfile
jakedipity: Trying to limit access to the registries for only when we need it.

Jenkinsfile
jakedipity: Same comment as above. I would imagine we want to only be logged in when we absolutely need to be to minimize any attack surfaces.

If we want to have the Jenkins slaves to always be logged in to the registries, I can look into it.

Makefile
gregswift: The starting `-` prevents it from failing

Jenkinsfile
gregswift: ah yeah.. where did all of that end up ? i'm with you, would be good to make it public, or find some way to just use the public image

Jenkinsfile
gregswift: Its not really minimizing the attack surfaces much. every running job has access to the credentials.

Jenkinsfile
jakedipity: Moved to this repo: https://github.com/answerbook/docker-images

__Note__: The rust Dockerfile isn't quite up to date with the needs with the latest changes, I needed to tweak some things. I'm holding off on making another PR for that repo just in case there's more tweaking that needs to be done.

Jenkinsfile
jakedipity: I opted to replace this with `docker.withRegistry`

docker/entrypoint.sh
esatterwhite: Rustup solves the version probably doesn't it?
you might want to look at the [notify](https://github.com/notify-rs/notify/tree/v4.0.15#notify) crate.

common/middleware/src/lib.rs
```diff
@@ -50,9 +51,10 @@ impl Executor {
             }
         }
 
-        match skipped {
-            true => None,
-            false => Some(lines),
+        if skipped {
+            None
+        } else {
+            Some(lines)
```
esatterwhite, line 57: Nit pickey, but rust encourages making these kinds of things look like ternaries but collapsing the curlies

```rust
if  skipped { None } else { Some(lines) }
```

esatterwhite (dismissed):

common/middleware/src/lib.rs
jakedipity: `rustfmt` seems to prefer it on separate lines instead of on just one:
```git
Diff in /build/common/middleware/src/lib.rs at line 51:
             }
         }

-        if skipped { None } else { Some(lines) }
+        if skipped {
+            None
+        } else {
+            Some(lines)
+        }
     }
 }
```

Tested on `1.42` and `1.46`

esatterwhite (approved):

gregswift (approved):
Docker, Jenkins, and Make lgtm

PR #18, opened 2020-09-23
Use agent binary's CARGO_PKG/VERSION in useragent

c-nixon:
Previously the user agent was built using the CARGO_PKG and CARGO_PKG_VERSION for the common/config module which was statically included during compilation.

This PR changes the behaviour so that the main.rs statically includes it's CARGO_PKG and CARGO_PKG_VERSION which the config module references as an extern &'static str.

Before this change the user-agent string showed as:
`config/0.1.0 (Red Hat Enterprise Linux/8.2 (Ootpa))`

it now appears as:
`logdna-agent/2.1.9-beta.2 (Red Hat Enterprise Linux/8.2 (Ootpa))`

Internal Ref: LOG-7367

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -23,6 +23,15 @@ static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
 static SLEEP_DURATION: Duration = Duration::from_millis(10);
 
+// Statically include the CARGO_PKG_NAME and CARGO_PKG_VERSIONs in the binary
+// and export under the PKG_NAME and PKG_VERSION symbols.
+// These are used to identify the application and version, for example as part
+// of the user agent string.
+#[no_mangle]
+pub static PKG_NAME: &str = env!("CARGO_PKG_NAME");
+#[no_mangle]
+pub static PKG_VERSION: &str = env!("CARGO_PKG_VERSION");
+
 #[cfg(use_systemd)]
 fn register_journald_source(source_reader: &mut SourceReader) {
     source_reader.register(JournaldSource::new());
--- a/common/config/src/lib.rs
+++ b/common/config/src/lib.rs
@@ -21,6 +21,12 @@ pub mod env;
 pub mod error;
 pub mod raw;
 
+// Symbols that will be populated in the main.rs file
+extern "Rust" {
+    static PKG_NAME: &'static str;
+    static PKG_VERSION: &'static str;
+}
+
 #[derive(Debug)]
 pub struct Config {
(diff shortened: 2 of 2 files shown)


bin/src/main.rs
```diff
@@ -23,6 +23,11 @@ static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
 static SLEEP_DURATION: Duration = Duration::from_millis(10);
 
+#[no_mangle]
+pub static PKG_NAME: &str = env!("CARGO_PKG_NAME");
+#[no_mangle]
+pub static PKG_VERSION: &str = env!("CARGO_PKG_VERSION");
+
```
jakedipity, line 30: Let's add a comment to explain what this is doing

common/config/src/lib.rs
```diff
@@ -21,6 +21,11 @@ pub mod env;
 pub mod error;
 pub mod raw;
 
+extern "Rust" {
+    static PKG_NAME: &'static str;
+    static PKG_VERSION: &'static str;
+}
```
jakedipity, line 27: More comments

common/config/src/lib.rs
```diff
@@ -137,12 +142,10 @@ impl TryFrom<RawConfig> for Config {
...
         }
 
-        template_builder.user_agent(format!(
-            "{}/{} ({})",
-            env!("CARGO_PKG_NAME"),
-            env!("CARGO_PKG_VERSION"),
-            info
-        ));
+        // Safety: this is reading from extern "Rust" statics declared in main.rs or the test module
+        let (pkg_name, pkg_version) = unsafe { (PKG_NAME, PKG_VERSION) };
```
jakedipity, line 146: Can we update the comment to indicate how we know this line of code is ultimately safe?

If I understand correctly this will only panic if the main binary is missing these unmangled variables. Is it possible to provide a default for that case?

jakedipity (changes requested):

common/config/src/lib.rs
c-nixon: There won't be any panics, if the symbol is not defined then the linker will error out before the build completes.

The only error case would be if the symbols are exported but have the wrong type.
Given that we are the ones exporting the symbols, and any duplicate defined elsewhere in the dependecy graph will result in a linker error that shouldn't be an issue.

jakedipity (approved):

jakedipity (changes requested):
Let's rebase this into a single commit

jakedipity (approved):

PR #19, opened 2020-09-24
Add k8s labels

c-nixon:
This adds the app.kubernetes.io/{name,version,instance} labels to the kubernetes yaml and extends the existing version bump functionality in the makefile to update the version when a release is made.

### DIFF
--- a/k8s/agent-resources-supertenant.yaml
+++ b/k8s/agent-resources-supertenant.yaml
@@ -3,6 +3,10 @@ kind: DaemonSet
 metadata:
   name: logdna-agent
   namespace: logdna-agent
+  labels:
+    app.kubernetes.io/name: logdna-agent
+    app.kubernetes.io/instance: logdna-agent
+    app.kubernetes.io/version: 2.1.9-beta.2
 spec:
   updateStrategy:
     type: RollingUpdate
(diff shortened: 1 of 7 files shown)


k8s/agent-resources-supertenant.yaml
```diff
@@ -15,6 +19,9 @@ spec:
     metadata:
       labels:
         app: logdna-agent
```
mdeltito, line 21: any reason to leave this in the DaemonSet definitions?

c-nixon: Only that it's used in the matchLabels (which could easily be updates) and that someone's tooling might be expecting it. But yeh, I believe "app" has been deprecated in favour of app.kubernetes.io/name. I'm not opposed to removing it if you think the tidiness is worth it

mdeltito: we should clean it up at some point IMO, but removing labels might need to be considered a major/breaking change for the reason you mentioned. I'm fine with it as-is here, just stood out in these changes.

mdeltito:
@c-nixon can we rebase + squash these into one commit before this gets merged?

k8s/agent-resources-supertenant.yaml
c-nixon: I've added a ticket to the brand new rust/native agent v3 epic

mdeltito (approved):

jakedipity (dismissed):

jakedipity (approved):

PR #20, opened 2020-09-24
Reorder kubernetes yamls with dependencies first

c-nixon:
Reorders the k8s yamls so that items depended on by other items are created before their dependants.

### DIFF
--- a/k8s/logdna-agent.yaml
+++ b/k8s/logdna-agent.yaml
@@ -8,6 +8,81 @@ metadata:
     app.kubernetes.io/instance: logdna-agent
     app.kubernetes.io/version: 2.1.9-beta.2
 ---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: logdna-agent
+  namespace: logdna-agent
+  labels:
+    app.kubernetes.io/name: logdna-agent
+    app.kubernetes.io/instance: logdna-agent
+    app.kubernetes.io/version: 2.1.9-beta.2
+---
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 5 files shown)


k8s/logdna-agent.yaml
```diff
@@ -8,6 +8,81 @@ metadata:
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: logdna-agent
+subjects:
+  - kind: ServiceAccount
+    name: logdna-agent
+    namespace: logdna-agent
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
```
mdeltito, line 70: should this be after `ClusterRole`? here and in the others.

mdeltito (commented):
one question but the yaml looks sane

k8s/logdna-agent.yaml
c-nixon: Looks like it refers to the clusterrole on line 25, it's in a different order from the other files, which is odd... It was that way before though, and the role is still defined before the binding so I think it's correct

mdeltito: yeah seems fine as-is with the dependency order, I was thinking more in terms of organization keeping them together. no strong feelings here though

mdeltito (dismissed):

jakedipity (approved):

mdeltito (dismissed):

jakedipity (approved):

jakedipity (approved):

mdeltito (approved):

PR #21, opened 2020-10-05
Refactor fs tailer to return a Stream

c-nixon:
Refactor the fs tailer to return a Stream<Item=Vec<LineBuilder>> and update the agent's main to consume it.

Internal Ref: LOG-5685


common/fs/src/cache/mod.rs
```diff
@@ -73,64 +79,90 @@ impl<T: Default> FileSystem<T> {
                 if !path_cpy.exists() {
                     path_cpy.pop();
                 } else {
-                    fs.insert(&path_cpy, &mut |_, _| {});
+                    fs.insert(&path_cpy, &mut Vec::new());
```
c-nixon, line 82: Is this right? I think it's equivalent... but I'm not 100% sure why we're ignoring the top level dirs

common/fs/src/cache/watch.rs
```diff
@@ -55,82 +72,210 @@ impl Watcher {
...
-                        _ => continue,
-                    };
-                }
+    pub fn event_stream<'a>(
+        &mut self,
+        buffer: &'a mut [u8],
+    ) -> std::io::Result<WatchEventStream<'a>> {
+        Ok(WatchEventStream {
+            event_stream: self.inotify.event_stream(buffer)?,
+        })
```
jakedipity, line 81: Why return the wrapper `WatchEventStream` around the stream? Why not just return the stream?

common/fs/src/cache/watch.rs
```diff
@@ -2,7 +2,15 @@ use inotify::{EventMask, Inotify, WatchDescriptor, WatchMask};
...
 use std::path::Path;
+use std::sync::Arc;
 
+use futures::future::Either;
+use futures::{Stream, StreamExt};
+use tokio::time::Instant;
+
+use tokio::sync::Mutex;
+
+const INOTIFY_EVENT_GRACE_PERIOD_MS: u64 = 1000;
```
jakedipity, line 13: I know we're probably gonna play around with this value, but 1 second seems a bit excessive of a starting point. I imagine a move event split across multiple events will appear a couple milliseconds from each other.

common/fs/src/tail.rs
```diff
@@ -167,5 +208,34 @@ impl Tailer {
...
+//                 error!("can't start filesystem source: missing rules");
+//                 return;
+//             }
+//         };
+
+//         spawn(async move {
+//             let fs = FileSystem::<u64>::new(watched_dirs, rules);
+//         });
+//     }
+// }
```
jakedipity, line 241: Can probably just remove dead code

jakedipity (commented):
Couple of questions but largely looks good to me. Everything makes sense

bin/src/main.rs
```diff
@@ -67,22 +66,32 @@ fn main() {
...
 
-    loop {
-        source_reader.drain(Box::new(|lines| {
-            if let Some(lines) = executor.process(lines) {
-                for line in lines {
-                    client.borrow_mut().send(line)
+    let mut sources = futures::stream::SelectAll::new();
+    sources.push(&mut fs_source);
+    // Create the runtime
+    let mut rt = Runtime::new().unwrap();
```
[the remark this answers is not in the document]
c-nixon: Given we're starting up a pretty much default runtime, no, no advantage really except that it's less magical. If we ever do need to configure the runtime this gives us a handle to it, but it's a simple.refactor then if you'd prefer to use the annotation here

bin/src/main.rs
```diff
@@ -67,22 +66,32 @@ fn main() {
             Err(e) => warn!("{}", e),
         };
     }
+    executor.init();
 
-    let mut source_reader = SourceReader::new();
-    register_journald_source(&mut source_reader);
-    source_reader.register(FSSource::new(config.log.dirs, config.log.rules));
+    let mut fs_tailer_buf = [0u8; 4096];
```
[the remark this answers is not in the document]
c-nixon: Yeh, having the caller provide stratch space for the callee is a little ugly, it makes things simpler when things need to live long enough to get back to the callee and you'd prefer not to allocate, but it can sometimes be worked around. I'll see if I can smuggle the buffer around using a once_cell or an additional wrapper object, but we might be stuck with this...

common/fs/src/cache/mod.rs
```diff
@@ -73,64 +79,90 @@ impl<T: Default> FileSystem<T> {
...
-        for event in events {
-            self.process(event, &mut callback);
-        }
+        // TODO prepend these onto event stream?
+        // let mut acc = Vec::new();
+        // if !self.initial_events.is_empty() {
+        //     for event in std::mem::replace(&mut self.initial_events, Vec::new()) {
+        //         acc.push(event)
+        //     }
+        // }
```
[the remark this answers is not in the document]
c-nixon: I thought I'd pushed up the fix for this, my bad, I have it fixed using StreamExt's chain method

common/fs/src/cache/watch.rs
c-nixon: Mostly it was just for the convenience of having a type to impl against rather that whatever monstrous thing into_stream's concrete return type resolves too. A tangential benefit is that by wrapping it up it allows us to return the stream and strart processing it with WatchEventStream::into_stream in separate steps.

I could just return the impl Stream here, but it would be a bit less flexible, do you have a strong feeling either way?

bin/src/main.rs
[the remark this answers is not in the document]
jakedipity: The explanation makes sense. I don't prefer one to the other so I think it's okay to leave as is.

common/fs/src/cache/watch.rs
jakedipity: Whatever we do we should be consistent. I don't have an issue with doing things as they are now. In my implementation of converting a Journald instance into a stream I'll try and implement something similar for the same minor benefits.

jakedipity (commented):
The changed items look good. I think there's some outstanding comments that need addressing.

common/fs/src/cache/watch.rs
c-nixon: I've reduced it down to 100ms, that still feels really big, but I havn't been able to find documentation anywhere on what the maximum could be.

The only current art I could find is [notify-rs](https://github.com/notify-rs/notify/blob/80204c6f246b08278065b8cb5362630b10423fcf/src/inotify.rs#L384), which uses 10ms. Should I drop it further down to match that? I'm not entirely sure what the repercussions are either way

jakedipity (dismissed):
Looks good to me, just squash the commits before merging. Just lmk so I can approve again

jakedipity (approved):

PR #23, opened 2020-10-12
Integration test

c-nixon:
This PR adds a simple example integration test for the agent CLI.

It creates a temp dir, configures the agent to watch the temp dir, starts the agent, creates some files and checks that the agent successfully runs.

It does not currently validate that the expected events happen, only that it does not crash as long as a valid API key is present.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,10 +16,22 @@ pipeline {
     }
     stages {
         stage('Test') {
+            environment {
+                CREDS_FILE = credentials('pipeline-e2e-creds')
+                LOGDNA_HOST = "logs.use.stage.logdna.net"
+            }
             steps {
+                script {
+                    def creds = readJSON file: CREDS_FILE
+                    // Assumes the pipeline-e2e-creds format remains the same. Chase
+                    // refer to the e2e tests's README's authorization docs for the
+                    // current structure
+                    LOGDNA_INGESTION_KEY = creds["packet-stage"]["account"]["ingestionkey"]
+                }
                 sh """
                     make lint
                     make test
+                    make integration-test LOGDNA_INGESTION_KEY=${LOGDNA_INGESTION_KEY}
                 """
             }
             post {
--- a/Makefile
+++ b/Makefile
@@ -56,19 +56,23 @@ REMOTE_BRANCH := $(shell git branch -vv | awk '/^\*/{split(substr($$4, 2, length
 
 .PHONY:build
 build: ## Build the agent
-	$(RUST_COMMAND) "cargo build"
+	$(RUST_COMMAND) "--env RUST_BACKTRACE=full" "cargo build"
 
 .PHONY:build-release
 build-release: ## Build a release version of the agent
-	$(RUST_COMMAND) "cargo build --release && strip ./target/release/logdna-agent"
+	$(RUST_COMMAND) "--env RUST_BACKTRACE=full" "cargo build --release && strip ./target/release/logdna-agent"
 
 .PHONY:test
 test: ## Run unit tests
-	$(RUST_COMMAND) "cargo test"
+	$(RUST_COMMAND) "--env RUST_BACKTRACE=full" "cargo test"
+
+.PHONY:integration-test
+integration-test: ## Run integration tests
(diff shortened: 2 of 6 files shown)


Jenkinsfile
```diff
@@ -13,10 +13,22 @@ pipeline {
     }
     stages {
         stage('Test') {
+            environment {
+                CREDS_FILE = credentials('pipeline-e2e-creds')
+                LOGDNA_HOST = "logs.use.stage.logdna.net"
```
jakedipity, line 18: Not sure we want to consider the staging environment stable enough to use for reliable testing. We can isolate the agent completely and use a mock ingester  that tracks what lines have come through. I believe we already have a yaml for that

Makefile
```diff
@@ -19,7 +19,7 @@ DOCKER_PUBLIC_IMAGE := docker.io/logdna/logdna-agent
 DOCKER_IBM_IMAGE := icr.io/ext/logdna-agent
 
 RUST_COMMAND := $(DOCKER_DISPATCH) $(RUST_IMAGE)
-HADOLINT_COMMAND := $(DOCKER_DISPATCH) $(HADOLINT_IMAGE)
+HADOLINT_COMMAND := $(DOCKER_DISPATCH) $(HADOLINT_IMAGE) ""
```
jakedipity, line 22: This change seems a little mysterious

bin/tests/cli.rs
```diff
@@ -0,0 +1,82 @@
...
+use tempfile::tempdir;
+
+use std::fs::File;
+use std::io::Write;
+use std::process::Command;
+use std::thread;
+
+#[test]
+fn api_key_missing() {
+    let mut cmd = Command::cargo_bin("logdna-agent").unwrap();
```
jakedipity, line 13: We seem to instantiate this object in many places. Does it make sense to create it once and have it available for the entire program?

Makefile
c-nixon: Yeh, it's a bit awkward, I can move it to where we invoke the hadolint command to make it more obvious, it's just the equivalent of passing a null argument, unfortunately bash doesn't have named args.

Jenkinsfile
c-nixon: I was going to build that functionality into a future version of the test binary, I didn't know we had a mock ingester.

I used staging as that's what Test suggested. I'm not _too_ concerned about the investees being available in staging, but I agree having something capture the lines and validate delivery would be better here.

I would still like to test that we successfully write to the real ingesters somewhere, maybeit should like in a separate test stage that doesn't automatically fail the build, or checks the api is available and skips if it isn't?

bin/tests/cli.rs
c-nixon: It's the initial state of a mutable builder object which we configure differently in each test, see the next line vs it's equivalent in the api_key_present test. Each of the methods takes a &mut, so to reuse it we'd need to clone it each time before we called any methods and unfortunately it doesn't impl Clone.

We _could_ factor it out into a function to reduce the boilerplate a little?

jakedipity: I should have payed closer attention. I'm assuming the builder object is modified in the steps below in which case my suggestion doesn't make sense. Wrapping a function around it won't save much of any boilerplate... so it's probably fine to leave as is

jakedipity (approved):
I think what's here looks good. I'm a bit hesitant as the testing seems partially complete - we don't validate lines have been ingested. I think it would be okay to tackle this in a later update

jakedipity (dismissed):
Looks good to me, before you merge can you squash the commits into one with a tidied up commit message

bin/tests/cli.rs
```diff
@@ -0,0 +1,118 @@
...
+    thread::sleep(std::time::Duration::from_secs(1));
+
+    handle.kill().unwrap();
+    let mut output = String::new();
+
+    let stderr_ref = handle.stderr.as_mut().unwrap();
+    stderr_ref.read_to_string(&mut output).unwrap();
+
+    // Check that the agent logs that it has sent lines from each file
+    assert!(predicate::str::contains(&format!(
```
c-nixon, line 84: Tada, it now actually checks things.

jakedipity (dismissed):
LGTM

PR #24, opened 2020-10-13
Run logdna-agent as non-root user

jakedipity:
This change gives the logdna-agent special read capabilities over the filesystem so we don't need to run as the root user. This alleviates a common security issue with the agent


c-nixon (approved):
Does what it says on the tin.

PR #27, opened 2020-10-14
Add fs and sccache caching to Makefile

c-nixon:
Add fs and sccache caching to the makefile to cut down on local and CI run times.

Also add Shellcheck linting and modify the docker_dispatch script to handle signals


Makefile
```diff
@@ -73,6 +74,7 @@ integration-test: ## Run integration tests
 .PHONY:clean
 clean: ## Clean all artifacts from the build process
 	$(RUST_COMMAND) "--env RUST_BACKTRACE=full" "cargo clean"
+	rm -fr $(CARGO_CACHE)/*
```
jakedipity, line 77: the `clean-all` command doesn't benefit from this change, but it should. The easiest thing to do is to make `clean` a dependency of `clean-all` and remove the target specific logic

docker/dispatch.sh
```diff
@@ -1,13 +1,70 @@
 #!/usr/bin/env sh
 
+_term() {
```
jakedipity, line 3: I think there's a couple issues with this script (and maybe was before too) which you can see highlighted with a shell checker. We should add it as part of the CI process and you can fix the issues it raises

c-nixon: Huh, shellcheck really wants me to quote ints. I've applied the suggestions that don't break things and ignored the rest.

Makefile
```diff
@@ -178,7 +190,8 @@ build-image: ## Build a docker image as specified in the Dockerfile
 		--build-arg BUILD_VERSION=$(BUILD_VERSION) \
 		--build-arg REPO=$(REPO) \
 		--build-arg VCS_REF=$(VCS_REF) \
-		--build-arg VCS_URL=$(VCS_URL)
+		--build-arg VCS_URL=$(VCS_URL) \
+		--build-arg SCCACHE_BUCKET=$(SCCACHE_BUCKET)
```
jakedipity, line 194: Do you also need
```
--build-arg SCCACHE_REGION=$(SCCACHE_REGION)
```

docker/dispatch.sh
```diff
@@ -1,13 +1,69 @@
...
+cargo_home=$(docker inspect -f \
+        '{{range $index, $value := .Config.Env}}{{println $value}}{{end}}' \
+            "$3" | grep CARGO_HOME | cut -d"=" -f 2)
+if [ -z "$cargo_home" ]; then
+	cargo_home="/usr/local/cargo"
+fi
+
+[ ! -d "$CARGO_CACHE" ] && mkdir -p "$CARGO_CACHE"
+
+volume_mounts="$volume_mounts -v $CARGO_CACHE:$cargo_home/registry:Z"
```
jakedipity, line 27: Will `$volume_mounts` ever not be just `""` because of line 15?

c-nixon: Yeh, it's quite possible, in that case this line should resolve to " -v $CARGO_CACHE:$cargo_home/registry:Z", which is still valid

jakedipity (approved):

PR #28, opened 2020-10-16
Changed paths in install to assets.logdna.com

johnpadilla:
Paths referred to unreleased version


c-nixon (approved):
This makes sense to me. If we want to provide instructions for installing an image based on a pre-release version we can put them in a separate document.

PR #29, opened 2020-10-19
Port from quickerror to thiserror

c-nixon:
Thiserror (and it's partner crate Anyhow) are more or less the de-facto standard for managing custom error types in rust. thiserror operates as a proc macro so the compiler has an easier time looking inside it than with quick-error.

For example while porting over to it the compiler detected that the k8s::middleware::metadata::Error::Regex variant is actually dead code and never constructed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -494,8 +494,8 @@ dependencies = [
  "log",
  "metrics",
  "pcre2",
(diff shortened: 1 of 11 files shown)


jakedipity (approved):
Seems like a good change!

PR #30, opened 2020-10-20
Add journald streaming

jakedipity:


Makefile
```diff
@@ -172,7 +172,7 @@ release: ## Create a new release from the current beta and push to github
 .PHONY:build-image
 build-image: ## Build a docker image as specified in the Dockerfile
 	$(DOCKER) build . -t $(REPO):$(BUILD_TAG) \
-		$(PULL_OPTS) --no-cache=true --rm \
```
jakedipity, line 175: These shouldn't be here, leftover to streamline the dev process.

c-nixon (commented):
Noted a few things, I think it mostly looks good, I think we definitely need some tests around the stream alike to the fs tests

common/journald/src/stream.rs
```diff
@@ -0,0 +1,344 @@
...
+    Files(Vec<PathBuf>),
+}
+
+enum RecordStatus {
+    Line(LineBuilder),
+    BadLine,
+    NoLines,
+}
+
+struct SharedState {
```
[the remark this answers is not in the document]
jakedipity: I've since added a second field to the object to ensure the thread is cleaned up when the main stream is dropped. I don't think AtomicCell is the correct primitive to be using here and I don't think the single use case warrants importing yet another package (although with sccache now I can hardly tell the difference :P)

jakedipity:
test this please

jakedipity:
test this please

jakedipity:
test this please

jakedipity:
@logdnabot test this

jakedipity:
@logdnabot can you please...

common/journald/src/stream.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+impl Stream {
+    pub fn new(path: Path) -> Self {
+        let mut stream = Self {
+            thread: None,
+            receiver: None,
+            shared_state: Arc::new(Mutex::new(SharedState {
+                waker: None,
+                is_alive: true,
```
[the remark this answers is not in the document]
c-nixon: Example implementation [here](https://github.com/c-nixon/logdna-agent-v2/commit/0eed097424b0739efcf92ac00e04dd41fe5a8bcc), with the channel we have the added advantage that dropping the stream struct would signal the worker thread to exit (by closing the stop channel) even without Stream's drop implementation

common/journald/src/stream.rs
```diff
@@ -0,0 +1,356 @@
...
+impl FutureStream for Stream {
+    type Item = Vec<LineBuilder>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+        let mut self_ = self.as_mut();
+
+        if let Some(ref receiver) = self_.receiver {
+            match receiver.try_recv() {
+                Ok(line) => {
+                    return Poll::Ready(Some(vec![line]));
```
c-nixon, line 156: I'd love some way to avoid or reuse the vecs we're passing back here, allocating for every return hurts me on some basic level. That's a future problem thought

common/journald/src/stream.rs
```diff
@@ -0,0 +1,356 @@
...
+                }
+                Err(TryRecvError::Disconnected) => {
```
c-nixon, line 158: This looks like the agent will go into an infinite retry loop. Is that the best thing we can do here? Or should we allow the panic to propagate upward and kill the agent after a certain number or retries?

I don't think this blocks us from merging this, but I'd like to add a TODO around improving the fault handling

common/journald/src/stream.rs
```diff
@@ -0,0 +1,356 @@
...
+            .expect("Could not seek to tail of journald logs");
+
+        Self { reader }
+    }
+
+    fn process_next_record(&mut self) -> RecordStatus {
+        let record = match self.reader.next_entry() {
+            Ok(Some(record)) => record,
+            Ok(None) => return RecordStatus::NoLines,
+            Err(e) => panic!("Unable to read next record from journald: {}", e),
```
c-nixon, line 209: Is this condition unrecoverable? Or could it arise if the journald daemon was restarted or something? In which case I think it would be better return a Result here and let the worker thread try recreating it's journal object before resorting to the infinite retry logic in the Stream? (which feels like exception handling and a bit un-rustic)

common/journald/src/stream.rs
```diff
@@ -0,0 +1,356 @@
...
+        };
+
+        let timestamp = match self.reader.timestamp() {
+            Ok(timestamp) => Local
+                .timestamp(
+                    timestamp.duration_since(UNIX_EPOCH).unwrap().as_secs() as i64,
```
c-nixon, line 215: I don't think we can always unwrap this, especially if we ever look back to the start of the log. It's possible that the system would come up with it's state some time in the future, the get set to the correct time by ntpd. Between boot and ntpd setting things straight logs could be generated, then when we start reading the logs would look like they were in the future and this would panic.

We can't do anything about the dodgy timestamps, but we could still handle this error and warn! that we have sent some logs from the future

common/journald/src/stream.rs
```diff
@@ -0,0 +1,356 @@
...
+    const JOURNALD_LOG_PATH: &str = "/var/log/journal";
+
+    #[tokio::test]
+    #[serial]
+    async fn reader_gets_new_logs() {
+        journal::print(1, "Reader got the correct line!");
+        sleep(Duration::from_millis(50));
+        let mut reader = Reader::new(Path::Directory(JOURNALD_LOG_PATH.into()));
+
+        let record_status = reader.process_next_record();
```
c-nixon, line 295: This could get weird if the tests are running as the root user, right? Like another process could write a log line and we'd get it instead?

If it works in our container then that's cool, but maybe we could add a todo to switch to use [journal::log](https://docs.rs/systemd/0.7.0/systemd/journal/fn.log.html) and passing the [current line](https://docs.rs/systemd/0.7.0/systemd/journal/fn.log.html) and such to make the message uniquely identifiable.

c-nixon (commented):
Nice job, just a couple of comments and queries. Don't think there's anything blocking here, but it might be work adding some todos.

common/journald/src/stream.rs
jakedipity: dropped a comment

c-nixon (dismissed):
I'm good with this as is, we can address the more flexible testing as we go.

jakedipity:
@logdnabot test this

jakedipity:
@logdnabot test this

c-nixon (approved):

PR #31, opened 2020-10-22
Add dynamic targets for testing individual crates

c-nixon:
This dynamically generates a test- target for each sub crate allows which means you can run

make test-<component name>

and only run the unit tests for a specific sub-crate

eg `make test-fs` will run cargo test -p fs

Do you think this is useful enough to include @jakedipity?


jakedipity (approved):
LGTM

PR #37, opened 2020-11-02
K8s event stream

c-nixon:


common/k8s/src/middleware/metadata.rs
```diff
@@ -1,14 +1,13 @@
 use crate::errors::K8sError;
```
jakedipity, line 1: Just to be clear, all the changes here are just because of the updated dependencies e.g. the behavior hasn't changed. That's what it appears to be and the logic looks sound to me when compared with the updated docs 👍

common/k8s/src/event_source.rs
```diff
@@ -0,0 +1,212 @@
...
+                    .map(|age| {
+                        age.num_seconds()
+                            .try_into()
+                            .map_err(|_| warn!("age too large, could not fit {} into i32", age))
+                            .ok()
+                    })
+                    .flatten(),
+                count,
+            },
+        }
```
jakedipity, line 121: This structure looks good to me. Would be useful if we could compare what this line looks like in the web UI and compare it with the kube reporter - want to make sure everything matches up.

jakedipity (approved):
Mostly looks good to me. There are some areas you marked that need to be cleaned up.

common/k8s/src/middleware/metadata.rs
c-nixon: Yup, this is purely around using the newer API version which has better types for deserializing events and a more reboust watcher mechanism

common/k8s/src/event_source.rs
c-nixon: Yeh, this is based entirely on the reporter's class/EventWatcher.js handling. I was going to add a link in a comment but it's not in a public repo. I'll add a full example in the unit tests and include the reporter's example payload and test case

common/k8s/src/restarting_stream.rs
```diff
@@ -0,0 +1,60 @@
+use std::pin::Pin;
```
c-nixon, line 1: At some point we should probably split this out into it's own utility crate. It's pretty generic

common/k8s/src/event_source.rs
```diff
@@ -0,0 +1,226 @@
...
+        let start_stream = {
+            let node_name = self.node_name.clone();
+            let api = self.api.clone();
+            move || {
+                let mut params = ListParams::default();
+                if let Some(ref node_name) = node_name {
+                    params.field_selector = Some(format!("source.host=={}", node_name));
+                } else {
+                    info!("Node name not available, logging all cluster events, will cause duplication per node.")
+                }
```
jakedipity, line 202: I believe this still needs to be changed because you can't rely on `source.host` as a field selector as you've mentioned before?

k8s/logdna-agent.yaml
```diff
@@ -121,6 +121,16 @@ spec:
                   key: logdna-agent-key
             - name: RUST_LOG
               value: info
+            - name: LOGDNA_LOG_K8S_EVENTS
+              value: always
```
c-nixon, line 125: Should this be on by default?

common/k8s/src/restarting_stream.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl<F, R, S, T, Fut> RestartingStream<F, R, S, Fut>
+where
+    S: Stream<Item = T>,
+    R: FnMut(&T) -> RequiresRestart,
+    F: FnMut() -> Fut,
+    Fut: Future<Output = S>,
+{
+    pub async fn new(mut f: F, r: R) -> Self {
+        let stream = f().await;
```
c-nixon, line 32: Have to put this up here or it complains about a use after move

common/k8s/src/event_source.rs
```diff
@@ -0,0 +1,408 @@
...
+                return Err(K8sError::InitializationError(format!(
+                    "unable to get cluster configuration info: {}",
+                    e
+                )))
+            }
+        };
+        Self::new(config, pod_name, namespace)
+    }
+
+    pub async fn event_stream(self) -> Result<impl Stream<Item = Vec<LineBuilder>> + Send, String> {
```
c-nixon, line 215: This is in dire need or being split up, but this is a first working version.

The general idea is that there are a couple of streams, on that will keep producing "wait" until the agent needs to start logging events, at which point the "wait" stream will end, and the event logging stream can start.

k8s/logdna-agent-st.yaml
```diff
@@ -123,6 +123,17 @@ spec:
               value: /supertenant/logs/ingest
             - name: RUST_LOG
               value: info
+            - name: LOGDNA_LOG_K8S_EVENTS
```
c-nixon, line 126: Which yamls do I need to add this to?

common/k8s/src/event_source.rs
```diff
@@ -0,0 +1,408 @@
...
+                                            move |e| {
+                                                match e {
+                                                    Ok(watcher::Event::Deleted(e)) => {
+                                                        info!("previous k8s event logger deleted");
+                                                        // We've detected the old one is dead, now what?
+                                                        delete_time.store(
+                                                            e.metadata
+                                                                .deletion_timestamp
+                                                                .map(|t| {
+                                                                    info!("Ignoring events before {}", t.0 - chrono::Duration::seconds(2));
```
jakedipity, line 302: Might be good to specify that it's ignoring kubernetes events

common/k8s/src/event_source.rs
```diff
@@ -0,0 +1,408 @@
...
+                                        chrono::NaiveDateTime::from_timestamp(earliest.into(), 0);
+                                    event.as_ref().ok().and_then(|e| {
+                                        e.last_timestamp
+                                            .as_ref()
+                                            .map(|l| earliest < l.0.naive_utc())
+                                    })
+                                });
+                                async move { ret.unwrap_or(true) }
+                            }
+                        })
```
jakedipity, line 349: If I'm understanding correctly, once the `event_stream` is unblocked by the `waiting_stream`, it's going to use the shared resource `previous_event_logger_delete_time` to filter out all the old events. This approach is a bit janky as it requires potentially reading through a long list of events to catch up. I'm not sure how many events Kube keeps around and for how long so it might not be a real cause concern.

If the Kube API allows us to specify a start time, maybe we can use your `RestartingStream` to build a fresh new stream once the `waiting_stream` unblocks. It does require a bit of rethinking tho... and might possible be even more confusing...

common/k8s/src/event_source.rs
```diff
@@ -0,0 +1,408 @@
...
+                        .map(|event| {
+                            match event.map(|e| {
+                                serde_json::to_string(&EventLogLine::from(e))
+                                    .map_err(K8sEventStreamError::SerializationError)
+                                    .map(|e| {
+                                        debug!("logging event: {}", e);
+                                        // if e.time > previous_event-logger_delete_time
+                                        vec![LineBuilder::new().line(e)]
+                                        // else None
```
jakedipity, line 358: These comments should be removed

k8s/logdna-agent-st.yaml
jakedipity: We might want to keep `logdna-agent.yaml` and `logdna-agent-st.yaml` unchanged, instead modifying `agent-resources.yaml` and `agent-resources-supertenant.yaml`

common/k8s/src/event_source.rs
c-nixon: Unfortunately you can't do > or < on labels with the kube api, so we can't set a start point there, I tried that first and wasn't able to find any mechanism to allow it.

By default kubernetes expires events after 1 hour has passed and pages them back 500 at a time so even completely reprocessing a stream of a few hundred thouasnd events shouldn't be a _huge_ load. And once it's caught up it's there isn't much work to do. I'd definitely like to do some testing around this though. Unfortunately k8s doesn't even necessarily ship events in order, so we have to keep the filter in place rather that just skipping until we're at desired point... https://github.com/kubernetes/kubernetes/issues/29838

k8s/logdna-agent.yaml
jakedipity: Not sure, should clarify during our meeting today

common/k8s/src/restarting_stream.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use futures::{ready, Stream};
+
+pub enum RequiresRestart {
+    Yes,
+    No,
+}
+
+#[pin_project]
+#[must_use = "streams do nothing unless polled"]
```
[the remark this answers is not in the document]
c-nixon: It seems to be needed if you have a variable with type RestartingStream rather than a binding with Stream as it's type

k8s/logdna-agent-st.yaml
c-nixon: I've added this (and the CAP_ stuff to all of the yamls as I don't think any of them would run without the CAP stuff.

Do you think I should remove the LOGDNA_LOG_K8S_EVENTS from logdna-agent.yaml and logdna-agent-st.yaml? I'm still not 100% clear on what each is used for

jakedipity: I had to really think back to the purpose of this file. These files are for backwards compatibility for the YAML files we host on our asset server. Those files are generated from the github files here. Keeping the old files was in case we had lingering documentation which referenced those old links to `logdna-agent.yaml`.

That being the case it's probably best to keep this files "old".

jakedipity (dismissed):

PR #38, opened 2020-11-02
Fix some issues found with hardening the agent:

jakedipity:
1. Make sure the group created in the Dockerfile is 5000 as intended
2. Give the `logdna` binary with the effective bit since it's not capability aware
3. Remove `securityContext` since we don't want to run as non-root in the default behavior
4. Add documentation about running as non-root


c-nixon (dismissed):
This looks good to me. I didn't realise allowPrivilegeEscalation: false implied no_new_privs last time I looked at this!

Could we try without the i and p flags on the capability? I don't mind p, but if we can do without i that would be good

jakedipity:
@logdnabot test this

jakedipity:
@logdnabot test this

c-nixon (approved):

PR #39, opened 2020-11-03
Add Kubernetes capabilities to the default yamls

jakedipity:
The linux cap `CAP_DAC_READ_SEARCH` needs to be enabled in the k8s yaml since the binary has the effective bit set. This is required regardless of running as root or non-root.


jakedipity:
@logdnabot test this

jakedipity:
@logdnabot test this

c-nixon (approved):
Tested and this works, LGTM

PR #40, opened 2020-11-04
Make a couple of changes to the Dockerfile

jakedipity:
1. Update the base image from 8.2 -> 8.3
2. Remove the complicated build steps (used to optimize layer caching and build times) which is largely replaced by sccache
3. Combine a couple RUN commands together to shave off ~10MiB from the final image


jakedipity:
@logdnabot test this please 👍

c-nixon (approved):
This looks good to me. It'll slow down repeated builds slightly for local dev but shouldn't effect CI. I think it's worth the simplification.

Happy with the bump to 8.3

PR #41, opened 2020-11-05
Change fs rule matching to use PathBufs

c-nixon:
This fixes a panic if a cached filepath is non-utf8

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -225,9 +225,9 @@ where
                     }
                 };
 
-                let is_to_path_ok = self.passes(to_path.to_str().unwrap());
+                let is_to_path_ok = self.passes(&to_path);
 
-                let is_from_path_ok = self.passes(from_path.to_str().unwrap());
+                let is_from_path_ok = self.passes(&from_path);
 
                 if is_to_path_ok && is_from_path_ok {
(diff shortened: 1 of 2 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -386,15 +386,15 @@ where
...
         paths: &mut Vec<PathBuf>,
-        mut components: Vec<String>,
+        mut components: Vec<OsString>,
     ) {
-        let mut base_components: Vec<String> = into_components(&self.resolve_direct_path(entry))
+        let mut base_components: Vec<OsString> = into_components(&self.resolve_direct_path(entry))
             .iter()
-            .map(|x| x.to_str().unwrap().into())
+            .map(|x| x.as_os_str().into())
             .collect(); // build all the components of the path leading up to the current entry
```
jakedipity, line 394: `into_components()` returns a `Vec<OsString>` so I think you can just omit the entire `.iter().map().collect()`

jakedipity (changes requested):

common/fs/src/cache/mod.rs
jakedipity: pushed up the change myself

c-nixon: Good spot. I'm good with this if you are

jakedipity (approved):

jakedipity (approved):

PR #42, opened 2020-11-06
Update documentation for 2.2.0

jakedipity:


docs/README.md
```diff
@@ -140,14 +101,27 @@ env:
 
 Check out [Kubernetes documentation](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/) for more information about injecting environment variables into applications!
 
-### Run as Non-Root
```
tanberry, line 143: maybe add this "Run as non-root user" back in?

docs/README.md
```diff
@@ -140,14 +101,27 @@ env:
...
+
+### Configuring Events
+
+Kubernetes and OpenShift Events are by default automatically captured by the agent. This feature can be controlled by the `LOGDNA_LOG_K8S_EVENTS` option with only two valid values:
+
+* `always` - Always capture Events
+* `never` - Never capture Events
+* __Note:__ The default option is `always`
+
+> :warning: Due to a number of issues with Kubernetes, the agent collects events from the entire cluster including multiple nodes. To prevent duplicate logs when running multiple pods, the agents defer responsibilty of capturing Events to the oldest pod in the cluster. If that pod is killed, the next oldest pod will take over responsibility and continue from where the previous pod left off.
```
tanberry, line 152: let's try to find a word other than "issues" here. Perhaps say something like "Doe to the native Kubernetes networking configurations, the agent collects..." or "... the default Kubernetes configuration..."

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,186 @@
...
+  2. If you want to continue using journald, follow the steps for [enabling journald monitoring on the agent](#enabling-journald-monitoring-on-the-agent).
+  3. Overwrite the DaemonSet as well as create the new OpenShift Objects.
+     1. Run `oc apply -f k8s/agent-resources-openshift.yaml`
+
+> :warning: Exporting OpenShift Objects with "oc get \<resource\> -o yaml" includes extra information about the Object's state. This data does not need to be copied over to the new yaml.
+
+### Upgrading the Image
+
+The image contains the actual agent code that is run on the Pods created by the DaemonSet. New versions of the agent always strive to be backwards compatibility with old configuration versions. Any breaking changes will be outlined on our release page. We always recommend upgrading to the latest configuration to get the best feature support for the agent.
+
```
tanberry, line 98: ".. always strive to be backwards compatible..." or "always strive for backwards compatibility..."

docs/KUBERNETES.md
```diff
@@ -0,0 +1,201 @@
...
+  * [Upgrading the Image](#upgrading-the-image)
+* [Uninstalling](#uninstalling)
+* [Run as Non-Root](#run-as-non-root)
+* [Collecting Node Journald Logs](#collecting-node-journald-logs)
+  * [Enabling Journald on the Node](#enabling-journald-on-the-node)
+  * [Enabling Journald Monitoring on the Agent](#enabling-journald-monitoring-on-the-agent)
+
+## Installing
+
+The agent can be effortless installed in your cluster using a set of yamls we provide. These yamls contain the minimum necessary Kubernetes Objects and settings to run the agent. Teams should review and modify these yamls for the specific needs of their clusters.
```
alfeng6, line 23: "can be effortlessly installed"

docs/KUBERNETES.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+Older versions of our configurations do not provide these labels. In that case, each upgrade path belows provides an example of each configuration to be compared to whats running on your cluster.
+
+#### Upgrading from Configuration v1.x.x or v2.0.x
+
+* **Example Configuration Yamls:**
+  * [v1.x.x](https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-ds.yaml)
+  * [v2.0.x](https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-v2.yaml)
+* **Differences:** This configuration does not include the new logdna-agent namespace and is lacking a number of new Kubernetes Objects.
```
alfeng6, line 81: Maybe "Differences with latest Agent v2"

docs/KUBERNETES.md
```diff
@@ -0,0 +1,201 @@
...
+* **Upgrade Steps:**
```
alfeng6, line 82: It's a little unclear with sub 1. 2. etc. Can you add in some headers?

docs/KUBERNETES.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+Otherwise, if your DaemonSet is configured with a different tag e.g. `logdna/logdna-agent:2.1.7`, you'll need to update the image and tag which will trigger a roll over of all the pods.
+
+```console
+kubectl patch daemonset -n logdna-agent logdna-agent --type json -p '[{"op":"replace","path":"/spec/template/spec/containers/0/image","value":"logdna/logdna-agent:2.2.0"}]'
+```
+
+The specific tag you should use depends on your requirements, we offer a list of tags for varying compatibility:
+1. `stable` - Updates with each major, minor, and patch version updates
```
alfeng6, line 125: don't we also have a `latest` tag?

docs/KUBERNETES.md
```diff
@@ -0,0 +1,201 @@
...
+## Collecting Node Journald Logs
+
+The agent by default only captures logs generated by the containers running on the Kubernetes clusters container runtime environment. It does not; however, collect system component logs from applications running directly on the node such as the kubelet and container runtime. With some configuration on both the node and the agent, these journald logs can be exposed from the node to the agent.
+
+The agent can access Journald logs from the host node by mounting the logs from `/var/log/journald`. This requires enabling journald log storage in the node as well as configuring the agent to monitor the directory.
+
+### Enabling Journald on the Node
+
+Follow the steps below to ensure journald logs are written to `/var/log/journald`:
+1. Gain root access to your node
```
alfeng6, line 181: by root access you mean at least global read right?

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,186 @@
+# LogDNA Agent on OpenShift
+
+The agent has been tested on OpenShift 4.4, but should be compatible with any OpenShift cluster packaged with Kubernetes version 1.9 or greater.
```
alfeng6, line 3: let's just recommend on OpenShift 4.4+ — because the customer doesn't care if you've tested or not, just a guarantee that it works with X version

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,186 @@
...
+  * [Upgrading the Image](#upgrading-the-image)
+* [Uninstalling](#uninstalling)
+* [Run as Non-Root](#run-as-non-root)
+* [Collecting Node Journald Logs](#collecting-node-journald-logs)
+  * [Enabling Journald on the Node](#enabling-journald-on-the-node)
+  * [Enabling Journald Monitoring on the Agent](#enabling-journald-monitoring-on-the-agent)
+
+## Installing
+
+The agent can be effortless installed in your cluster using a set of yamls we provide. These yamls contain the minimum necessary OpenShift Objects and settings to run the agent. Teams should review and modify these yamls for the specific needs of their clusters.
```
alfeng6, line 22: "can be effortlessly"

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+Otherwise, if your DaemonSet is configured with a different tag e.g. `logdna/logdna-agent:2.1.7`, you'll need to update the image and tag which will trigger a roll over of all the pods.
+
+```console
+oc patch daemonset logdna-agent --type json -p '[{"op":"replace","path":"/spec/template/spec/containers/0/image","value":"logdna/logdna-agent:2.2.0"}]'
+```
+
+The specific tag you should use depends on your requirements, we offer a list of tags for varying compatibility:
+1. `stable` - Updates with each major, minor, and patch version updates
```
alfeng6, line 114: latest?

docs/README.md
```diff
@@ -16,61 +16,45 @@ The LogDNA agent is a blazingly fast, resource efficient log collection client,
...
-    * [Run as Non-Root](#run-as-non-root)
+  * [Options](#options)
+  * [Configuring the Environment](#configuring-the-environment)
+  * [Configuring Lookback](#configuring-lookback)
 
-## Upgrading
+## Managing Deployments
 
-### Preamble
+The agent has been tested for deployment to Kubernetes 1.9+ and OpenShift 4.6+ environments.
```
alfeng6, line 34: let's say something along the lines of "this agent supports X version"

docs/README.md
```diff
@@ -97,7 +81,7 @@ The resulting image can be found by listing the images:
 ```console
 foo@bar:~$ docker images
 REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
-<none>              <none>              f541543bcd7f        64 seconds ago      119MB
+logdna-agent-v2     dcd54a0             e471b3d8a409        22 seconds ago      135MB
 ```
 
 ## Configuration
```
alfeng6, line 87: can you add a third column title? also the spacing can be changed so that the second column has more width

alfeng6 (changes requested):
A few changes to current PR — overall looks great.

docs/README.md
```diff
@@ -16,61 +16,45 @@ The LogDNA agent is a blazingly fast, resource efficient log collection client,
...
-## Installing
+### Uninstalling
 
-### Prerequisites
+* [Uninstalling Kubernetes](KUBERNETES.md#uninstalling)
+* [Uninstalling OpenShift](OPENSHIFT.md#uninstalling)
 
-* A LogDNA Account. Create an account with LogDNA by following our [quick start guide](https://docs.logdna.com/docs/logdna-quick-start-guide).
-* A LogDNA Ingestion Key. You can get your ingestion key at the top of [your account's Add a Log Source page](https://app.logdna.com/pages/add-host).
+### More Information
```
tanberry, line 51: Thanks for adding in mention of "non-root" here. 

For the title, maybe "Additional Installation Options"? 

Tho... journald... is that really an install option or a configuration option? I'm not sure why journald gets listed here but not Lookback, for example?

docs/OPENSHIFT.md
tanberry: yeah true... we need to avoid subjective words like "effortless" and "easy" and "fast."

tanberry (dismissed):

docs/KUBERNETES.md
```diff
@@ -0,0 +1,201 @@
...
+* [Uninstalling](#uninstalling)
+* [Run as Non-Root](#run-as-non-root)
+* [Collecting Node Journald Logs](#collecting-node-journald-logs)
+  * [Enabling Journald on the Node](#enabling-journald-on-the-node)
+  * [Enabling Journald Monitoring on the Agent](#enabling-journald-monitoring-on-the-agent)
+
+## Installing
+
+The agent can be effortless installed in your cluster using a set of yamls we provide. These yamls contain the minimum necessary Kubernetes Objects and settings to run the agent. Teams should review and modify these yamls for the specific needs of their clusters.
+
```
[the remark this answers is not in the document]
jakedipity: Updated all references to correctly specify `YAML files`

docs/KUBERNETES.md
jakedipity: I think the documentation will be easier to maintain and more robust if we don't specify the current version the docs refer to.

docs/KUBERNETES.md
jakedipity: I don't understand the ask here

docs/KUBERNETES.md
jakedipity: This doesn't pertain to the agent's container and running as a non-root user. This is for administrators to directly enable journald logging on the node itsself. They will require READ+WRITE permissions for some systemd configuration files and root is the easiest option to get that.

docs/OPENSHIFT.md
jakedipity: Fix with the latter

docs/README.md
```diff
@@ -16,61 +16,21 @@ The LogDNA agent is a blazingly fast, resource efficient log collection client,
```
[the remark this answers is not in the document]
jakedipity: Removed!

docs/README.md
```diff
@@ -16,61 +16,21 @@ The LogDNA agent is a blazingly fast, resource efficient log collection client,
...
 ## Table of Contents
 
-* [Upgrading](#upgrading)
-    * [Preamble](#preamble)
-    * [Upgrading on Kubernetes](#upgrading-on-kubernetes)
-* [Installing](#installing)
-    * [Prerequisites](#prerequisites)
-    * [Installing on Kubernetes](#installing-on-kubernetes)
+* [Deploying](#deploying)
 * [Building](#building)
```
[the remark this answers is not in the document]
jakedipity: There's no two separate steps for deploying and installing in our documentation - they're one and the same. Deploying is a better term for it but we've been going with installing for a while.

I added some text in the install step to reference both installing and deploying.

docs/README.md
jakedipity: fixed

docs/README.md
jakedipity: I went with your first suggestion of raising the non-root section directly. It does seem like an important feature for cluster maintainers so I do see the case for raising it directly in the landing page documentation.

The configuration section of the Journald feature is environment agnostic and only talks about the feature as it pertains to the agent. If you understand how journald works, you can use that information to do whatever they want with the feature.

The environment specific journald setup instructions talk about how to actually enable journald in their perspective environments. They're actual instructions.

docs/README.md
jakedipity: Added more width, but didn't understand what you meant with the first ask

docs/README.md
jakedipity: Added it back in at the top of the page

docs/README.md
```diff
@@ -16,61 +16,21 @@ The LogDNA agent is a blazingly fast, resource efficient log collection client,
...
     * [Building on Linux](#building-on-linux)
     * [Building on Docker](#building-on-docker)
 * [Configuration](#configuration)
     * [Options](#options)
```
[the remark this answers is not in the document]
jakedipity: done

docs/README.md
jakedipity: I've updated this to directly call out the issue in question

docs/KUBERNETES.md
jakedipity: Fixed by removing the subjective word

docs/KUBERNETES.md
jakedipity: fixed

docs/OPENSHIFT.md
jakedipity: Sounds good, also this should be `4.6` and not `4.4`

docs/OPENSHIFT.md
jakedipity: Removed!

docs/OPENSHIFT.md
jakedipity: fixed

docs/KUBERNETES.md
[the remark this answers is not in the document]
johnpadilla: "Objects" shouldn't be capitalized

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+### Installation Prerequisites
+
+* LogDNA Account - Create an account with LogDNA by following our [quick start guide](https://docs.logdna.com/docs/logdna-quick-start-guide).
+* LogDNA Ingestion Key - You can find an ingestion key at the top of [your account's Add a Log Source page](https://app.logdna.com/pages/add-host).
+* Kubernetes cluster running at least version 1.9.
+* Local clone of this repository.
+
+### Installation Steps
+
+1. Navigate to the root directory of the cloned `logdna-agent` repository.
```
johnpadilla, line 34: "logdna-agent-v2"?

johnpadilla (changes requested):
Wow, it's a lot. Sorry...

docs/KUBERNETES.md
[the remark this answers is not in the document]
jakedipity: Fixed this an other capitalization errors around Kubernetes primitives.

docs/README.md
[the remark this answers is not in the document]
jakedipity: Fixed the first one, left the second one alone. The version statement isn't true with the Linux Agent 2.1 so I've made it not refer to version at all and instead refer to just "This version of the agent".

docs/README.md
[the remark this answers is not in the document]
jakedipity: Fixed the parallel issue. Leaving the `Installing` for now to see what @tanberry has to say

docs/KUBERNETES.md
jakedipity: Fix the "Objects"

docs/KUBERNETES.md
jakedipity: AFAIK I can't control the notation used for lists. Fixed the unnecessary substep here and somewhere else. Changed the last step to to how you suggested.

docs/KUBERNETES.md
jakedipity: Simple enough

docs/README.md
```diff
@@ -140,14 +136,27 @@ env:
...
+
+### Configuring Events
+
+Kubernetes and OpenShift events are by default automatically captured by the agent. This feature can be controlled by the `LOGDNA_LOG_K8S_EVENTS` option with only two valid values:
+
+* `always` - Always capture events
+* `never` - Never capture events
+* __Note:__ The default option is `always`
+
+> :warning: Due to a ["won't fix" bug in the Kubernetes API](https://github.com/kubernetes/kubernetes/issues/41743), the agent collects events from the entire cluster including multiple nodes. To prevent duplicate logs when running multiple pods, the agents defer responsibilty of capturing events to the oldest pod in the cluster. If that pod is killed, the next oldest pod will take over responsibility and continue from where the previous pod left off.
```
tanberry, line 162: Consider this rewrite:
"Due to a ["won't fix" bug in the Kubernetes API](https://github.com/kubernetes/kubernetes/issues/41743), the LogDNA agent collects events from the entire cluster, including multiple nodes. To prevent duplicate logs when running multiple pods, the LogDNA agents on each pod defer responsibility of capturing events to the agent on the oldest pod in the cluster. If that pod is down, the LogDNA agent on the next oldest pod will take over responsibility and continue from where the previous pod left off.

docs/README.md
```diff
@@ -140,14 +136,27 @@ env:
...
+### Configuring Journald
+
+If the agent pods have access to journald log files or directories, monitoring can be enabled on them with the `LOGDNA_JOURNALD_PATHS`. Common values include `/var/log/journald` and `/run/systemd/journal`. To specify both, use a comma separated list: `/var/log/journald,/run/systemd/journal`.
+
+Take a look at enabling journald monitoring for [Kubernetes](KUBERNETES.md#collecting-node-journald-logs) or [OpenShift](OPENSHIFT.md#collecting-node-journald-logs).
+
+### Configuring Events
+
+Kubernetes and OpenShift events are by default automatically captured by the agent. This feature can be controlled by the `LOGDNA_LOG_K8S_EVENTS` option with only two valid values:
+
```
tanberry, line 157: Consider this instead, which highlights the "by default"part and avoids passive voice:

"By default, the LogDNA agent captures Kubernetes events (and Openshift events, as well, since Openshift is built on Kubernetes clusters). 

To control whether the LogDNA agent collects Kubernetes events, configure the `LOGDNA_LOG_K8S_EVENTS` option using one of these two values:"
+++++

Also, we need to tell readers WHERE to configure the `LOGDNA_LOG_K8S_EVENTS` option.

docs/README.md
```diff
@@ -140,14 +136,27 @@ env:
...
-  runAsGroup: 5000
-```
+### Configuring Journald
+
+If the agent pods have access to journald log files or directories, monitoring can be enabled on them with the `LOGDNA_JOURNALD_PATHS`. Common values include `/var/log/journald` and `/run/systemd/journal`. To specify both, use a comma separated list: `/var/log/journald,/run/systemd/journal`.
+
+Take a look at enabling journald monitoring for [Kubernetes](KUBERNETES.md#collecting-node-journald-logs) or [OpenShift](OPENSHIFT.md#collecting-node-journald-logs).
+
+### Configuring Events
+
```
tanberry, line 155: Let's add a bit about what the Kube Events resource is:

"A Kubernetes event is exactly what it sound like: a resource type that is automatically generated when state changes occur in other resources, or when errors or other messages manifest across the system. Monitoring events is useful for debugging your Kubernetes cluster.

tanberry (changes requested):

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
+> :warning: Exporting Kubernetes Objects with "kubectl get \<resource\> -o yaml" includes extra information about the Object's state. This data does not need to be copied over to the new YAML file.
+
+### Upgrading the Image
+
+The image contains the actual agent code that is run on the Pods created by the DaemonSet. New versions of the agent always strive for backwards compatibility with old configuration versions. Any breaking changes will be outlined on our release page. We always recommend upgrading to the latest configuration to get the best feature support for the agent.
+
+The upgrade path for the image depends on which image tag you are using in your DaemonSet.
+
+If your DaemonSet is configured with `logdna/logdna-agent:stable`, our default configuration setting, then you just need to delete the pods to trigger them to recreate and pull down the latest stable version of the logdna-agent image.
```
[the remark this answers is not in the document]
jakedipity: Fixed. Kind of annoyed Kubernetes documentation refers to `Pods` as a proper noun but not `Events`. But okay fixed. Tried to reword the second half of the sentence to read a bit more fluidly and actively.

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
```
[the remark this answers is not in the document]
jakedipity: fixed

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+```console
+kubectl delete pod -n logdna-agent -l app.kubernetes.io/name=logdna-agent
```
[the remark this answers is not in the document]
jakedipity: You don't gain anything by doing a rolling update since it's one pod to a node, but it does seem like the cleaner approach. Fixed.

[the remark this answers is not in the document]
jakedipity: Also tested on Kubernetes. Will test on OpenShift later tonight, but I'm 98% sure it'll work

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+Otherwise, if your DaemonSet is configured with a different tag e.g. `logdna/logdna-agent:2.1.7`, you'll need to update the image and tag which will trigger a roll over of all the pods.
```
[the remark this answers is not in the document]
jakedipity: fixed

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
```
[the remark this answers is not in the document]
jakedipity: Good call. fixed

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+1. `latest` - Update with each new revision including public betas
+2. `stable` - Updates with each major, minor, and patch version updates
+3. `2` - Updates with each minor and patch version updates under `2.x.x`
+4. `2.2` - Updates with each patch version update under `2.2.x`
+5. `2.2.0` - Targets a specific version of the agent
+
+**Note:** This list isn't exhaustive; for a full list check out the [logdna-agent dockerhub page](https://hub.docker.com/r/logdna/logdna-agent)
+
```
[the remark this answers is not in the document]
jakedipity: Yup, fixed

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+If you're sharing the namespace with other applications, you can also remove all traces of the agent by deleting with a label filter. You'll also need to remove the logdna-agent-key secret which doesn't have a label:
+
+```console
+kubectl api-resources --verbs=list --namespaced -o name | xargs -n 1 kubectl get --show-kind --ignore-not-found -n <NAMESPACE> -l app.kubernetes.io/name=logdna-agent -o name | xargs kubectl delete -n <NAMESPACE>
+kubectl delete secret -n <NAMESPACE> logdna-agent-key
+```
+
+## Run as Non-Root
+
+By default the agent is configured to run as root. If this behavior is not desired, the DaemonSet can be modified to run the agent as a non-root user.
```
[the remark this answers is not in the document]
jakedipity: Fixed the first sentence but tweaked it a bit from your suggestion; lmk what you think.

Updated the second paragraph to just be better written.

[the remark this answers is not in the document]
jakedipity: Already talked about this bit in a DM, but just sharing here again:

That change would imply the existence of multiple root users when in fact there's just one and it's generally referred to by as just `root`.

Also removed the reference to `local configuration file` to make it less confusing

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
+This is accomplished through Linux Capabilities that makes the agent a "Capability-dumb binary." Specifically, the agent is only allowed global read access with `CAP_DAC_READ_SEARCH`. This capability is already baked into the image and configuration. The only required step is configuring the agent DaemonSet to run as the user and group `5000:5000`.
```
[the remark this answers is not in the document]
jakedipity: Appreciate the thoroughness and accuracy! fixed

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
+To update the local configuration file, add two new fields, `runAsUser` and `runAsGroup`, to the `securityContext` section found in the `logdna-agent` container in the `logdna-agent` DaemonSet inside of `k8s/agent-resources.yaml` [`spec.template.spec.containers.0.securityContext`]:
+
```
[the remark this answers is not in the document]
jakedipity: I'm hesitant to reference what the updated yaml will look like. This doesn't take into account what the customer might have added themselves and is likely to fall behind as new changes are made to the yamls. But I also understand why it's helpful to have it.

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+```yaml
+securityContext:
+  runAsUser: 5000
+  runAsGroup: 5000
+```
+
+Apply the new updated configuration to your cluster:
```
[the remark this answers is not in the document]
jakedipity: fixed

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
+```console
+kubectl apply -f k8s/agent-resources.yaml
+```
+
+Alternatively, to update the DaemonSet configuration directly in your cluster use the following patch command:
```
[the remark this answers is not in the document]
jakedipity: Fixed the wordiness.

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
+```console
+kubectl patch daemonset -n logdna-agent logdna-agent --type json -p '[{"op":"add","path":"/spec/template/spec/containers/0/securityContext/runAsUser","value":5000},{"op":"add","path":"/spec/template/spec/containers/0/securityContext/runAsGroup","value":5000}]'
+```
+
+## Collecting Node Journald Logs
+
+The agent by default only captures logs generated by the containers running on the Kubernetes clusters container runtime environment. It does not; however, collect system component logs from applications running directly on the node such as the kubelet and container runtime. With some configuration on both the node and the agent, these journald logs can be exposed from the node to the agent.
```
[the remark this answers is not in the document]
jakedipity: Nice catch, fixed

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
+The agent can access Journald logs from the host node by mounting the logs from `/var/log/journald`. This requires enabling journald log storage in the node as well as configuring the agent to monitor the directory.
+
+### Enabling Journald on the Node
+
+Follow the steps below to ensure journald logs are written to `/var/log/journald`:
+1. Gain root access to your node
+2. Ensure the `journald.conf`, usually found at `/etc/systemd/`, sets `Storage=persistent`. Look at the [journald.conf documentation](https://www.freedesktop.org/software/systemd/man/journald.conf.html) for more information
+3. Create the directory `/var/log/journald`: `mkdir -p /var/log/journald`
```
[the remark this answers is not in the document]
jakedipity: Fixed this and other instances.

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Enabling Journald Monitoring on the Agent
+
+To enable Journald monitoring in the agent, add a new environment variable, `LOGDNA_JOURNALD_PATHS` with a value of `/var/log/journald`, to the logdna-agent DaemonSet:
+* If you are updating an already deployed agent:
```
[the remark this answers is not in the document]
jakedipity: fixed

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,188 @@
+# LogDNA Agent on OpenShift
+
+The agent is supported for OpenShift 4.6.
+
+## Table of Contents
```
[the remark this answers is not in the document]
jakedipity: I replicated all changes above to this file as well (where applicable)

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,188 @@
+# LogDNA Agent on OpenShift
+
+The agent is supported for OpenShift 4.6.
```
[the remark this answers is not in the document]
jakedipity: I dunno, I wasn't sure which one would be more correct either :p @tanberry @alfeng6

docs/README.md
```diff
@@ -4,73 +4,67 @@
...
+The agent is supported for Kubernetes 1.9+ and OpenShift 4.6+ environments. 
 
-The agent comes with configuration files that allow it to be deployed to a number of environments. Often, these configuration files, once set, do not change whereas the image they use can update. This introduces issues where configuration files can fall behind and not provide the up to date features that come with evolving tools. We aim to make changes in the agent backwards compatible with old configuration files, but recommend regularly ensuring your config is up to date. The agent will try to alert users when it can't use new features.
+### Installing
 
-### Upgrading on Kubernetes
+Environment specifc instructions regarding installing and deploying the agent to your cluster.
 
-We recommend doing a complete reinstall of the agent by removing the existing configuration and installing the latest one.
+* [Installing Kubernetes](KUBERNETES.md#installing)
```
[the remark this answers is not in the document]
jakedipity: Fixed the lists to include "on". Also removed the sentences. Less is more, right?

docs/README.md
```diff
@@ -4,73 +4,67 @@
...
+Environment specific instructions for removing the agent entirely.
 
-### Installing on Kubernetes
+* [Uninstalling Kubernetes](KUBERNETES.md#uninstalling)
+* [Uninstalling OpenShift](OPENSHIFT.md#uninstalling)
 
-The agent is compatible with Kubernetes clusters running `v1.9` or greater. It can be quickly and effortlessly deployed to all nodes to forward logs from your entire kubernetes cluster by running two commands:
+### Run as Non-Root
 
+By default the agent is ran as root. Below are environment specific instructions for running the agent as a non-root user.
```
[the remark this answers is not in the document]
jakedipity: Using "run" still feels awkard. Changed it to `will run as root`. Fixed the second issue.

docs/README.md
```diff
@@ -125,12 +119,14 @@ The agent accepts configuration from two sources, environment variables and a co
...
+|`LOGDNA_LOOKBACK`|The lookback strategy on startup|`smallfiles`|
+|`LOGDNA_LOG_K8S_EVENTS`|Whether the agent shoudl capture Kubernetes Events|`always`|
 
 1. We support [this flavor of globber syntax](https://github.com/CJP10/globber).
 
-### Configuring Kubernetes
+### Configuring the Environment
 
-To configure the kubernetes daemonset, copy the [logdna-agent yaml](../k8s/logdna-agent.yaml) and modify the `env` section. For example, to change the hostname add the following:
+To configure the DaemonSet, make modifications to the envs section of the DameonSet [`spec.template.spec.containers.0.env`]. For example, to change the hostname add the following environment variable to the `env` list:
```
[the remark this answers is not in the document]
jakedipity: sorry. I'm so used to trying to meet those minimum word counts :D

docs/README.md
```diff
@@ -140,14 +136,27 @@ env:
 
 Check out [Kubernetes documentation](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/) for more information about injecting environment variables into applications!
 
-### Run as Non-Root
+### Configuring Lookback
 
-Beginning with version 2.2 of the agent, the Dockerfile supports running the agent as a non-root user. This behavior; however, is not the default as to make the new version backwards compatible with old versions of the agent yamls. Make the following changes to your DaemonSet to run the agent as non-root:
+The lookback strategy determines how the agent handles existing container logs on startup. This strategy is determined by `LOGDNA_LOOKBACK`. There's a limited set of valid values for this option:
```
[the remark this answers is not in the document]
jakedipity: Removed most traces of weakness 💪

docs/README.md
```diff
@@ -140,14 +136,27 @@ env:
...
 
-Add two new fields, `runAsUser` and `runAsGroup`, to the `securityContext` section found in the `logdna-agent` container in the `logdna-agent` DaemonSet [`spec.template.spec.containers.0.securityContext`]:
+* `start` - Always start at the beginning of the files.
```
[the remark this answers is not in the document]
jakedipity: The section above, `Configuring the Environment` tells them where and how to make their changes.

docs/README.md
```diff
@@ -140,14 +136,27 @@ env:
...
+* `smallfiles` - If the file is less than 8KiB, start at the beginning. Otherwise, start at the end.
+* `none` - Always start at the end of the files.
+* __Note:__ The default option is `smallfiles`
```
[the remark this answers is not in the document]
jakedipity: You kind of have to combine the information here with the statement above: `The lookback strategy determines how the agent handles existing files on startup.`

So on startup the agent will either tail from the beginning of a file, the end of a file, or by the 8KiB heuristic.

docs/README.md
jakedipity: Makes sense to add and that's a very nice to read paragraph 👍

docs/README.md
jakedipity: fixed

docs/README.md
jakedipity: Made this change mostly, made some slight tweaks:

1. `the LogDNA agents on each pod` => `the LogDNA agent pods`
2. `the LogDNA agent on the next oldest pod` => `the next oldest LogDNA agent pod`

I understand the distinction you're trying to make. I think these changes weed out some wordiness and keep the same distinctions :). lmk what you think

docs/KUBERNETES.md
```diff
@@ -0,0 +1,202 @@
...
+## Uninstalling
+
+The default configuration places all of the Kubernetes Objects in a unique namespace. To completely remove all traces of the agent you need to simply delete this namespace:
+
+```console
+kubectl delete -f k8s/agent-namespace.yaml
+```
+
+If you're sharing the namespace with other applications, you can also remove all traces of the agent by deleting with a label filter. You'll also need to remove the logdna-agent-key secret which doesn't have a label:
+
```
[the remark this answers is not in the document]
johnpadilla: "If you can't delete the namespace because other applications are using it, you can remove..." ?

docs/KUBERNETES.md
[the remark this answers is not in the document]
johnpadilla: Agreed: "root"

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Installation Steps
+
+1. Navigate to the root directory of the cloned `logdna-agent-v2` repository.
+2. Run the following commands to create and configure a new project, secret, and service account:
+```console
+oc new-project logdna-agent
+oc create serviceaccount logdna-agent
+oc create secret generic logdna-agent-key --from-literal=logdna-agent-key=<YOUR LOGDNA INGESTION KEY>
+oc adm policy add-scc-to-user privileged system:serviceaccount:logdna-agent:logdna-agent
```
c-nixon, line 39: Just testing if we can avoid this and get away with one of the less powerful SCCs. Watch this space

c-nixon: Nope, doesn't look like it's possible, we may need a custom SCC or something. This is how fluentd does it, so at least we're not doing anything worse than they are

docs/KUBERNETES.md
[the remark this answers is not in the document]
stephen-cameron-logdna: Might we suggest users run kubectl delete -f agent-resources.yaml, that should clean up all of the serviceroles and bindings as well related to the agent.

docs/OPENSHIFT.md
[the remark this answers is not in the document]
stephen-cameron-logdna: Similiar to my notes on the readme.io doc the IBM's highest supported version they deploy right now is 4.5

docs/KUBERNETES.md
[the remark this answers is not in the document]
johnpadilla: @stephen-cameron-logdna Agreed. I looked at this from the perspective of "ok, so you've uninstalled, now here's how you finish the job." That's not a good way of looking at it though.

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
+# LogDNA Agent on OpenShift
+
+The agent is supported for OpenShift 4.6.
```
c-nixon, line 3:
```suggestion
The agent is supported for OpenShift 4.5.
```

docs/KUBERNETES.md
```diff
@@ -0,0 +1,197 @@
...
+* Local clone of this repository.
+
+### Installation Steps
+
+1. Navigate to the root directory of the cloned `logdna-agent-v2` repository.
+2. Run the following commands to configure and start the agent:
+```console
+kubectl apply -f k8s/agent-namespace.yaml
+kubectl create secret generic logdna-agent-key -n logdna-agent --from-literal=logdna-agent-key=<YOUR LOGDNA INGESTION KEY>
+kubectl apply -f k8s/agent-resources.yaml
```
stephen-cameron-logdna, line 39: Will we need to update this to the beta yamls? @c-nixon

docs/KUBERNETES.md
```diff
@@ -0,0 +1,197 @@
...
+#### Upgrading from Configuration v1.x.x or v2.0.x
+
+* **Example Configuration YAML Files:**
+  * [v1.x.x](https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-ds.yaml)
+  * [v2.0.x](https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-v2.yaml)
+* **Differences:** The configuration does not include the new logdna-agent namespace and is lacking a number of new Kubernetes objects.
+* **Upgrade Steps:**
+  1. If you have changes you want to persist to the new DaemonSet, backup the old DaemonSet.
+     1. Run `kubectl get daemonset -o yaml logdna-agent > old-logdna-agent-daemon-set.yaml`.
+     2. Copy any desired changes from `old-logdna-agent-daemon-set.yaml` to the DaemonSet object in `k8s/agent-resources.yaml`.
```
stephen-cameron-logdna, line 82: Another one for the beta yaml? @c-nixon

docs/KUBERNETES.md
```diff
@@ -0,0 +1,197 @@
...
+#### Upgrading from Configuration v2.1.x
+
+* **Example Configuration YAML Files:**
+  * [v2.1.x](https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-v2-beta.yaml)
+* **Differences:** The configuration contains the same namespace and Kubernetes objects. The only differences are some changes to the DaemonSet.
+* **Upgrade Steps:**
+  1. If you have changes you want to persist to the new DaemonSet, backup the old DaemonSet.
+     1. Run `kubectl get daemonset -o yaml -n logdna-agent logdna-agent > old-logdna-agent-daemon-set.yaml`.
+     2. Copy any desired changes from `old-logdna-agent-daemon-set.yaml` to the DaemonSet object in `k8s/agent-resources.yaml`.
+  2. Apply the latest configuration YAML file; run `kubectl apply -f k8s/agent-resources.yaml`.
```
stephen-cameron-logdna, line 97: beta yaml update @c-nixon

docs/KUBERNETES.md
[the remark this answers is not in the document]
stephen-cameron-logdna: beta yaml update @c-nixon

docs/KUBERNETES.md
```diff
@@ -0,0 +1,197 @@
...
+```yaml
+securityContext:
+  runAsUser: 5000
+  runAsGroup: 5000
+```
+
+Apply the updated configuration to your cluster:
+
+```console
+kubectl apply -f k8s/agent-resources.yaml
```
stephen-cameron-logdna, line 159: beta yaml @c-nixon

docs/KUBERNETES.md
```diff
@@ -0,0 +1,197 @@
...
+
+To enable Journald monitoring in the agent, add a new environment variable, `LOGDNA_JOURNALD_PATHS` with a value of `/var/log/journal`, to the logdna-agent DaemonSet:
+* If you are updating an already deployed agent:
+  1. You can patch the existing agent by running.
+```console
+kubectl patch daemonset -n logdna-agent logdna-agent --type json -p '[{"op":"add","path":"/spec/template/spec/containers/0/env/-","value":{"name":"LOGDNA_JOURNALD_PATHS","value":"/var/log/journal"}}]'
+```
+* If you are modifying a YAML file:
+  1. Add the new environment variable to the envs section of the DaemonSet object in `k8s/agent-resources.yaml` [`spec.template.spec.containers.0.env`].
+  2. Apply the new configuration file, run `kubectl apply -f k8s/agent-resources.yaml`.
```
stephen-cameron-logdna, line 191: beta yaml for 190 and 191 @c-nixon

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+2. Run the following commands to create and configure a new project, secret, and service account:
+```console
+oc new-project logdna-agent
+oc create serviceaccount logdna-agent
+oc create secret generic logdna-agent-key --from-literal=logdna-agent-key=<YOUR LOGDNA INGESTION KEY>
+oc adm policy add-scc-to-user privileged system:serviceaccount:logdna-agent:logdna-agent
+```
+3. Create the remaining resources:
+```console
+oc apply -f k8s/agent-resources-openshift.yaml
```
stephen-cameron-logdna, line 43: I'm assuming the openshift yamls will be tagged with beta too @c-nixon

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### Upgrading from Configuration v1.x.x
+
+* **Example Configuration YAML Files:**
+  * [v1.x.x](https://raw.githubusercontent.com/logdna/logdna-agent/master/logdna-agent-ds-os.yaml)
+* **Differences:** The configuration is lacking a number of new OpenShift objects. It also uses a removed environment variable for controlling journald monitoring, `USEJOURNALD`.
+* **Upgrade Steps:**
+  1. If you have changes you want to persist to the new DaemonSet, backup the old DaemonSet.
+     1. Run `oc get daemonset -o yaml logdna-agent > old-logdna-agent-daemon-set.yaml`.
+     2. Copy any desired changes from `old-logdna-agent-daemon-set.yaml` to the DaemonSet object in `k8s/agent-resources-openshift.yaml`.
```
stephen-cameron-logdna, line 85: beta yaml @c-nixon

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+  2. If you want to continue using journald, follow the steps for [enabling journald monitoring on the agent](#enabling-journald-monitoring-on-the-agent).
+  3. Overwrite the DaemonSet as well as create the new OpenShift objects; run `oc apply -f k8s/agent-resources-openshift.yaml`.
```
stephen-cameron-logdna, line 87: beta yaml @c-nixon

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+oc delete serviceaccount logdna-agent
+```
+
+## Run as Non-Root
+
+By default the agent is configured to run as root; however, the DaemonSet can be modified to run the agent as a non-root user.
+
+This is accomplished through Linux capabilities and turning the agent binary into a "capability-dumb binary." The binary is given `CAP_DAC_READ_SEARCH` to read all files on the file system. The image already comes with this change and the necessary user and group. The only required step is configuring the agent DaemonSet to run as the user and group `5000:5000`.
+
+Add two new fields, `runAsUser` and `runAsGroup`, to the `securityContext` section found in the `logdna-agent` container in the `logdna-agent` DaemonSet inside of `k8s/agent-resources-openshift.yaml` [`spec.template.spec.containers.0.securityContext`]:
```
stephen-cameron-logdna, line 138: beta yaml @c-nixon

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+```yaml
+securityContext:
+  runAsUser: 5000
+  runAsGroup: 5000
+```
+
+Apply the updated configuration to your cluster:
+
+```console
+oc apply -f k8s/agent-resources-openshift.yaml
```
stephen-cameron-logdna, line 149: beta yaml @c-nixon

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+
+To enable Journald monitoring in the agent, add a new environment variable, `LOGDNA_JOURNALD_PATHS` with a value of `/var/log/journal`, to the logdna-agent DaemonSet:
+* If you are updating an already deployed agent:
+  1. You can patch the existing agent by running.
+```console
+oc patch daemonset -n logdna-agent logdna-agent --type json -p '[{"op":"add","path":"/spec/template/spec/containers/0/env/-","value":{"name":"LOGDNA_JOURNALD_PATHS","value":"/var/log/journal"}}]'
+```
+* If you are modifying a YAML file:
+  1. Add the new environment variable to the envs section of the DaemonSet object in `k8s/agent-resources-openshift.yaml` [`spec.template.spec.containers.0.env`].
+  2. Apply the new configuration file, run `oc apply -f k8s/agent-resources-openshift.yaml`.
```
stephen-cameron-logdna, line 178: 177 and 178 beta yaml @c-nixon

c-nixon: These ones refer to github, so they should be ok, I think it's only the cdn ones that we'll have issues with

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
+# LogDNA Agent on OpenShift
+
```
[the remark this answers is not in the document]
tanberry: @c-nixon please add, at very top, the Beta tag:  [Public Beta](#)

Also please add the registered trademark symbol after the first instance of OpenShift using "OpenShift&reg"

docs/KUBERNETES.md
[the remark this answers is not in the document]
c-nixon: Nope, we're good as it's referring to the one in the repo

docs/README.md
```diff
@@ -140,14 +130,31 @@ env:
...
+
+Take a look at enabling journald monitoring for [Kubernetes](KUBERNETES.md#collecting-node-journald-logs) or [OpenShift](OPENSHIFT.md#collecting-node-journald-logs).
+
+### Configuring Events
+
+A Kubernetes event is exactly what it sounds like: a resource type that is automatically generated when state changes occur in other resources, or when errors or other messages manifest across the system. Monitoring events is useful for debugging your Kubernetes cluster.
+
+By default, the LogDNA agent captures Kubernetes events (and OpenShift events, as well, since OpenShift is built on top of Kubernetes clusters).
+
+To control whether the LogDNA agent collects Kubernetes events, configure the `LOGDNA_LOG_K8s_EVENTS` option using on of these two values:
```
c-nixon, line 154:
```suggestion
To control whether the LogDNA agent collects Kubernetes events, configure the `LOGDNA_LOG_K8s_EVENTS` environment veriable using on of these two values:
```

docs/README.md
c-nixon: Yeh, it's in the k8s yaml env section, I think this should be fairly intuitive based on how the rest of the configuration is done

docs/OPENSHIFT.md
c-nixon: It always need to run as privileged, non-root or not, it's what openshift requires if you want to access the host

docs/OPENSHIFT.md
```diff
@@ -0,0 +1,184 @@
...
+```console
+oc api-resources --verbs=list --namespaced -o name | xargs -n 1 oc get --show-kind --ignore-not-found -l app.kubernetes.io/name=logdna-agent -o name | xargs oc delete
+oc delete secret logdna-agent-key
+oc delete serviceaccount logdna-agent
+```
+
+## Run as Non-Root
+
+By default the agent is configured to run as root; however, the DaemonSet can be modified to run the agent as a non-root user.
+
```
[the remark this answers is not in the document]
c-nixon: You can also modify the DaemonSet directly on the cluster, without touching the YAML file, so I think it's better to leave that bit out

PR #43, opened 2020-11-10
Fix bug where Journald ingests stale records

jakedipity:
Found an issue with Journald that occurs in the following scenario

1. The agent is already started and monitoring a journal such as `/var/log/journal`
2. The node does not currently have Journald logging enabled
3. Modifying the node to enable journald which results in the creation of a new subdirectory `/var/log/journal/<GUID>`
4. The agent will start going through all the stale journald records from the beginning and giving them the false timestamp of `now`

This issue occurs because we only seek the tail of the Journal stream on startup. This fix checks to make sure all records have a timestamp of within 30 seconds. If a timestamp is older than that, then the journal is reseeked to the end.

```
[2020-11-10T10:10:52Z WARN  journald::stream] Received a stale journald record, reseeking pointer
```

### DIFF
--- a/common/journald/src/stream.rs
+++ b/common/journald/src/stream.rs
@@ -1,8 +1,7 @@
 use crate::error::JournalError;
-use chrono::{Local, TimeZone};
 use futures::{channel::oneshot, stream::Stream as FutureStream};
 use http::types::body::LineBuilder;
-use log::warn;
+use log::{info, warn};
 use metrics::Metrics;
 use std::{
     mem::drop,
@@ -14,7 +13,7 @@ use std::{
     },
     task::{Context, Poll, Waker},
     thread::{self, JoinHandle},
-    time::{Duration, UNIX_EPOCH},
+    time::{Duration, SystemTime},
 };
 use systemd::journal::{Journal, JournalFiles, JournalRecord, JournalSeek};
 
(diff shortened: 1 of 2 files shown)


common/journald/src/stream.rs
```diff
@@ -208,28 +207,35 @@ impl Reader {
...
-                .timestamp(timestamp.as_secs() as i64, 0)
-                .format("%b %d %H:%M:%S")
-                .to_string(),
-            Ok(Err(_)) | Err(_) => {
-                warn!("unable to read timestamp associated with journald record");
-                Local::now().format("%b %d %H:%M:%S").to_string()
+            Some(duration) => {
+                // Reject any records with a timestamp older than 30 seconds
+                if duration >= Duration::from_secs(30) {
+                    warn!("Received a stale journald record, reseeking pointer");
```
jakedipity, line 220: Do you think this should be a warn or just an info?

jakedipity:
@logdnabot test this

common/journald/src/stream.rs
c-nixon: I think an info for now. Looking at https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html I think there might be some combination of fields we can use to do this more precisely, but I think this makes sense until we add something like lookback for journald

common/journald/src/stream.rs
```diff
@@ -208,28 +207,35 @@ impl Reader {
             Err(e) => return Err(JournalError::BadRead(e)),
         };
 
-        let timestamp = match self
+        match self
             .reader
             .timestamp()
-            .map(|timestamp| timestamp.duration_since(UNIX_EPOCH))
+            .ok()
+            .map(|timestamp| SystemTime::now().duration_since(timestamp).ok())
```
c-nixon, line 214: Clippy will probably complain about the map followed by flatten

jakedipity: Did fail, but because of a cargo fmt issue with a comment. Fixed now

jakedipity:
@logdnabot test this

c-nixon (dismissed):
This makes sense to me for our first release with Journald support. LGTM

common/journald/src/stream.rs
jakedipity: Updated to info!

jakedipity:
@logdnabot test this

jakedipity:
@logdnabot test this

c-nixon (approved):

PR #45, opened 2020-11-13
Add OpenShift YAML files

jakedipity:

### DIFF
--- a/k8s/agent-resources-openshift.yaml
+++ b/k8s/agent-resources-openshift.yaml
@@ -1,0 +1,164 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: logdna-agent
+  name: logdna-agent
+  labels:
+    app.kubernetes.io/name: logdna-agent
+    app.kubernetes.io/instance: logdna-agent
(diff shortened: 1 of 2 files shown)


k8s/agent-resources-openshift.yaml
```diff
@@ -0,0 +1,156 @@
...
+        app.kubernetes.io/instance: logdna-agent
+        app.kubernetes.io/version: 2.1.9-beta.2
+    spec:
+      serviceAccountName: logdna-agent
+      containers:
+        - name: logdna-agent
+          image: logdna/logdna-agent:stable
+          imagePullPolicy: Always
+          securityContext:
+            privileged: true
```
c-nixon, line 98: Is it possible for us to reduce this to "hostmount-anyuid"?  https://www.openshift.com/blog/managing-sccs-in-openshift.

We might be able to look into this alongside run as non-root

c-nixon: I'm asking this is the wrong place, I think we should be able to remove this since we grant the service account the privileged SSC in the docs

c-nixon: And nope, looks like we need a custom SCC

c-nixon (approved):

PR #55, opened 2020-12-17
Add performance regression testing to Jenkins pipeline

jorgebay:
It adds an optional stage to the build to run performance regression testing. This approval process has a 30s timeout and if it's not approved, the build can continue without reporting a failure.

The stage generates a memory series ([like this one](https://agent-benchmarks-results.s3.us-east-2.amazonaws.com/perf-regression-testing-27/memory-series.png)) chart and CPU histogram files that are linked in the stage output. Hopefully, CPU histograms [will be charts in the future too](https://github.com/logdna/agent-benchmarks/issues/6).

cc @jmoses @c-nixon

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,106 +38,110 @@ pipeline {
                 sh "docker pull ${RUST_IMAGE_REPO}:${RUST_IMAGE_TAG}"
             }
         }
-        parallel {
-            stage ("Lint and Test"){
-                environment {
-                    CREDS_FILE = credentials('pipeline-e2e-creds')
-                    LOGDNA_HOST = "logs.use.stage.logdna.net"
-                }
-                steps {
-                    script {
-                        def creds = readJSON file: CREDS_FILE
-                        // Assumes the pipeline-e2e-creds format remains the same. Chase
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -68,6 +69,90 @@ pipeline {
...
+                                    # Install terraform, it should be moved to the image
+                                    if ! command -v terraform &> /dev/null
+                                    then
+                                        echo "Installing terraform"
+                                        curl -o terraform.zip "https://releases.hashicorp.com/terraform/0.14.2/terraform_0.14.2_linux_amd64.zip"
+                                        unzip terraform.zip
+                                        sudo mv terraform /usr/local/bin/
+                                        terraform --version
+                                    fi
+                                ''', label: 'Install terraform'
```
jorgebay, line 116: This should be removed [once terraform is installed in the image](https://github.com/answerbook/jenkins/pull/156).

jorgebay:
Given the different set of configurations we want to test (with or without redaction / lookback / ...), I think its better to launch the load tests manually for now. We can integrate it with jenkins when we have a better picture of what we need.

PR #58, opened 2020-12-21
Test file appended in the background

jorgebay:
Generate a file that gets appended in the background and create an integration test that validates that the data is read by the agent.

For `start_append_to_file` fn in the tests, I've ended up using a separate thread for the file appends and use sync IO because I find it easier than doing it w/ tokio, given that is a test setup I hope its ok.

### DIFF
--- a/bin/tests/common.rs
+++ b/bin/tests/common.rs
@@ -1,0 +1,80 @@
+use assert_cmd::cargo::CommandCargoExt;
+use core::time;
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::path::PathBuf;
+use std::process::{Child, Command, Stdio};
+use std::sync::mpsc;
+use std::sync::mpsc::TryRecvError;
+use std::thread;
+
+pub struct FileContext {
+    pub file_path: PathBuf,
+    pub stop_handle: Box<dyn FnOnce() -> i32>,
+}
+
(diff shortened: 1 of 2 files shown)


bin/tests/common.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::sync::mpsc;
+use std::sync::mpsc::TryRecvError;
+use std::thread;
+
+pub struct FileContext {
+    pub file_path: PathBuf,
+    pub stop_handle: Box<dyn FnOnce() -> i32>,
+}
+
+pub fn start_append_to_file(dir: &PathBuf, delay_ms: u64) -> FileContext {
```
c-nixon, line 16: It's a little nicer to use a Path rather than a pathbuf for function parameters if you don't need it to be mutable.

It's trivial to convert a PathBuf to a path, but converting. Path to a PathBuf requires an allocation and copy (like &str to String)

c-nixon (dismissed):
This looks good to me. Minor nit around passing &PathBuf vs Path, but it's not blocking, just a style note for future

bin/tests/common.rs
jorgebay: Makes sense, I'll change it.

c-nixon (approved):
LGTM!

PR #60, opened 2020-12-22
Add direct resource links to k8s resource files (v2.2)

jondkelley:

This PR covers the 2.2 agent / branch.


c-nixon (approved):
This looks good to me, We'll have to repeat the exercise each time we cut a minor release branch, but I think it's worthwhile!

PR #61, opened 2020-12-22
Add direct resource links to k8s resource files (v3.0)

jondkelley:

This PR covers the 3.0 agent / branch.


c-nixon (approved):
LGTM

PR #62, opened 2020-12-22
Add direct resource links to k8s resource files (v2.2-nonroot)

jondkelley:

This PR covers the 2.2 agent / branch.


c-nixon (commented):
The 2.2-nonroot branch has been deprecated now that we have 2.2 and 3.0, so I'm going to close this PR and clean up those old branches, sorry for the confusion!

PR #63, opened 2020-12-22
Integration tests for moving and removing files

jorgebay:
Adds integration tests for:
- [x] file appended, deleted and created again.
- [x] file moved and created again.
- ~directory moved and created again~
- [x] dangling symlinks 
- [x] logrotate copytruncate

Note that it uses #58 as base branch.


Makefile
```diff
@@ -98,7 +98,7 @@ test: test-journald ## Run unit tests
 
 .PHONY:integration-test
 integration-test: ## Run integration tests
-	$(RUST_COMMAND) "--env LOGDNA_INGESTION_KEY=$(LOGDNA_INGESTION_KEY) --env LOGDNA_HOST=$(LOGDNA_HOST) --env RUST_BACKTRACE=full" "cargo test --manifest-path bin/Cargo.toml --features integration_tests -- --nocapture"
+	$(RUST_COMMAND) "--env LOGDNA_INGESTION_KEY=$(LOGDNA_INGESTION_KEY) --env LOGDNA_HOST=$(LOGDNA_HOST) --env RUST_BACKTRACE=full" "cargo test --manifest-path bin/Cargo.toml --features integration_tests -- --nocapture --test-threads=1"
```
c-nixon, line 101: Hmm, what errors were you running in to running the tests in parallel?

jorgebay: No errors, it's just that starting multiple agents in parallel and trying to debug/see output makes everything harder.

Do you prefer to continue having integration tests in parallel?

c-nixon: If we can run them in parallel I'd prefer to, Maybe we should paramaterise it in the makefile?

jorgebay: will do.

jorgebay:
I've removed the failing test in 9a3b3d4 that was surfacing an issue with unsafe pointer usage in `fs` package.

We'll work on that on a separate branch.

c-nixon (approved):
This is great. LGTM

PR #64, opened 2021-01-08
Test removing linked nested directories

jorgebay:
Add an integration test that uses nested directories w/ a symlink to the log directory that is removed.

Based on #63. 

This currently fails on master awaiting for the LOG-7968 fix.


jorgebay:
Rebased it against `master` after #63 was merged.

c-nixon (approved):
Excellent, I'll try and duplicate the logic in a unit test for the fs module too as a regression test

PR #65, opened 2021-01-08
Make publishing image Jenkins step optional

jorgebay:
With this change, the Jenkins step waits for a few minutes for user input and in case
publishing canceled or timed out, the build is still marked as successful.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,7 @@
 library 'magic-butler-catalogue'
 def PROJECT_NAME = 'logdna-agent-v2'
 def TRIGGER_PATTERN = '.*@logdnabot.*'
+def publishImage = false
 
 pipeline {
     agent any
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -95,10 +96,23 @@ pipeline {
                 branch pattern: "\\d\\.\\d.*", comparator: "REGEXP"
             }
             stages {
+                stage('Check Publish Image or Timeout') {
+                    steps {
+                        script {
+                            publishImage = true
+                            try {
+                                timeout(time: 5, unit: 'MINUTES') {
```
darinspivey, line 104: This is awesome, thank you.  It'll free up runners.

charlesthomas (approved):
i think if the timeout is this short, a slack alert would be a good thing to add. i don't think that needs to block this, though. it could be a separate PR

c-nixon:
@charlesthomas Slack alert on push or on timeout? Or both?

charlesthomas:
oh, sorry for not being more clear. I was thinking an alert saying "hey this job is now pending a human approval, and will timeout in 5 minutes without it"

c-nixon:
Ahh, cool, something like https://github.com/answerbook/pipeline-e2e/blob/baab9e7e1f222cdd169b83087e5996b2ab4ef5fc/Jenkinsfile#L94 ?

charlesthomas:
> Ahh, cool, something like https://github.com/answerbook/pipeline-e2e/blob/baab9e7e1f222cdd169b83087e5996b2ab4ef5fc/Jenkinsfile#L94 ?

yes, although that example is firing at the end of the job. i think for this you'd want the `notifySlack` between lines 100 and 101

PR #66, opened 2021-01-11
Fix smallvec security vulnerability

jorgebay:
Based on https://rustsec.org/advisories/RUSTSEC-2021-0003

Bumped the versions on both the manifest and lock files (newer non-breaking `kube` version).


c-nixon (approved):
This looks good. We can do a full cargo update as part of 3.1

PR #71, opened 2021-01-12
Integration test for exclusion rules

jorgebay:
Cover exclusion rules with an integration test.

### DIFF
--- a/bin/tests/cli.rs
+++ b/bin/tests/cli.rs
@@ -1,7 +1,6 @@
 use crate::common::AgentSettings;
 use assert_cmd::prelude::*;
 use predicates::prelude::*;
-use regex::Regex;
 use std::fs;
 use std::fs::File;
 use std::io::{BufRead, BufReader, Read, Write};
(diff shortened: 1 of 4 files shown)


bin/tests/common.rs
```diff
@@ -87,7 +87,22 @@ pub fn truncate_file(file_path: &PathBuf) -> Result<(), std::io::Error> {
     Ok(())
 }
 
-pub fn spawn_agent(dir_path: &str) -> Child {
+#[derive(Default)]
+pub struct AgentSettings<'a> {
+    pub log_dirs: &'a str,
+    pub exclusion_regex: &'a str,
```
c-nixon, line 93: Minor nit: It's a bit more idiomatic to use an Option for fields that may or not be present, rather than checking for the empty str

c-nixon (approved):
Very minor style nit, otherwise LGTM!

bin/tests/common.rs
jorgebay: makes sense, I'll change it.

bin/tests/cli.rs
```diff
@@ -215,6 +214,37 @@ fn test_truncate_file() {
...
+
+    let mut agent_handle = common::spawn_agent(AgentSettings {
+        log_dirs: &dir.to_str().unwrap(),
+        exclusion_regex: r"\w+ile2\.\w{3}",
+    });
+
+    let mut stderr_reader = BufReader::new(agent_handle.stderr.as_mut().unwrap());
+
+    let lines = common::wait_for_file_event("initialized", &included_file, &mut stderr_reader);
+    let re = Regex::new(r"initialized [^\n]*file2\.log").unwrap();
```
c-nixon, line 234: One other minor thing is that the predicate crate has Regex predicates, ie
https://docs.rs/predicates/1.0.6/predicates/str/struct.RegexPredicate.html which would save the additional crate import, but again very minor.

c-nixon (approved):

PR #76, opened 2021-01-21
Replace sys-info dependency

jorgebay:
https://rustsec.org/advisories/RUSTSEC-2020-0100.html

Ref: LOG-8600


c-nixon (approved):
And it fixes a potential issue with dodgy hostname. LGTM

PR #77, opened 2021-01-21
Use hostname package to instead of libc

c-nixon:
This cleans up some more unsafe usage in the config module, in this case it was causing a use-after-free in my local tests.


jorgebay:
We can use the new `sysinfo` crate in #76 and have one less dependency

We can add the logic there (`sys.get_host_name()`), wdyt?

jakedipity (approved):
Less packages; love it.

PR #78, opened 2021-01-22
feat: Add 'auditable' to agent so we can track the dep versions

c-nixon:
Bonus PR while I'm waiting for builds.

This adds support for https://github.com/Shnatsel/rust-audit, which includes cargo lock metadata in the agent binaries themselves so that we can check the exact versions that were compiled into any given binary.

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -13,6 +13,7 @@ name = "logdna-agent"
 version = "3.0.0-beta.1"
 authors = ["CJP10 <connor.peticca@logdna.com>"]
 edition = "2018"
+build = "build.rs"
 
 [[bin]]
 name = "logdna-agent"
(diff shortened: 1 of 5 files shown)


bin/Cargo.toml
```diff
@@ -30,12 +31,16 @@ journald = { package = "journald", path = "../common/journald" }
...
 env_logger = "0.7"
+anyhow = "1.0"
 serde_yaml = "0.8"
 jemallocator = "0.3"
 futures = "0.3"
 tokio = { version = "0.2", features = ["rt-threaded"] }
 pin-utils = "0.1"
 
+auditable = "0.1"
+miniz_oxide = "0.4"
```
jorgebay, line 42: Is this used?

c-nixon: I forgot to add a file :)

bin/src/main.rs
```diff
@@ -42,6 +44,10 @@ fn main() {
     env_logger::from_env(Env::default().default_filter_or("info")).init();
     info!("running version: {}", env!("CARGO_PKG_VERSION"));
 
+    // Actually use the data to work around a bug in rustc:
+    // https://github.com/rust-lang/rust/issues/47384
+    trace!("{:?}", dep_audit::get_auditable_dependency_list());
```
jorgebay, line 49: We are sending to output the value wrapped in the Result, should we unwrap it?

jorgebay (commented):
Nice!

I've added some questions below.

jorgebay (approved):

PR #79, opened 2021-01-25
Use recoverable errors in fs cache

jorgebay:
This patch replaces `unwrap()` and `expect()` calls that would cause thread panic with recoverable errors.

Ref: LOG-8421.

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -269,20 +269,19 @@ where
         // directories can't be a hard link so we're guaranteed the watch descriptor maps to one
         // entry
         let entry_key = self.get_first_entry(watch_descriptor)?;
-        if let Some(entry) = _entries.get(entry_key) {
-            let mut path = self.resolve_direct_path(&entry.borrow(), _entries);
-            path.push(name);
+        let entry = _entries.get(entry_key).ok_or("failed to find entry")?;
+        let mut path = self.resolve_direct_path(&entry.borrow(), _entries);
+        path.push(name);
 
+        if let Some(new_entry) = self.insert(&path, events, _entries)? {
             let mut errors = vec![];
-            if let Some(new_entry) = self.insert(&path, events, _entries)? {
-                if matches!(_entries.get(new_entry).map(|n_e| n_e.borrow()), Some(_)) {
-                    for new_path in recursive_scan(&path) {
-                        if let Err(e) = self.insert(&new_path, events, _entries) {
-                            errors.push(e);
-                        }
+            if matches!(_entries.get(new_entry).map(|n_e| n_e.borrow()), Some(_)) {
+                for new_path in recursive_scan(&path) {
+                    if let Err(e) = self.insert(&new_path, events, _entries) {
+                        errors.push(e);
                     }
-                };
-            }
+                }
+            };
 
             if !errors.is_empty() {
                 return Err(format!(
@@ -290,11 +289,9 @@ where
                     errors.join(";")
                 ));
             }
-
-            Ok(())
-        } else {
-            Err("failed to find entry".into())
         }
+
+        Ok(())
     }
 
     fn process_modify(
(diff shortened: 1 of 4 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -1001,10 +937,37 @@ where
         false
     }
 
-    // a helper for checking if a path passes exclusion/inclusion rules
+    /// Helper method for checking if a path passes exclusion/inclusion rules
     fn passes(&self, path: &PathBuf, _entries: &EntryMap<T>) -> bool {
         self.is_initial_dir_target(path) || self.is_symlink_target(path, _entries)
     }
+
+    fn entry_path_passes(&self, entry: EntryKey, name: OsString, _entries: &EntryMap<T>) -> bool {
```
c-nixon, line 945: We could cut down some copies if name was an OsStr as it impl's AsRef<Path>

common/fs/src/cache/mod.rs
```diff
@@ -199,66 +214,36 @@ where
...
+                    self.process_create(&to_wd, to_name, events, &mut _entries)
                 } else if is_from_path_ok {
-                    self.process_delete(&from_wd, from_name, events, &mut _entries);
+                    self.process_delete(&from_wd, from_name, events, &mut _entries)
+                } else {
+                    Err("Moving paths are invalid".into())
                 }
             }
-            WatchEvent::Overflow => panic!("overflowed kernel queue!"),
+            WatchEvent::Overflow => Err(
```
c-nixon, line 237: Hmm, is this recoverable? I'm not actually sure of the implications, maybe we should dump the fs cache and repopulate?

I think panic'ing is sensible here until we handle it more gracefully as the pod restart should get things back into a clean state.

common/fs/src/cache/mod.rs
```diff
@@ -35,6 +35,7 @@ type WatchDescriptors = HashMap<WatchDescriptor, Vec<EntryKey>>;
 pub type EntryKey = DefaultKey;
 
 type EntryMap<T> = SlotMap<EntryKey, RefCell<entry::Entry<T>>>;
+type FsResult<T> = Result<T, String>;
```
c-nixon, line 38: By having this as a String it flattens a few different kinds of errors into one(String), so we lose some information if we wanted to handle it higher up.

Could we add an FsCacheError enum and make this type FsResult<T> = Result<T, FsCacheError> ?

The thiserror crate makes creating these kinds of strongly typed errors very easy

c-nixon: For examples of different kinds of errors I think there are Watch, WatchEvent, CacheLookup , InvalidFsObject at least, there may be more in there. It would be useful to be able to tell them apart

c-nixon (changes requested):
Very nice refactoring of the duplicated code, and yay for more test cases!

Just a couple of minor things and a question about the WatchEvent overflow

common/fs/src/cache/mod.rs
jorgebay: Makes sense, I'll change it to use an enum.

common/fs/src/cache/mod.rs
jorgebay: I'm also in doubt of what to do in this situation...
Queue overflow might indicate a third party system that's misbehaving and `fsync`ing lots of small changes in lots of files or a ridiculously low value in `/proc/sys/fs/inotify/max_queued_events`.

I think that logging at error level acknowledging the issue is less aggressive than exiting the process. We could add improvements like sending an ERROR log message to logdna servers for operators to easily pick up on them.

wdyt? Should I create a ticket for an improvement? Should I revert to panic?

common/fs/src/cache/mod.rs
jorgebay: I've added an enum to represent the different errors.

common/fs/src/cache/mod.rs
c-nixon: We havn't had any reports of this panic happening, maybe we could create the err, log it and then immediately .expect it as an interim measure until we handle it properly?

I think handling it properly involves recreating the fs cache and the tailers from the last known good position which will be possible when I get the stateful lookback set up.

jorgebay: > We haven't had any reports of this panic happening

I agree that this is a rare edge case.

> .expect it as an interim measure until we handle it properly

ok, I'll update the patch to log at error level and panic.

common/fs/src/cache/mod.rs
```diff
@@ -35,6 +35,46 @@ type WatchDescriptors = HashMap<WatchDescriptor, Vec<EntryKey>>;
 pub type EntryKey = DefaultKey;
 
 type EntryMap<T> = SlotMap<EntryKey, RefCell<entry::Entry<T>>>;
+type FsResult<T> = Result<T, Error>;
+
+#[derive(std::fmt::Debug)]
+pub enum Error {
```
c-nixon, line 41: Very nice.

Would you mind redoing this using thiserror's proc macro? You can see examples in the k8s middleware. It derives std::error::Error and various other convenience things.

It should look something like
``` rust
use thiserror::Error;

#[derive(Clone, Debug, Error)]
pub enum K8sError {
    #[error("pod missing {0}")]
    PodMissingMetaError(&'static str),
    #[error("failed to initialize kubernetes middleware {0}")]
    InitializationError(String),
}


#[derive(Clone, Debug, Error)]
pub enum Error {
    #[error("error watching: {0:?} {1:?}")]
    Watch(PathBuf, io::Error),
    #[error("got event for untracked watch descriptor: {0:?}")]
    WatchEvent(WatchDescriptor),
    #[error("the inotify event queue has overflowed and events have presumably been lost")]
    WatchOverflow,
    #[error("unexpected existing entry")]
    Existing,
    #[error("failed to find entry")]
    Lookup,
    #[error("failed to find parent entry")]
    ParentLookup,
    #[error("parent should be a directory")]
    ParentNotValid,
    #[error("path is not valid")]
    PathNotValid,
    #[error("encountered errors when inserting recursively: {0:?}")]
    InsertRecursively(Vec<Error>),
}
```

jorgebay: oh! TIL!

Thanks for pointing it out, I'll update it.

c-nixon (approved):
💯

PR #82, opened 2021-01-27
docs: installing with helm

jorgebay:
Documentation page for installing the agent with helm.

I've rearranged the content from the main README to add the sections "manual install" (manifest files) or "helm install".


jorgebay:
I'll create a follow up ticket for supporting setting the user `runAsUser` in the helm charts and test deployment with helm on OpenShift (and probably mark helm deployments as the recommended route for new users).

docs/HELM.md
```diff
@@ -0,0 +1,79 @@
...
+
+Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
+
+```bash
+$ helm install --set logdna.key=LOGDNA_INGESTION_KEY,logdna.tags=production my-release logdna/agent
+```
+
+Alternatively, a YAML file that specifies the values for the above parameters can be provided while installing the
+chart. For example,
+
```
tanberry, line 62: To avoid "passive voice" consider this rewrite:
"Alternatively, create a YAML file called `values.yaml` that specifies the values for the above parameters.  This file is then accessed and read during the chart installation."

docs/README.md
```diff
@@ -83,7 +71,7 @@ More information about managing your deployments is documented for [Kubernetes](
 * Version specific upgrade paths
 * Collecting system logs through Journald
 
-## Building Agent v2
+## Building
```
tanberry, line 75: How about "Building the LogDNA Agent" ?? Just the single word "Building" isn't informative enough, imo.

tanberry, line 75: How about "Building the LogDNA Agent" ?? Just the single word "Building" isn't informative enough, imo.

tanberry (changes requested):
Nice docs, @jorgebay !

docs/HELM.md
```diff
@@ -0,0 +1,79 @@
+# Deploying the LogDNA Agent on Kubernetes using Helm
+
+[Helm][helm] is a package manager for Kubernetes, you can use it to deploy the LogDNA Agent in your Kubernetes
+cluster. 
+
+## Prerequisites
+
+- [Helm 3+][helm-install]
+- Kubernetes 1.13+
+
```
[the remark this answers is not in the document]
alfeng6: I think it is fine because folks on this page will know that the info here is most updated to the requirements for helm specifically.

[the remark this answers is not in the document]
jorgebay: Good question @tanberry, I've added K8s 1.13+ because is the minimum version Helm 3 supports.

docs/HELM.md
```diff
@@ -0,0 +1,79 @@
...
+## Installing the Chart
+
```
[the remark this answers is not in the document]
jorgebay: makes sense, I'll add it.

docs/README.md
jorgebay: The reason I've changed the title is because the table of contents was linking to a `#building` anchor, so I changed the title. I'll change it to "Building the LogDNA Agent" and update the anchor.

jorgebay:
I've just pushed some changes to the PR to address the highlighted issues.

alfeng6 (approved):
Looks great to me!

c-nixon (approved):
This looks great to me

PR #83, opened 2021-01-27
Client 0.3 refactor

c-nixon:
This ports the agent over to the prerelease version of the rust client library.

It also adds the mock ingester and some more integration tests.

There is further work to do to take advantage of the client's new buffer pooling and streaming serialization capabilities, but this is a good first step.


bin/tests/cli.rs
```diff
@@ -391,3 +398,194 @@ fn test_directory_symlinks_delete() {
...
+    let key = pemfile::pkcs8_private_keys(&mut cert.serialize_private_key_pem().as_bytes())
+        .expect("couldn't load rsa_private_key");
+    let addr = "0.0.0.0:1337".parse().unwrap();
+
+    let mut cert_file = tempfile::NamedTempFile::new().expect("Couldn't create cert file");
+    cert_file
+        .write_all(cert.serialize_pem().unwrap().as_bytes())
+        .expect("Couldn't write cert file");
+
+    let (server, received, shutdown_handle) = https_ingester(addr, certs, key[0].clone());
```
jorgebay, line 421: This could be extracted to a reusable method and/or use https once, the rest of the test go through the http API.

bin/tests/cli.rs
```diff
@@ -391,3 +398,194 @@ fn test_directory_symlinks_delete() {
...
+
+    let dir = tempdir().expect("Couldn't create temp dir...");
+
+    let mut cmd = Command::cargo_bin("logdna-agent").unwrap();
+
+    let dir_path = format!("{}/", dir.path().to_str().unwrap());
+
+    let agent = cmd
```
jorgebay, line 429: We could reuse `common::spawn_agent()` here.

bin/tests/cli.rs
```diff
@@ -391,3 +398,194 @@ fn test_directory_symlinks_delete() {
...
```
jorgebay, line 526: Same here, we could reuse `spawn_agent()`

common/test/mock-ingester/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+                                OpenOptions::new()
+                                    .create(true)
+                                    .append(true)
+                                    .open(&file_name)
+                                    .unwrap(),
+                            )
+                        });
+
+                    line_count.fetch_add(1, Ordering::Relaxed);
+                    file.write_all(raw_line.as_bytes()).unwrap();
```
jorgebay, line 127: What's the reason to use a file? should we store it in memory instead?

We could use some kind of mutex and just append in memory (just a few kbs/mbs).

jorgebay (commented):

common/http/src/client.rs
```diff
@@ -89,7 +82,7 @@ impl Client {
         self.last_retry.elapsed() > Duration::from_secs(3)
     }
 
-    fn flush(&mut self) {
+    async fn flush(&mut self) {
```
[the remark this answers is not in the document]
c-nixon: Because flush takes a &mut self it's impossible for multiple threads to call flush at the same time as the borrow checker disallows the creation of multiple mutable references. With flush being private it's also impossible for an external caller to interleave send and flush incorrectly, so I think the type system eliminates the race condition here

common/test/mock-ingester/src/lib.rs
c-nixon: The file logic was lifter directly from the original mock ingester code, we could quite easily discard the content for now, or simply append to a Vec of Bytes, any thoughts?

common/http/src/client.rs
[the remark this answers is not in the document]
jorgebay: oh true!
that's a nice guarantee by compiler!

jorgebay (approved):
Looks great :shipit:

jorgebay:
Good idea to pick a random port.

PR #84, opened 2021-01-29
Remove entry after delete event

jorgebay:
Ref: LOG-8681

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,8 +99,8 @@ test: test-journald ## Run unit tests
 	$(RUST_COMMAND) "--env RUST_BACKTRACE=full" "cargo test --no-run && cargo test"
 
 .PHONY:integration-test
-integration-test: ## Run integration tests
-	$(RUST_COMMAND) "--env LOGDNA_INGESTION_KEY=$(LOGDNA_INGESTION_KEY) --env LOGDNA_HOST=$(LOGDNA_HOST) --env RUST_BACKTRACE=full" "cargo test --manifest-path bin/Cargo.toml --features integration_tests -- --nocapture --test-threads=$(INTEGRATION_TEST_THREADS)"
+integration-test: ## Run integration tests using image with additional tools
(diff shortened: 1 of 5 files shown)


jorgebay:
I'm trying to come up w/ a integration test to cover it, I think I can have it before EOD.

c-nixon (approved):
Very nice, this is definitely part of the symlink issue from before. I look forward to seeing how you test this! 😄

Makefile
```diff
@@ -99,8 +99,8 @@ test: test-journald ## Run unit tests
 	$(RUST_COMMAND) "--env RUST_BACKTRACE=full" "cargo test --no-run && cargo test"
 
 .PHONY:integration-test
-integration-test: ## Run integration tests
-	$(RUST_COMMAND) "--env LOGDNA_INGESTION_KEY=$(LOGDNA_INGESTION_KEY) --env LOGDNA_HOST=$(LOGDNA_HOST) --env RUST_BACKTRACE=full" "cargo test --manifest-path bin/Cargo.toml --features integration_tests -- --nocapture --test-threads=$(INTEGRATION_TEST_THREADS)"
+integration-test: ## Run integration tests using image with additional tools
+	$(DOCKER_JOURNALD_DISPATCH) "--env LOGDNA_INGESTION_KEY=$(LOGDNA_INGESTION_KEY) --env LOGDNA_HOST=$(LOGDNA_HOST) --env RUST_BACKTRACE=full" "cargo test --manifest-path bin/Cargo.toml --features integration_tests -- --nocapture --test-threads=$(INTEGRATION_TEST_THREADS)"
```
jorgebay, line 103: I'll create a follow up ticket to rename the journald scripts and dockerfile folder to something like "dev"

jorgebay:
I've pushed a test using `lsof` and verified that it fails against `master`.

docker/journald/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM docker.io/logdna/build-images:rust-buster-stable
-RUN apt update -y && apt install -y systemd
+RUN apt update -y && apt install -y systemd lsof procps
```
c-nixon, line 2: We should add these to the base image going forward

c-nixon (approved):
This is great

docker/journald/Dockerfile
jorgebay: oh, TIL about https://github.com/logdna/build-images

Good call, I'll move it there as part of the follow up ticket.

PR #87, opened 2021-02-01
Add journald integration test

jorgebay:
Ref: LOG-8374

### DIFF
--- a/bin/tests/cli.rs
+++ b/bin/tests/cli.rs
@@ -5,7 +5,6 @@ use predicates::prelude::*;
 use std::fs;
 use std::fs::File;
 use std::io::{BufRead, BufReader, Read, Write};
-use std::path::PathBuf;
 use std::process::Command;
 use std::thread::{self, sleep};
 use std::time::Duration;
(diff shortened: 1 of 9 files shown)


bin/tests/cli.rs
```diff
@@ -391,3 +396,36 @@ fn test_directory_symlinks_delete() {
...
+        &mut stderr_reader,
+    );
+
+    for i in 0..10 {
+        journal::print(1, format!("Sample alert {}", i).as_str());
+        journal::print(6, format!("Sample info {}", i).as_str());
+    }
+
+    common::wait_for_event("sending journald line", &mut stderr_reader);
+    common::wait_for_event("sending http request", &mut stderr_reader);
```
jorgebay, line 427: Not very elegant...

Once we have the mock ingester as a dev dependency we can verify what was received.

bin/tests/common.rs
```diff
@@ -172,7 +184,7 @@ pub fn create_dirs<P: AsRef<Path>>(dirs: &[P]) {
 
 pub fn open_files_include(id: u32, file: &PathBuf) -> Option<String> {
     let child = Command::new("lsof")
-        .args(&["-p", &id.to_string()])
+        .args(&["-l", "-p", &id.to_string()])
```
jorgebay, line 187: lsof was showing in the output "no pwd entry for UID" which was irritating, I've fixed it here even though it's unrelated to this PR.

docker/journald/Dockerfile
```diff
@@ -1,5 +1,4 @@
 FROM docker.io/logdna/build-images:rust-buster-stable
-RUN apt update -y && apt install -y systemd lsof procps
```
jorgebay, line 2: We should clean the image cache on our local dev env as a result of this.

bin/tests/common.rs
```diff
@@ -184,32 +185,43 @@ pub fn spawn_agent(settings: AgentSettings) -> Child {
...
 
+    if let Some(journald_dirs) = settings.journald_dirs {
+        agent.env("LOGDNA_JOURNALD_PATHS", journald_dirs);
+    }
+
     agent.spawn().expect("Failed to start agent")
 }
 
+/// Blocks until a certain event referencing a file name is logged by the agent
+pub fn wait_for_file_event(event: &str, file_path: &PathBuf, reader: &mut dyn BufRead) -> String {
```
c-nixon, line 196: 💯

c-nixon (approved):

bin/tests/common.rs
c-nixon: The &PathBuf could probably be a Path, but there's very little difference

jorgebay: Changed it!

PR #88, opened 2021-02-05
buffer fs events while matching moved_to/moved_from

c-nixon:
This changes the fs events stream so that it retains ordering of the underlying inotify events while pairing up the MovedFrom/MoveTo events.


jorgebay (approved):
lgtm!

It's looking stable under the different log rotation strategies.

PR #91, opened 2021-02-10
Support directories with dot in their names

jorgebay:
Default glob rule `!(*.*)` was preventing directories w/ dot in their names to be considered.
This removes the rule and adds validation to include sub directories by default, regardless of the name.

Semver: minor
Ref: LOG-8788

This changes the default configuration, so we shouldn't include it in a patch release (the change in settings should not affect user in any case).


c-nixon (approved):
LGTM

PR #93, opened 2021-02-15
Fix follow_links() loop and lookup()

jorgebay:
Fixes infinite loop in `follow_links()` method and fixes `lookup()` to only return an `Option<EntryKey>` (instead of wrapped in a result).

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -469,7 +469,7 @@ where
         }
 
         let parent_ref = self.create_dir(&path.parent().unwrap().into(), _entries)?;
-        let parent_ref = self.follow_links(parent_ref, _entries)?;
+        let parent_ref = self.follow_links(parent_ref, _entries);
 
         if parent_ref.is_none() {
             return Ok(None);
@@ -630,7 +630,7 @@ where
     ) -> FsResult<()> {
         let path_buf = path.parent().ok_or(Error::PathNotValid)?.into();
         let parent = self
-            .lookup(&path_buf, _entries)?
+            .lookup(&path_buf, _entries)
             .ok_or(Error::ParentLookup)?;
         let component = into_components(path).pop().ok_or(Error::PathNotValid)?;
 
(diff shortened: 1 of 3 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -849,65 +849,53 @@ where
...
-                        continue;
-                    }
+            parent = self.follow_links(parent, _entries).and_then(|e| {
+                if let Some(e) = _entries.get(e) {
+                    e.borrow()
+                        .children()?
+                        .get(&component)
+                        .and_then(|entry| self.follow_links(*entry, _entries))
+                } else {
+                    error!("Failed to find entry");
```
c-nixon, line 875: Is this always an error? Or should it be a warn/info?

common/fs/src/tail.rs
```diff
@@ -204,7 +204,7 @@ impl Tailer {
                             // if Some(..) = entries.get(key) clauses)
                             let mut entries = fs.entries.borrow_mut();
                             if entries.remove(entry_ptr).is_some() {
-                                debug!(
+                                info!(
```
c-nixon, line 207: Should this be put back to debug? If we'd like to keep it at info the 
message should be tweaked to not talk about the map, as that's an implementation detail that doesn't need logged

c-nixon (commented):
Awesome, great catch.

2 small queries around the logging, but the change looks solid

common/fs/src/cache/mod.rs
jorgebay: Good point, I'll change it to info.

common/fs/src/tail.rs
jorgebay: I see what you mean, having it an info level can be useful but the message doesn't make much sense for the user, how about: `"Removed file information, currently tracking {} files"`

c-nixon: That sounds good to me!

c-nixon (approved):

PR #95, opened 2021-02-16
Fix follow_links() loop and lookup()

jorgebay:
fix: Fix follow_links() loop and lookup()
chore: fix cargo-audit

Cherry pick fix to `master`.


c-nixon (approved):
🎉

PR #96, opened 2021-02-18
Use mock ingester to test tags

jorgebay:
Integration test using the mock ingester to validate that tags are being sent.

### DIFF
--- a/bin/tests/common.rs
+++ b/bin/tests/common.rs
@@ -14,7 +14,7 @@ use std::thread;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener};
 
 use futures::Future;
-use logdna_mock_ingester::{https_ingester, FileLineCounter, HyperError};
+use logdna_mock_ingester::{http_ingester, https_ingester, FileLineCounter, HyperError};
 
 use rcgen::generate_simple_self_signed;
 use rustls::internal::pemfile;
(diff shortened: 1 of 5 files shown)


jorgebay:
I'll update the journald tests #87 to use the mock ingester.

bin/tests/common.rs
```diff
@@ -119,8 +140,15 @@ impl<'a> AgentSettings<'a> {
 pub fn spawn_agent(settings: AgentSettings) -> Child {
     let mut cmd = Command::cargo_bin("logdna-agent").unwrap();
 
-    let ingestion_key =
-        std::env::var("LOGDNA_INGESTION_KEY").expect("LOGDNA_INGESTION_KEY env var not set");
+    let ingestion_key: String;
```
c-nixon, line 143: Extremely minor, but pretty much everything in rust is an expression, so you can do

```
let ingestion_key = if let Some(key) = settings.ingester_key {
        ingestion_key = key.to_string();
    } else {
        ingestion_key = std::env::var("LOGDNA_INGESTION_KEY")
            .expect("LOGDNA_INGESTION_KEY env var could not be obtained");
    };
```

c-nixon (approved):
This is great, and once we add a heartbeat to the event stream we'll be able to replace the force_flush with a no-op and these those all still pass.

PR #92, opened 2021-02-11
Replaced Slack link with new Discourse link. Updated rustc version.

tanberry:

PR #97, opened 2021-02-18
docs(community): switch to Discourse

nimbinatus:
Switch to send community members to Discourse instead of Slack.


jorgebay:
There's an existing PR up for this #92.

nimbinatus:
Ah! Then I'll close this as a duplicate :) Thanks!

PR #100, opened 2021-02-25
K8s improvements: backoff, small memoization and support disabling k8s log enrichment

jorgebay:
Adds backoff when handling with upstream k8s errors and supports disabling the use of k8s api altogether.

It also increases test coverage of k8s logic internals and config.

Ref: LOG-8885, LOG-8887 and LOG-8888.

This patch is out of scope of 3.1, so we have can merge it once 3.1 was released or start with 3.2 branch.

### DIFF
--- a/common/config/src/env.rs
+++ b/common/config/src/env.rs
@@ -82,9 +82,9 @@ pub struct Config {
     #[example("none")]
     pub lookback: Option<String>,
 
-    #[env(LOGDNA_USE_K8S_API)]
+    #[env(LOGDNA_USE_K8S_LOG_ENRICHMENT)]
     #[example("always")]
-    pub use_k8s_api: Option<String>,
+    pub use_k8s_enrichment: Option<String>,
 
     #[env(LOGDNA_LOG_K8S_EVENTS)]
     #[example("always")]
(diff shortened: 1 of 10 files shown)


common/config/src/env.rs
```diff
@@ -82,6 +82,10 @@ pub struct Config {
     #[example("none")]
     pub lookback: Option<String>,
 
+    #[env(LOGDNA_USE_K8S_API)]
```
c-nixon, line 85: Can we name this something like

LOGDNA_K8S_METADATA_LOG_ENRICHMENT

or something slightly more descriptive? I realise that it's very long for a name, USE_K8S_API is quite vague

common/k8s/src/middleware/metadata.rs
```diff
@@ -9,12 +9,17 @@ use kube::{api::ListParams, config::Config, Api, Client};
 use kube_runtime::watcher;
 use kube_runtime::watcher::Event as WatcherEvent;
 
+use backoff::backoff::Backoff;
```
c-nixon, line 12: Nice, replacing out custom "limit" implementation in common/http with this lib is on my TODO list

common/k8s/src/middleware/metadata.rs
```diff
@@ -9,12 +9,17 @@ use kube::{api::ListParams, config::Config, Api, Client};
...
+use backoff::ExponentialBackoff;
 use metrics::Metrics;
 use middleware::{Middleware, Status};
 use parking_lot::Mutex;
+use pin_utils::core_reexport::option::Option::Some;
```
c-nixon, line 17: Do we need to pull this in? It appears to just be core::option::Option, which is already imported as part of the rust prelude (via it's other re-export in std::option::Option)

jorgebay: Good catch. It was probably auto imported by the IDE... I'll remove it.

c-nixon (approved):

PR #101, opened 2021-02-25
feat: Split out TailedFile and flatten log line stream

c-nixon:
This is the first step toward stateful lookback.

It changes the tailer into a true multiplexed stream of lines from the various files the agent is tracking

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -36,7 +36,7 @@ type WatchDescriptors = HashMap<WatchDescriptor, Vec<EntryKey>>;
 
 pub type EntryKey = DefaultKey;
 
-type EntryMap<TailedFile> = SlotMap<EntryKey, RefCell<entry::Entry<TailedFile>>>;
+type EntryMap = SlotMap<EntryKey, RefCell<entry::Entry>>;
 type FsResult<T> = Result<T, Error>;
 
 #[derive(Debug, Error)]
@@ -63,7 +63,7 @@ pub enum Error {
 
 pub struct FileSystem {
     watcher: Watcher,
-    pub entries: Rc<RefCell<EntryMap<TailedFile>>>,
+    pub entries: Rc<RefCell<EntryMap>>,
     root: EntryKey,
 
     symlinks: Symlinks,
@@ -276,7 +276,7 @@ impl FileSystem {
         watch_descriptor: &WatchDescriptor,
         name: OsString,
         events: &mut Vec<Event>,
-        _entries: &mut EntryMap<TailedFile>,
+        _entries: &mut EntryMap,
     ) -> FsResult<()> {
         // directories can't be a hard link so we're guaranteed the watch descriptor maps to one
         // entry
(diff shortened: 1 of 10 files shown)


common/fs/src/tail.rs
```diff
@@ -129,20 +127,24 @@ impl Tailer {
...
-                                    file_handle,
                                     ..
                                 } = entry.borrow_mut().deref_mut()
                                 {
                                     info!("added {:?}", paths[0]);
-                                    *data = 0;
-                                    if let Some(mut lines) = Tailer::tail(file_handle, &paths, data) {
-                                        final_lines.append(&mut lines);
-                                    }
+                                    data.seek(0).unwrap_or_else(|e| error!("error seeking {:?}", e));
```
jorgebay, line 134: What's the reason to seek to 0?

`BufReader`'s `seek()` method will discard the internal buffer even if the offsets match.

jorgebay (commented):
It's looking good, some questions and suggestions below.

common/fs/src/tail.rs
c-nixon: Yeh, I think this is a no-op as it's only used for New files. This is just a leftover from the previous implementation, I'll remove it.

common/fs/src/cache/mod.rs
```diff
@@ -34,7 +36,7 @@ type WatchDescriptors = HashMap<WatchDescriptor, Vec<EntryKey>>;
 
 pub type EntryKey = DefaultKey;
 
-type EntryMap<T> = SlotMap<EntryKey, RefCell<entry::Entry<T>>>;
+type EntryMap<TailedFile> = SlotMap<EntryKey, RefCell<entry::Entry<TailedFile>>>;
```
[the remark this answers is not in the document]
c-nixon: Nice Catch, I hadn't gone far enough :)

common/fs/src/cache/tailed_file.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        // if the offset is greater than the file's len
+        // it's very likely a truncation occurred
+        if self.offset > len {
+            info!(
+                "{:?} was truncated from {} to {}",
+                &paths[0], self.offset, len
+            );
+            // Reset offset back to the start... ish?
+            // TODO: Work out the purpose of the 8192 something to do with lookback? That seems wrong.
```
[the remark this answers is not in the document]
c-nixon: It could be something to do with a truncate logrotation strategy trancating to 8k or something? I think it's worth digging into separately in case removing it makes us start duplicating 8k or logs every time a log file is rotated.

We should really work out a better system here thought...

jorgebay (approved):
Nice!

common/fs/src/cache/tailed_file.rs
```diff
@@ -0,0 +1,109 @@
...
+#[derive(Debug)]
+pub struct TailedFile {
+    reader: std::io::BufReader<File>,
+    offset: u64,
+}
+
+impl TailedFile {
+    pub(crate) fn new(path: &Path) -> Result<Self, std::io::Error> {
+        Ok(Self {
+            reader: BufReader::new(OpenOptions::new().read(true).open(path)?),
```
jorgebay, line 19: Fixed on 3.1 :)

PR #102, opened 2021-02-25
feat: fully streamify fs tailer

c-nixon:


common/fs/src/cache/tailed_file.rs
```diff
@@ -1,109 +1,233 @@
+#![allow(clippy::await_holding_refcell_ref)]
```
c-nixon, line 1: Clippy doesn't like us holding onto refcell borrows across await points. Generally this is a good lint, especially if you're expecting to access the things inside the RefCell concurrently. 

Unfortunately we're storing the filehandle inside the refcell, so we need to hold it. However, we aren't accessing the file handle concurrently, we're only using the RefCell for interior mutability, only the Lines stream should access it and only one of those should exist at once.

common/fs/src/cache/tailed_file.rs
```diff
@@ -1,109 +1,233 @@
...
+use std::fs::OpenOptions;
+use std::io;
+use std::mem;
 use std::path::{Path, PathBuf};
+use std::pin::Pin;
+use std::rc::Rc;
+use tokio::io::{BufReader, SeekFrom};
+use tokio_util::compat::{Compat, Tokio02AsyncReadCompatExt};
+
+fn read_until_internal<R: AsyncBufRead + ?Sized>(
```
c-nixon, line 22: Lifted straight from the futures::io crate

common/fs/src/cache/tailed_file.rs
```diff
@@ -1,109 +1,233 @@
...
 
-        if line_groups.is_empty() {
-            None
-        } else {
-            Some(futures::stream::iter(line_groups))
-        }
+        Some(
+            Lines::new(self.inner.clone())
+                .filter_map({
+                    let paths = paths.clone();
```
c-nixon, line 209: The path copying is unfortunate as most lines only appear in one tracked file...

jorgebay (commented):
I took a first pass at the patch, I've added some comments below.

I personally don't see the benefit of making the tailer reads async and adds some complexity to the code.

common/fs/src/cache/tailed_file.rs
c-nixon: Unfortunately not without a fairly significant amount of overhead, as I can't work out a way to feed in a pointer from outside and repeatedly yield it. We would need some unsafe code or maybe an RC<RefCell> for the output in order to avoid allocating.

I agree that it's unfortunate, but it should also never need to change and it's been fairly battle tested by the rest of the ecosystem

c-nixon:
> I personally don't see the benefit of making the tailer reads async and adds some complexity to the code.

This PR is mostly setup for tailer-refactor-3 which allows us to be completely lazy in reading lines from the tailed files. We need to be fully async in order to make that possible

PR #106, opened 2021-03-04
Tailer refactor 3

c-nixon:


bin/src/stream_adapter.rs
```diff
@@ -0,0 +1,175 @@
...
+    Lazy(LazyLineSerializer),
+}
+
+#[async_trait]
+impl IngestLineSerialize<String, bytes::Bytes, std::collections::HashMap<String, String>>
+    for StrictOrLazyLines<'_>
+{
+    type Ok = ();
+
+    fn has_annotations(&self) -> bool {
```
jorgebay, line 28: NIT: A macro can be used to avoid repetitive blocks and make sure the calls are the same.

It doesn't have to be part of this PR, maybe a small enhancement in the future.

I've manually checked that the calls are the same per each (except for `has_file()`) 👍

common/fs/src/cache/tailed_file.rs
```diff
@@ -1,21 +1,34 @@
...
+use std::collections::HashMap;
 use std::convert::TryInto;
 use std::fs::OpenOptions;
 use std::io;
 use std::mem;
+use std::ops::DerefMut;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
-use std::rc::Rc;
+//use std::rc::Arc;
```
jorgebay, line 30: NIT leftover.

common/fs/src/cache/tailed_file.rs
```diff
@@ -68,33 +81,33 @@ fn read_line_lossy<R: AsyncBufRead + ?Sized>(
...
-            read: 0,
-        }
+impl LineBuilderLines {
+    pub fn new(reader: Arc<Mutex<TailedFileInner>>) -> Self {
+        Self { reader, read: 0 }
     }
 }
 
-impl Stream for Lines {
+impl Stream for LineBuilderLines {
```
jorgebay, line 97: As I mentioned in https://github.com/logdna/logdna-agent-v2/pull/102 it would be nice if we could find a solution, like `stream::unfold()` w/ `read_until()` instead of having to code the stream internals ourselves.

These are sensitive lines of code that could hide hard to track issues.

jorgebay (commented):

common/k8s/src/middleware/metadata.rs
```diff
@@ -180,12 +180,19 @@ impl Middleware for K8sMetadata {
         });
     }
 
-    fn process(&self, mut line: LineBuilder) -> Status {
-        if let Some(ref file_name) = line.file {
+    fn process<'a>(&self, line: &'a mut dyn LineMetaMut) -> Status<&'a mut dyn LineMetaMut> {
```
jorgebay, line 183: We calling per each `LazyLineSerializer` but the annotations should be the same per each `TailedFile`.

bin/src/stream_adapter.rs
c-nixon: Yeh, I didn't want to prematurely optimise with a macro, if we add a third type (which I think we probably will as we add sources or port the existing ones) a macro makes a lot more sense

common/k8s/src/middleware/metadata.rs
c-nixon: Maybe it makes sense to inject this into the Tailers somehow? rather than applying it to the lines after the fact?

jorgebay: or when creating a `LazyLineSerializer` we could try to copy the previous values and check here if the annotations/labels are already filed in.

common/fs/src/cache/tailed_file.rs
c-nixon: Yeh, I'd like to get away from the raw trait implementations. I had a look at trying to use unfold for the lower level read_until function, but I couldn't work out a way to get data out without allocating further.

The imperative, manual state machine makes it a lot easier to deal with passing pointers across polls (you just stash them in the struct)

common/fs/src/cache/tailed_file.rs
```diff
@@ -107,39 +120,341 @@ impl Stream for Lines {
...
+    level: Option<String>,
+    meta: Option<Value>,
+
+    path: String,
+
+    reader: Arc<Mutex<TailedFileInner>>,
+}
+
+#[async_trait]
+impl IngestLineSerialize<String, bytes::Bytes, std::collections::HashMap<String, String>>
```
c-nixon, line 166: I think there is a refactoring possible to extract the has_* methods out as defaults using the LineMeta trait. It will require a change in the client too

common/k8s/src/middleware/metadata.rs
c-nixon: We need a way to update them over time, I suppose we could change TailedFile for the LazyLineSerializer to subscribe directly to the k8s metadata and maintain a template internally which it can update and clone? Then we'd only be changing the template when we need to and clones should be quick.

Makefile
```diff
@@ -106,7 +106,7 @@ test: test-journald ## Run unit tests
 integration-test: ## Run integration tests using image with additional tools
 	$(DOCKER_JOURNALD_DISPATCH) "--env LOGDNA_INGESTION_KEY=$(LOGDNA_INGESTION_KEY) --env LOGDNA_HOST=$(LOGDNA_HOST) --env RUST_BACKTRACE=full --env RUST_LOG=$(RUST_LOG)" "cargo test --manifest-path bin/Cargo.toml --features integration_tests -- --nocapture --test-threads=$(INTEGRATION_TEST_THREADS)"
 
-.PHONY:test-journald
+.PHONY:test-journal
```
jorgebay, line 109: Probably unintended :)

jorgebay (approved):
Looks good to merge once CI finishes

PR #107, opened 2021-03-05
Stateful lookback

c-nixon:
https://github.com/logdna/logdna-agent-v2/pull/new/stateful-lookback


common/state/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+authors = ["Chris Nixon <engineering@logdna.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+derivative = "2.2"
+futures = "0.3"
+rocksdb = "0.15"
+async-channel = "1.6"
```
jorgebay, line 13: Just curious, why you chose async-channel instead of tokio's `broadcast`?

common/state/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+        let rx = self
+            .rx
+            .borrow_mut()
+            .take()
+            .ok_or(FileOffsetStateError::AlreadyRunning)?;
+        let self_ref = self.db.clone();
+        Ok(rx
+            .fold(Some(WriteBatch::default()), {
+                let self_ref = self_ref.clone();
+                move |a, e| {
```
jorgebay, line 244: NIT: We could add more meaningful variable names, like `|acc, event|`.

common/state/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+                    let self_ref = self_ref.clone();
+                    async move {
+                        let cf_handle = self_ref.cf_handle(OFFSET_NAME).unwrap();
+                        match (a, e) {
```
jorgebay, line 248: Nice db grouping of operations and the pattern matching groups, simplifies the logic quite a bit!

common/state/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+    use super::*;
+    use tempfile::tempdir;
+
+    #[test]
+    fn it_works() {
+        let _ = env_logger::Builder::from_default_env().try_init();
+        let data_dir = tempdir().expect("Could not create temp dir").into_path();
+        let db_path = data_dir.join("agent_state.db");
+
+        fn _test(db_path: &std::path::Path, initial_count: usize) {
```
jorgebay, line 305: NIT: We could use some code comments for future-selves :)

jorgebay (commented):
It's looking really good, I've added some comments below.

common/state/Cargo.toml
c-nixon: By using async-channel it means that the module is runtime agnostic, at least unless we're running the tests.

It's a pretty minor nice-to-have given the rest of the agent all depends on Tokio, but in this case it was easy the module agnostic in case we want to use it somewhere else.

jorgebay: ok, gotcha

jorgebay:
I'm thinking maybe we should use a different setting for stateful lookback, enabled by default.
When enabled it will use the internal db to get the offset of each file at startup. When the file offset is not part of the db, we use small files (default) or start.

Even further, internal db is enabled by default, without a setting and continue using lookback enum:
- `Lookback::None` -> no store in db, no lookback, always start at the end of the file.
- `Lookback::SmallFiles` -> when there's no offset stored in the db (fresh startup), start from the beginning if the file has 8Kb or less.
- `Lookback::Start` -> when there's no offset stored in the db (fresh startup), start from the beginning of the file.

bin/src/main.rs
```diff
@@ -9,6 +9,7 @@ use futures::Stream;
 use crate::stream_adapter::{StrictOrLazyLineBuilder, StrictOrLazyLines};
 use config::Config;
 use env_logger::Env;
+use fs::tail::Lookback;
```
jorgebay, line 12: _Adding the comment here for convenience but the comment is not related to this line_

Should we keep `bin/default.profraw` out of source control?

bin/tests/cli.rs
```diff
@@ -705,3 +705,147 @@ async fn test_tags() {
...
+                    .lines;
+                shutdown_handle();
+
+                handle.wait().unwrap();
+                line_count
+            },
+            async move {
+                tokio::time::delay_for(tokio::time::Duration::from_millis(500)).await;
+                let mut file = OpenOptions::new()
+                    .append(true)
```
jorgebay, line 836: Should we use `.create(false)` to prevent false positives?

jorgebay (commented):
Looks good to me, minor comments below.

common/state/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+        let db = self.db.clone();
+        Ok(rx
+            .fold(Some(WriteBatch::default()), {
+                move |acc, event| {
+                    let db = db.clone();
+                    async move {
+                        let cf_handle = db.cf_handle(OFFSET_NAME).unwrap();
+                        match (acc, event) {
+                            (Some(wb), FileOffsetEvent::Flush) => {
+                                db.write(wb).expect("Couldn't flush state"); // TODO
```
jorgebay, line 253: Should we create a follow up ticket or should we address it now?

jorgebay (commented):
I think the change of logic (enabled by default) messed up with integration test logic (we should clear/remove the db before each).

Besides that, I think it's in good state 👍

jorgebay (approved):
Nice! :shipit: 

Let's file a ticket for the db write error.

PR #108, opened 2021-03-10
3.1 yamls

c-nixon:


docs/README.md
```diff
@@ -139,6 +139,7 @@ The agent accepts configuration from two sources, environment variables and a co
 |`LOGDNA_JOURNALD_PATHS`|Comma separated list of paths (directories or files) of journald paths to monitor||
 |`LOGDNA_LOOKBACK`|The lookback strategy on startup|`smallfiles`|
 |`LOGDNA_LOG_K8S_EVENTS`|Whether the agent should capture Kubernetes events|`always`|
+|`LOGDNA_DB_PATH`|The directory the agent will store it's state database||
```
jorgebay, line 142: We can add a small note here as well:
```suggestion
|`LOGDNA_DB_PATH`|The directory the agent will store it's state database. Note that the agent must have write access to the directory and be a persistent volume.||
```

k8s/agent-resources.yaml
```diff
@@ -116,6 +116,8 @@ spec:
                 secretKeyRef:
                   name: logdna-agent-key
                   key: logdna-agent-key
+            - name: LOGDNA_DB_PATH
+              value: /var/lib/logdna
```
jorgebay, line 120: We should remove it from the k8s yaml files.

k8s/agent-resources.yaml
```diff
@@ -142,6 +144,8 @@ spec:
               mountPath: /var/log
             - name: vardata
               mountPath: /var/data
+            - name: varliblogdna
+              mountPath: /var/lib/logdna
```
jorgebay, line 148: Do we still want to mount it after changing the default? should we document it and leave the yaml files as is?

jorgebay (commented):
Couple of questions below

k8s/agent-resources.yaml
c-nixon: I checked with Albert, he'd like it enabled by default on kubernetes

jorgebay (approved):
Apart the doc suggestion from above, lgtm.

PR #109, opened 2021-03-10
3.1 non root persistent lookback docs

c-nixon:


johnpadilla (commented):
Should we say "log messages" to make it clear? I'm pretty neutral. Just an idea.

docs/KUBERNETES.md
```diff
@@ -187,6 +187,27 @@ Alternatively, update the DaemonSet configuration directly by using the followin
...
 
+### Enabling persistent agent state
+
+To avoid skipping or duplicating logs when the agent pod restarts or is replaced the agent needs to store it's progress on the host node's filesystem. This is achieved using a hostPath volume. The host directory must be writable by the user or group specified in the securityContext.
+
+To achieve this the host directory must either already exist with the correct
+permissions or else Kubernetes will create the directory without write permissions for the agent user.
+
+In this case the permissions must be set before the agent starts. When running as non-root the agent pod does not have permissions to do this, so an initcontainer may be used.
+
```
tanberry, line 198: Suggested rewrite of lines 194-197: "To achieve this the host directory must already exist with the correct permissions before the agent starts (otherwise Kubernetes will create the directory without the needed write permissions for the agent user). When running as non-root the agent pod does not have permissions to configure the host directory, so we use an *init container*, which is run before the app containers are started, to set the host directory permissions."

docs/KUBERNETES.md
```diff
@@ -187,6 +187,27 @@ Alternatively, update the DaemonSet configuration directly by using the followin
...
+Below is an example manifest section which can be added to the pod specification alongside the containers array. It assumes the agent user/group are both 5000 and the volume is mounted at /var/lib/logdna
+
```
tanberry, line 200: can we say "...an example manifest section of the -file-name-.yaml which can be added..." just to again be uber-clear about **where** this section lives? If you think everyone knows, skip it, but... ;-)

Also can we add a few more words to explain what 5000 means.. is this a permission level? Like the top level or...? I tried googling it and found nothing.

docs/KUBERNETES.md
```diff
@@ -187,6 +187,27 @@ Alternatively, update the DaemonSet configuration directly by using the followin
 kubectl patch daemonset -n logdna-agent logdna-agent --type json -p '[{"op":"add","path":"/spec/template/spec/containers/0/securityContext/runAsUser","value":5000},{"op":"add","path":"/spec/template/spec/containers/0/securityContext/runAsGroup","value":5000}]'
 ```
 
+### Enabling persistent agent state
+
```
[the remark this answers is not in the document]
johnpadilla: My only issue - will users get "enabling persistent agent state" = "not omitting/duplicating logs?"? Otherwise I'd think fine. Thought this was a issue whether running as root or non-root.

docs/KUBERNETES.md
```diff
@@ -187,6 +187,27 @@ Alternatively, update the DaemonSet configuration directly by using the followin
...
+To avoid skipping or duplicating logs when the agent pod restarts or is replaced the agent needs to store it's progress on the host node's filesystem. This is achieved using a hostPath volume. The host directory must be writable by the user or group specified in the securityContext.
```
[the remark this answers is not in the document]
johnpadilla: again--"current state" ... sufficient?

docs/KUBERNETES.md
```diff
@@ -187,6 +187,27 @@ Alternatively, update the DaemonSet configuration directly by using the followin
...
+
```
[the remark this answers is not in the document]
johnpadilla: securityContext is just that...not an env variable. I'd say the yaml is a given.

docs/KUBERNETES.md
johnpadilla: Not sure why "init container" would be italicized above. We'd need to do the same in other places in the doc. Also, I don't believe *we* use an init container - we're telling the user to use one

docs/KUBERNETES.md
johnpadilla: I think the mention of "containers array"  is sufficient. 5000:5000 - just an example for setting ownership - user and group.

johnpadilla (changes requested):
Comments apply to openshift section as well. I'm not a fan of passive voice, but that's picky at this point.

docs/KUBERNETES.md
tanberry: the use of italics is standard for the first time a new term is introduced. A way to call attention to it, make sure the reader sees it. So we would only use it once. BUt.. we can also drop it, or use `init container`.

johnpadilla: Got that--but do we do the same for first use of other terms in the doc? That was my concern.

docs/KUBERNETES.md
```diff
@@ -187,6 +187,29 @@ Alternatively, update the DaemonSet configuration directly by using the followin
 kubectl patch daemonset -n logdna-agent logdna-agent --type json -p '[{"op":"add","path":"/spec/template/spec/containers/0/securityContext/runAsUser","value":5000},{"op":"add","path":"/spec/template/spec/containers/0/securityContext/runAsGroup","value":5000}]'
 ```
 
+### Enabling file offset tracking across restarts
+
+To avoid possible duplication or skipping of log messages during agent restart or upgrade, the agent stores its current file offsets on the host node's filesystem, using a hostPath volume.
```
jorgebay, line 192: Use host or node but not both.

```suggestion
To avoid possible duplication or skipping of log messages during agent restart or upgrade, the agent stores its current file offsets on the host's filesystem, using a `hostPath` volume.
```

docs/KUBERNETES.md
```diff
@@ -187,6 +187,29 @@ Alternatively, update the DaemonSet configuration directly by using the followin
...
+
+The host directory must be writable by the user or group specified in the securityContext.
```
jorgebay, line 194:
```suggestion
The host directory must be writable by the user or group specified in the `securityContext`.
```

docs/KUBERNETES.md
```diff
@@ -187,6 +187,29 @@ Alternatively, update the DaemonSet configuration directly by using the followin
...
+
+To achieve this the host directory must either already exist with the correct
+permissions or else Kubernetes will create the directory without write permissions for the agent user.
+
+In this case the permissions must be set before the agent starts. When running as non-root the agent pod does not have permissions to do this, so an initcontainer may be used.
+
+Below is an example manifest section which can be added to the pod specification alongside the containers array. It assumes the agent user/group are both 5000 and the volume is mounted at /var/lib/logdna
```
jorgebay, line 201:
```suggestion
Below is an example manifest section which can be added to the pod specification alongside the containers array. It assumes the agent user/group are both `5000` and the volume is mounted at `/var/lib/logdna`:
```

jorgebay (commented):
It's looking good, I've added minor suggestions below.

jorgebay (approved):

PR #110, opened 2021-03-11
Fix extra buffer clear

c-nixon:


common/fs/src/cache/tailed_file.rs
```diff
@@ -56,8 +59,9 @@ fn read_until_internal<R: AsyncBufRead + ?Sized>(
         if done || used == 0 {
             if used == 0 {
                 Metrics::fs().increment_partial_reads();
+                return Poll::Ready(None);
```
jorgebay, line 62: I think there's no need in adding a new set of possible states with `Option<T>`. Using `Poll::Ready(Ok(0))` should work.

jorgebay (commented):
I've added a comment below to reduce the number of states by avoiding `Option<T>`

common/fs/src/cache/tailed_file.rs
```diff
@@ -110,11 +110,14 @@ impl Stream for LineBuilderLines {
...
+        let (mut s, n) = match ready!(read_line_lossy(pinned_reader, cx, buf, this.read)) {
+            Ok(Some((s, n))) => (s, n),
+            Err(e) => return Poll::Ready(Some(Err(e))),
+            Ok(None) => return Poll::Ready(None),
+        };
         if n == 0 && s.is_empty() {
             return Poll::Ready(None);
-        }
-
+        };
```
jorgebay, line 120: Is this check needed? in which situation could it occur?

jorgebay (approved):

PR #116, opened 2021-03-18
Update to tokio 1.0 ecosystem and sanitise deps

c-nixon:
This updates the agent to tokio 1.0 and also updates our other deps to more recent versions.

Specifically I tried to minimise the number of copies of any given dependency. This helps with binary size, attack surface and compile times.

There should be no behavioural changes


jorgebay (approved):
Nice!

Thanks for taking care of this.

PR #117, opened 2021-03-22
Increment lines on lazy line codepath

c-nixon:
The new lazy lines tailer wasn't incrementing the lines metric.


common/fs/src/cache/tailed_file.rs
```diff
@@ -423,6 +423,7 @@ impl Stream for LazyLines {
                     current_offset.clone().unwrap(),
                 );
                 *path += 1;
+                Metrics::fs().increment_lines();
```
jorgebay, line 426: I think the same applies for `add_bytes()`

jorgebay (approved):

PR #119, opened 2021-03-22
 Exclude, include and redact log lines via regex

jorgebay:
Ref: LOG-9091 and LOG-9092

### DIFF
--- a/common/config/src/env.rs
+++ b/common/config/src/env.rs
@@ -81,7 +81,7 @@ pub struct Config {
     #[env(LOGDNA_LINE_INCLUSION_REGEX)]
     pub line_inclusion_regex: Option<EnvList<String>>,
 
-    #[env(LOGDNA_LINE_REDACT_REGEX)]
+    #[env(LOGDNA_REDACT_REGEX)]
     #[example(r"\S+@\S+\.\S+")]
     pub line_redact_regex: Option<EnvList<String>>,
 
--- a/common/config/src/env.rs
+++ b/common/config/src/env.rs
@@ -247,16 +247,16 @@ impl Config {
             raw.log.lookback = self.lookback;
         }
 
-        if let Some(mut list) = self.line_exclusion_regex {
-            raw.log.line_exclusion_regex.append(&mut list);
+        if let Some(list) = self.line_exclusion_regex {
+            raw.log.line_exclusion_regex = Some(list.deref().clone());
         }
 
-        if let Some(mut list) = self.line_inclusion_regex {
-            raw.log.line_inclusion_regex.append(&mut list);
+        if let Some(list) = self.line_inclusion_regex {
+            raw.log.line_inclusion_regex = Some(list.deref().clone());
         }
 
-        if let Some(mut list) = self.line_redact_regex {
-            raw.log.line_redact_regex.append(&mut list);
+        if let Some(list) = self.line_redact_regex {
+            raw.log.line_redact_regex = Some(list.deref().clone());
         }
 
         raw
--- a/common/config/src/env.rs
+++ b/common/config/src/env.rs
@@ -74,6 +74,17 @@ pub struct Config {
     #[example("/var/log/.*,/var/data/.*")]
     pub inclusion_regex_rules: Option<EnvList<String>>,
 
+    #[env(LOGDNA_LINE_EXCLUSION_REGEX)]
+    #[example("DEBUG")]
+    pub line_exclusion_regex: Option<EnvList<String>>,
+
+    #[env(LOGDNA_LINE_INCLUSION_REGEX)]
+    pub line_inclusion_regex: Option<EnvList<String>>,
+
+    #[env(LOGDNA_LINE_REDACT_REGEX)]
+    #[example(r"\S+@\S+\.\S+")]
+    pub line_redact_regex: Option<EnvList<String>>,
+
     #[env(LOGDNA_JOURNALD_PATHS)]
     #[example("/var/log/journal")]
     pub journald_paths: Option<EnvList<PathBuf>>,
@@ -234,6 +245,18 @@ impl Config {
 
         if self.lookback.is_some() {
             raw.log.lookback = self.lookback;
+        }
+
+        if let Some(mut list) = self.line_exclusion_regex {
+            raw.log.line_exclusion_regex.append(&mut list);
+        }
+
+        if let Some(mut list) = self.line_inclusion_regex {
+            raw.log.line_inclusion_regex.append(&mut list);
+        }
+
+        if let Some(mut list) = self.line_redact_regex {
+            raw.log.line_redact_regex.append(&mut list);
         }
 
         raw
--- a/common/fs/src/cache/tailed_file.rs
+++ b/common/fs/src/cache/tailed_file.rs
@@ -409,7 +409,7 @@ impl LineMetaMut for LazyLineSerializer {
             return (None, self.line_buffer.as_deref());
         }
 
-        return match self.reader.try_lock() {
+        match self.reader.try_lock() {
             Some(file_inner) => {
                 // Cache the value to avoid excessive cloning
                 self.line_buffer = Some(file_inner.buf.clone());
@@ -419,7 +419,7 @@ impl LineMetaMut for LazyLineSerializer {
                 // This should never happen but should be handled by callers
                 (None, None)
             }
-        };
+        }
     }
 }
 
--- a/common/fs/src/cache/tailed_file.rs
+++ b/common/fs/src/cache/tailed_file.rs
@@ -1,4 +1,4 @@
-use http::types::body::{KeyValueMap, LineBuilder, LineMeta, LineMetaMut, LineBufferMut};
+use http::types::body::{KeyValueMap, LineBufferMut, LineBuilder, LineMeta, LineMetaMut};
 use http::types::error::LineMetaError;
 use http::types::serialize::{
     IngestLineSerialize, IngestLineSerializeError, SerializeI64, SerializeMap, SerializeStr,
@@ -406,7 +406,7 @@ impl LineBufferMut for LazyLineSerializer {
                 self.line_buffer = Some(file_inner.buf.clone());
                 self.line_buffer.as_deref()
             }
-            None => None
+            None => None,
         }
     }
 
--- a/common/fs/src/cache/tailed_file.rs
+++ b/common/fs/src/cache/tailed_file.rs
@@ -285,7 +285,7 @@ impl IngestLineSerialize<String, bytes::Bytes, std::collections::HashMap<String,
             buf.clone()
         } else {
             let borrowed_reader = self.reader.lock().await;
-            bytes::Bytes::copy_from_slice(&borrowed_reader.buf[..borrowed_reader.buf.len() - 1])
+            line_bytes(&borrowed_reader.buf)
         };
         writer.serialize_utf8(bytes).await?;
 
@@ -403,7 +403,7 @@ impl LineBufferMut for LazyLineSerializer {
         match self.reader.try_lock() {
             Some(file_inner) => {
                 // Cache the value to avoid further cloning
-                self.line_buffer = Some(Bytes::from(file_inner.buf.clone()));
+                self.line_buffer = Some(line_bytes(&file_inner.buf));
                 self.line_buffer.as_deref()
             }
             None => None,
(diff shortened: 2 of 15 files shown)


jorgebay:
Depends on https://github.com/logdna/logdna-rust/pull/10

common/fs/src/cache/tailed_file.rs
```diff
@@ -390,6 +395,36 @@ impl LineMetaMut for LazyLineSerializer {
...
+        Ok(())
+    }
+
+    fn get_line(&mut self) -> (Option<&str>, Option<&[u8]>) {
+        if self.line_buffer.as_ref().is_some() {
+            // Get the value without locking
+            return (None, self.line_buffer.as_deref());
+        }
+
+        return match self.reader.try_lock() {
```
c-nixon, line 416: match is an expression, so you can remove the return and trailing semicolon

common/config/src/env.rs
```diff
@@ -74,6 +74,17 @@ pub struct Config {
...
     pub inclusion_regex_rules: Option<EnvList<String>>,
 
+    #[env(LOGDNA_LINE_EXCLUSION_REGEX)]
+    #[example("DEBUG")]
+    pub line_exclusion_regex: Option<EnvList<String>>,
+
+    #[env(LOGDNA_LINE_INCLUSION_REGEX)]
+    pub line_inclusion_regex: Option<EnvList<String>>,
+
+    #[env(LOGDNA_LINE_REDACT_REGEX)]
```
jorgebay, line 84: This should be `LOGDNA_REDACT_REGEX` instead.

common/middleware/src/line_rules.rs
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        // If any exclusion rule matches -> skip
+        if self.exclusion_text.iter().any(|r| r.is_match(value)) {
+            return Status::Skip;
+        }
+
+        if !self.redact_text.is_empty() {
+            let mut v = value.to_string();
+            for r in self.redact_text.iter() {
```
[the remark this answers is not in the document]
c-nixon: If we went with that method we could replace the Vec<Regex>'s with the regex crate's RegexSet https://docs.rs/regex/1.4.5/regex/struct.RegexSet.html which does a single pass match on all of the patterns contained in it, rather than having to match against every Regex separately

common/middleware/src/line_rules.rs
```diff
@@ -0,0 +1,260 @@
...
+use http::types::body::LineMetaMut;
+use regex::bytes::Regex as BytesRegex;
+use regex::Regex as TextRegex;
+use thiserror::Error;
+
+static REDACTED_TEXT: &str = "[REDACTED]";
+static REDACT_BYTES: &[u8] = &REDACTED_TEXT.as_bytes();
+
+pub struct LineRules {
+    exclusion_text: Vec<TextRegex>,
```
c-nixon, line 11: Do we need the TextRegexs? Could we not call as_bytes() on String lines and treat everything as bytes?

jorgebay: Good idea, I'll give it a try.

common/middleware/src/line_rules.rs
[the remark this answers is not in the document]
jorgebay: I think the logic would be more complex than that, like:

- Check each match of each regex
- Compare boundaries of each match, expanding when necessary
- Rebuild a new line with substrings of the original line and the redacted portions.

That compared to a loop of regex replace calls.

Looking at the docs for `RegexSet`, I think it only yields which regex matched (not the start and end indexes of the match). 

From the user pov, we can document the limitations of applying the redaction in the defined order and recommend the user to define the patterns in order, from more to less specific, wdyt?

jorgebay:
I've refactored it to use `regex::bytes` crate for both text and buffer lines.

jorgebay:
I think it's ready for another round of review @c-nixon
Also, lmkwyt about https://github.com/logdna/logdna-rust/pull/10

common/middleware/src/line_rules.rs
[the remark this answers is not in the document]
c-nixon: > Looking at the docs for RegexSet, I think it only yields which regex matched (not the start and end indexes of the match).

Ahh, darn, I've only ever used it to test for membership, pity it can't return locations.

> From the user pov, we can document the limitations of applying the redaction in the defined order and recommend the user to define the patterns in order, from more to less specific, wdyt?

I think we should do our best to handle the overlapping cases and try to reduce the chances of it surprising anyone as much as possible. There are some cases where it would be impossible to order the regexes so that bot applied correctly, eg in the case where the suffix of one pattern was the same as the prefix of another only one could possibly be applied correctly.

The algorithm for combining overlapping matches shouldn't be too bad, something like the following pseudocode:

```
let matches: Vec<(usize, usize)> = vec![];

for pattern in patterns.iter() {
    matches.extend(pattern.match(input);
}

let matches  = matches.sort_unstable();
let new_matches = vec![];

let mut current: Option<(usize, usize)> = None;
for i in (0..matches.len()) {
    let (start, mut end) = if let Some(start, end) = current {
        (start, end)
    } else {
        (matches[i].0, matches[i].1) 
    }

    if matches[i + 1].0 < end {
        // next match overlaps this one, extend the end of current
        end = matches[i + 1].1
    } else {
        // next match doesn't overlap, current match is complete, push it
        new_matches.push(current.take().unwrap())
    }
    
}

return new_matches
```

That's some pretty rought pseudocode and I'm pretty sure it could be condensed a bit, but I think it's the right algorithm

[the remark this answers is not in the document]
jorgebay: I'll update the logic to handle overlapping matches gracefully 👍

jorgebay:
I've updated the client and refactored redaction to support overlapping matches.

common/fs/src/cache/tailed_file.rs
```diff
@@ -164,8 +164,8 @@ pub struct LazyLineSerializer {
     labels: Option<KeyValueMap>,
     level: Option<String>,
     meta: Option<Value>,
-
     path: String,
+    line_buffer: Option<Vec<u8>>,
```
c-nixon, line 168: If we make this a bytes::Bytes we can save some more copies.

common/fs/src/cache/tailed_file.rs
```diff
@@ -279,7 +279,11 @@ impl IngestLineSerialize<String, bytes::Bytes, std::collections::HashMap<String,
     where
         S: SerializeUtf8<bytes::Bytes> + std::marker::Send,
     {
-        let bytes = {
+        // Try to use the cached value first
+        let replaced_buf = self.line_buffer.take();
```
c-nixon, line 283: If self.line_buffer is a Bytes object then we could replace this take with a clone() as Bytes has a very cheap Clone impl

common/fs/src/cache/tailed_file.rs
```diff
@@ -388,6 +393,29 @@ impl LineMetaMut for LazyLineSerializer {
...
+                // Cache the value to avoid excessive cloning
+                self.line_buffer = Some(file_inner.buf.clone());
+                self.line_buffer.as_deref()
+            }
+            None => None,
+        }
+    }
+
+    fn set_line_buffer(&mut self, line: Vec<u8>) -> Result<(), LineMetaError> {
+        self.line_buffer = Some(line);
```
c-nixon, line 414: if line_buffer is Bytes then we can do self.line_buffer = Some(line.into()) and the Vec allocation will be converted straight into a Bytes object without allocating

c-nixon (changes requested):
Minor tweaks around making the the line_buffer member more efficient by using Bytes

common/middleware/src/line_rules.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        Status::Ok(line)
+    }
+
+    fn redact<'a>(
+        &self,
+        value: Vec<u8>,
+        line: &'a mut dyn LineBufferMut,
+    ) -> Status<&'a mut dyn LineBufferMut> {
+        let mut matches: Vec<(usize, usize)> = vec![];
```
c-nixon, line 63: This might be a good place to use SmallVec? It depends how ofter we expect to find something to redact.

If there's no match 99% of the time then Vec makes sense as it doesn't allocate unless there's a match

If there are frequently 1-~5 or so matches then SmallVec would save a lot of allocations

If there are frequently more than 4 or 5 matches then it's porbably best for it to be a Vec again.


I think Vec is probably right for now, but I'm interested to hear your thoughts.

common/middleware/src/line_rules.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Applies inclusion and exclusion rules and replaces the redacted values.
+    fn process_line<'a>(
+        &self,
+        line: &'a mut dyn LineBufferMut,
+    ) -> Status<&'a mut dyn LineBufferMut> {
+        let value = line.get_line_buffer().unwrap();
+
+        // If it doesn't match any inclusion rule -> skip
+        if !self.inclusion.is_empty() && !self.inclusion.iter().any(|r| r.is_match(value)) {
```
c-nixon, line 42: Since we don't use the match locations anywhere the inclusion and exclusion rules could be turned into RegexSets, which would make the match marginally quicker since it can match in a single pass and the Regexset can optimise it's representation across all the Regexs at once

common/middleware/src/line_rules.rs
```diff
@@ -0,0 +1,293 @@
...
+        );
+        redact_match!(
+            p,
+            "my date of birth is 01-02-1985",
+            "my date of birth is [REDACTED]"
+        );
+    }
+
+    #[test]
+    fn should_support_overlapping_redactions() {
```
c-nixon, line 258: Very nice

docs/README.md
```diff
@@ -139,12 +139,22 @@ The agent accepts configuration from two sources, environment variables and a co
...
+|`LOGDNA_REDACT_REGEX`|Comma separated list of regex patterns used to mask matching sensitive information before sending it the log line.||
 |`LOGDNA_JOURNALD_PATHS`|Comma separated list of paths (directories or files) of journald paths to monitor||
 |`LOGDNA_LOOKBACK`|The lookback strategy on startup|`smallfiles`|
 |`LOGDNA_USE_K8S_LOG_ENRICHMENT`|Determines whether the agent should query the K8s API to enrich log lines from other pods.|`always`|
 |`LOGDNA_LOG_K8S_EVENTS`|Whether the agent should log Kubernetes resource events. This setting only affects tracking and logging Kubernetes resource changes via watches. When disabled, the agent may still query k8s metadata to enrich log lines from other pods depending on the value of `LOGDNA_USE_K8S_LOG_ENRICHMENT` setting value.|`never`|
 |`LOGDNA_DB_PATH`|The directory the agent will store it's state database. Note that the agent must have write access to the directory and be a persistent volume.||
 
+All regular expressions are [PCRE](https://www.pcre.org/) with case sensitivity by default. If you don't want to
```
c-nixon, line 151: Rust's Regex crate doesn't use PCRE under the hood, so it doesn't support all of PCRE's extensions, the supported regex patterns and extensions are at https://docs.rs/regex/1.4.5/regex/#syntax

c-nixon (commented):
Couple of comments around the matching and a suggested change for the docs

docs/README.md
jorgebay: Good catch, I'll update the docs.

common/middleware/src/line_rules.rs
jorgebay: From the use case, I think inclusion/exclusion rules match often and redact rules match rarely.

common/middleware/src/line_rules.rs
jorgebay: sure

common/fs/src/cache/tailed_file.rs
```diff
@@ -388,6 +393,29 @@ impl LineMetaMut for LazyLineSerializer {
...
+    fn get_line_buffer(&mut self) -> Option<&[u8]> {
+        if self.line_buffer.as_ref().is_some() {
+            // Get the value without locking
+            return self.line_buffer.as_deref();
+        }
+
+        match self.reader.try_lock() {
+            Some(file_inner) => {
+                // Cache the value to avoid further cloning
+                self.line_buffer = Some(Bytes::from(file_inner.buf.clone()));
```
c-nixon, line 406: This is doing approximately the same thing as https://github.com/logdna/logdna-agent-v2/pull/119/files#diff-a50000c4ee02c6f3895eebe2a870cac29962b2bb769f3cfad31793688e2de3d8R288, except this one is including the last character, which is always a trailing new line. Is that intentional?

jorgebay: Nice catch, I'll fix it.

jorgebay:
Fixed a bug related to redaction ordering and addressed the LF char issue.

c-nixon (approved):
This is great, All looks good to me!

PR #120, opened 2021-03-23
Test killing the agent with stateful lookback

jorgebay:
Use artificial delay to slow down the mock ingester in order to reproduce.


jorgebay:
At a closer look, this test shown server reordering: data was being appended by the mock ingester out of order when introducing delays, but the lines where being sent by the agent.

PR #121, opened 2021-03-25
Test symlink initialization

jorgebay:
Failing test that reproduces an issue found on k8s.


jorgebay:
We've included this in the fix PR #122

PR #124, opened 2021-03-29
Refactor tailer to split out event handling

c-nixon:
This splits out the event handling in the tailer into separate functions and flattens a lot of the if let x { Some(_) } else { None} logic

### DIFF
--- a/common/fs/src/tail.rs
+++ b/common/fs/src/tail.rs
@@ -149,7 +149,7 @@ impl Tailer {
     ) -> Option<impl Stream<Item = LazyLineSerializer>> {
         match event {
             Event::Initialize(entry_ptr) => {
-                debug!("Initialise Event");
+                debug!("Initialize Event");
                 // will initiate a file to it's current length
                 let entries = fs.entries.borrow();
                 let entry = entries.get(entry_ptr)?;
@@ -157,7 +157,7 @@ impl Tailer {
                 match entry {
                     Entry::File { name, data, .. } => {
                         // If the file's passes the rules tail it
-                        info!("Initialise event for entry {:?}, target {:?}", name, path);
+                        info!("initialize event for file {:?}, target {:?}", name, path);
                         let (_, offset) = Tailer::get_initial_offset(
                             &path,
                             &fs,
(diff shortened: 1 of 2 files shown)


common/fs/src/tail.rs
```diff
@@ -103,242 +297,263 @@ impl Tailer {
...
-                                            },
-                                            Lookback::SmallFiles => {
-                                                let offset = match initial_offsets.as_ref() {
-                                                    Some(initial_offsets) => {
-                                                        let offset = initial_offsets.get(&inode).copied().unwrap_or(0);
-                                                        debug!("Got offset {} from state for {:?} using key {:?} for path {:?}", offset, name, inode, path);
-                                                        offset
-                                                    }
-                                                    None => {
+        /*<<<<<<< HEAD
```
jorgebay, line 300: Leftover.

common/fs/src/tail.rs
```diff
@@ -87,6 +88,199 @@ impl Tailer {
...
+
+    async fn handle_event(
+        event: Event,
+        initial_offsets: Option<HashMap<FileId, u64>>,
+        lookback_config: Lookback,
+        fs: &FileSystem,
+    ) -> Option<impl Stream<Item = LazyLineSerializer>> {
+        match event {
+            Event::Initialize(entry_ptr) => {
+                debug!("Initialise Event");
```
jorgebay, line 149: We should continue using the same messages that we defined for 3.1 "`initialize`".

common/fs/src/tail.rs
```diff
@@ -87,6 +88,199 @@ impl Tailer {
...
+                // will initiate a file to it's current length
+                let entries = fs.entries.borrow();
+                let entry = entries.get(entry_ptr)?;
+                let path = fs.resolve_direct_path(&entry, &fs.entries.borrow());
+                match entry {
+                    Entry::File { name, data, .. } => {
+                        // If the file's passes the rules tail it
+                        info!("Initialise event for entry {:?}, target {:?}", name, path);
```
jorgebay, line 157: Same here, in 3.1 we also included "file" or "symlink" instead of entry.

common/fs/src/tail.rs
```diff
@@ -87,6 +88,199 @@ impl Tailer {
...
+        if let Entry::File { data, .. } = entry {
+            let inode: FileId = { (&data.borrow().deref().get_inode().await).into() };
+            Some((
+                entry_key,
+                match lookback_config {
+                    Lookback::Start => match initial_offsets.as_ref() {
+                        Some(initial_offsets) => {
+                            let offset = initial_offsets.get(&inode).copied().unwrap_or(0);
+                            debug!("Got offset {} from state using key {:?}", offset, path);
+                            offset
```
jorgebay, line 110: We can extract the duplicate code into a `fn` or move it above, as it's not dependent on the lookback value.

jorgebay (commented):
The new refactored code for event handling is easier to understand that the previous version, with far less code repetition 👍 


The leftover commented code made reviewing a little harder (diff doesn't help) but it's looking good. Thanks for porting it to master, that way we can run the full test suite in the face of 3.2 release.

common/fs/src/tail.rs
c-nixon: whoops!

common/fs/src/tail.rs
```diff
@@ -87,6 +88,199 @@ impl Tailer {
...
+                        if fs.is_initial_dir_target(&path) {
+                            return data.borrow_mut().tail(vec![path]).await;
+                        }
+                    }
+                    Entry::Symlink { name, link, .. } => {
+                        let sym_path = path;
+                        info!("Initialise event for entry {:?}, target {:?}", name, link);
+                        let entries = &fs.entries.borrow();
+                        let path = fs.resolve_direct_path(
+                            entries.get(Tailer::get_file_for_path(&fs, link)?)?,
```
jorgebay: Far less nesting also :)

jorgebay (approved):
Looks great!

PR #125, opened 2021-03-30
Use inode instead of filepath as lookback key

c-nixon:
This resolves an issue where symlinks to different hardlinks to the same underlying file were not being treated as the same file by the stateful lookback mechanism

### DIFF
--- a/bin/tests/cli.rs
+++ b/bin/tests/cli.rs
@@ -6,6 +6,7 @@ use proptest::prelude::*;
 use std::fs;
 use std::fs::{File, OpenOptions};
 use std::io::{BufRead, BufReader, Read, Write};
+use std::os::unix::fs::MetadataExt;
 use std::process::Command;
 use std::thread::{self, sleep};
 use std::time::Duration;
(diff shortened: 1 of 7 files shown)


jorgebay (commented):
Using the inode number is a good solution.

Given that we have to support the format that we choose for newer versions (forever!), I was thinking about windows support and how inode doesn't work like on that platform.

I was about to suggest to use a utf8 stringified u64 as rocksdb key instead, but thinking about it in more detail, we can use a different column family schema per platform (unix / windows).

common/fs/src/tail.rs
```diff
@@ -139,13 +139,13 @@ impl Tailer {
...
-                                                        let offset = initial_offsets.get(&path.as_os_str().as_bytes().into()).copied().unwrap_or(0);
-                                                        debug!("Got offset {} from state for {:?} using key {:?}", offset, name, path);
+                                                        let offset = initial_offsets.get(&inode).copied().unwrap_or(0);
+                                                        debug!("Got offset {} from state for {:?} using key {:?} for path {:?}", offset, name, inode, path);
                                                         offset
                                                     }
                                                     None => {
                                                         let len = path.metadata().map(|m| m.len()).unwrap_or(0);
-                                                        debug!("Smallfiles lookback {} from state for {:?} using key {:?}", len, name, path);
+                                                        debug!("Smallfiles lookback {} from len for {:?} using key {:?}", len, name, path);
```
jorgebay, line 148: oh, this message got me a few times while testing, thanks for taking care of it.

bin/tests/cli.rs
```diff
@@ -1022,6 +1023,152 @@ async fn test_symlink_to_symlink_initialization_excluded_file() {
...
+
+    let log_dir = tempdir().expect("Couldn't create temp dir...").into_path();
+    let excluded_dir = tempdir().expect("Couldn't create temp dir...").into_path();
+    let file_path = excluded_dir.join("test.log");
+    let excluded_hardlink_path = excluded_dir.join("test-hardlink.log");
+    let excluded_symlink_path = excluded_dir.join("test-symlink.log");
+    let symlink_path = log_dir.join("test-symlink.log");
+
+    File::create(&file_path).expect("Couldn't create temp log file...");
+    std::fs::hard_link(&file_path, &excluded_hardlink_path).unwrap();
```
jorgebay, line 1045: Nice coverage!

jorgebay (approved):

PR #127, opened 2021-04-06
3.1: Ignore new clippy errors

jorgebay:
With the build image update, clippy is complaining about rule changes.

For 3.1 branch, ignore the new rules to prevent a binary change.


c-nixon (approved):
Makes sense on 3.1

PR #128, opened 2021-04-06
Handle new clippy warnings

jorgebay:
Fix `from-over-into`, `ptr-arg` and `upper-case-acronyms` in `fs` and `http` crates

Relates to #127.


c-nixon (approved):
Simple refactor, lgtm

PR #129, opened 2021-04-06
Use stable rbac.authorization.k8s.io/v1 API

jorgebay:
RBAC mode is stable since k8s v1.8.

We should also update the helm charts.


jorgebay:
oh, new helm charts already use `rbac.authorization.k8s.io/v1` 💪

c-nixon (approved):
👍

PR #132, opened 2021-04-16
Minor doc typo fix

jorgebay:


c-nixon (approved):
Totally missed that when I was reading it last time!

PR #133, opened 2021-04-16
Support command line options

jorgebay:
Support cmd line args while continue to support env vars, deprecated env vars and config file.

This removes the config macro and replaces it by structops built-in support for environment variables.

The only command line arguments with short for are `-k` for ingestion key, `-t` for tags, `-d` for log dir and `-c` for config file.


jorgebay:
This is the output of `logdna-agent --help`:

```
LogDNA Agent 3.1.0-beta.2
A resource-efficient log collection agent that forwards logs to LogDNA.

USAGE:
    logdna-agent [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -c, --config <config>
            The config filename [env: LOGDNA_CONFIG_FILE=]  [default: /etc/logdna/config.yaml]

        --db-path <db-path>
            The directory the agent will store it's state database. Note that the agent must have write access to the
            directory and be a persistent volume. Defaults to "/var/lib/logdna-agent/" [env: LOGDNA_DB_PATH=]
        --endpoint-path <endpoint-path>
            The endpoint to forward logs to. Defaults to "/logs/agent" [env: LOGDNA_ENDPOINT=]

        --exclude-regex <exclusion-regex>...
            List of regex patterns to exclude files from monitoring [env: LOGDNA_EXCLUSION_REGEX_RULES=]

        --exclude <exclusion-rules>...
            List of glob patterns to exclude files from monitoring, to add to the default [env: LOGDNA_EXCLUSION_RULES=]

        --gzip-level <gzip-level>
            If compression is enabled, this is the gzip compression level to use. Defaults to 2 [env:
            LOGDNA_GZIP_LEVEL=]
        --host <host>
            The host to forward logs to. Defaults to "logs.logdna.com" [env: LOGDNA_HOST=]

        --include-regex <inclusion-regex>...
            List of regex patterns to include files from monitoring [env: LOGDNA_INCLUSION_REGEX_RULES=]

        --include <inclusion-rules>...
            List of glob patterns to includes files for monitoring, to add to the default (*.log) [env:
            LOGDNA_INCLUSION_RULES=]
        --ip <ip>
            The IP metadata to attach to lines forwarded from this agent [env: LOGDNA_IP=]

        --journald-paths <journald-paths>...
            List of paths (directories or files) of journald paths to monitor, for example: /var/log/journal or
            /run/systemd/journal [env: LOGDNA_JOURNALD_PATHS=]
    -k, --key <key>
            The ingestion key associated with your LogDNA account [env: LOGDNA_INGESTION_KEY=]

        --line-exclusion <line-exclusion>...
            List of regex patterns to exclude log lines. When set, the Agent will NOT send log lines that match any of
            these patterns [env: LOGDNA_LINE_EXCLUSION_REGEX=]
        --line-inclusion <line-inclusion>...
            List of regex patterns to include log lines. When set, the Agent will ONLY send log lines that match any of
            these patterns [env: LOGDNA_LINE_INCLUSION_REGEX=]
        --line-redact <line-redact>...
            List of regex patterns used to mask matching sensitive information before sending it the log line [env:
            LOGDNA_REDACT_REGEX=]
    -d, --logdir <log-dirs>...
            Adds log directories to scan, in addition to the default (/var/log) [env: LOGDNA_LOG_DIRS=]

        --log-k8s-events <log-k8s-events>
            Whether the agent should log Kubernetes resource events. This setting only affects tracking and logging
            Kubernetes resource changes via watches. When disabled, the agent may still query k8s metadata to enrich log
            lines from other pods depending on the value of `use_k8s_enrichment` setting value ("always" or "never").
            Defaults to "never" [env: LOGDNA_LOG_K8S_EVENTS=]
        --lookback <lookback>
            The lookback strategy on startup ("smallfiles", "start" or "none"). Defaults to "smallfiles" [env:
            LOGDNA_LOOKBACK=]
        --mac-address <mac>
            The MAC metadata to attach to lines forwarded from this agent [env: LOGDNA_MAC=]

        --os-hostname <os-hostname>
            The hostname metadata to attach to lines forwarded from this agent (defaults to os.hostname()) [env:
            LOGDNA_HOSTNAME=]
    -t, --tags <tags>...
            List of tags metadata to attach to lines forwarded from this agent [env: LOGDNA_TAGS=]

        --use-compression <use-compression>
            Whether to compress logs before sending. Defaults to "true" [env: LOGDNA_USE_COMPRESSION=]

        --use-k8s-enrichment <use-k8s-enrichment>
            Determines whether the agent should query the K8s API to enrich log lines from other pods ("always" or
            "never").  Defaults to "always" [env: LOGDNA_USE_K8S_LOG_ENRICHMENT=]
        --use-ssl <use-ssl>
            Whether to use TLS for sending logs. Defaults to "true" [env: LOGDNA_USE_SSL=]
```

common/config/src/argv.rs
```diff
@@ -0,0 +1,719 @@
...
+            ($key: ident, $var_name: ident, $ftype: ty) => {
+                if let Ok(v) = env_var(env::$var_name) {
+                    if let Ok(parsed) = std::str::FromStr::from_str(&v) {
+                        options.$key = parsed;
+                        info!("Using deprecated env var '$var_name'");
+                    } else {
+                        warn!("Deprecated env var '$var_name' could not be parsed to $ftype");
+                    }
+                }
+            };
```
jorgebay, line 335: My macro fu is weak, there's probably a way to avoid repetition for these two patterns.

docs/README.md
```diff
@@ -119,10 +119,12 @@ The compiled binary will be built to `./target/release/logdna-agent`.
 
 ### Options
 
-The agent accepts configuration from two sources, environment variables and a configuration YAML file. The default configuration yaml file is located at `/etc/logdna/config.yaml`. The following options are available:
+The agent accepts configuration from three different sources: environment variables, command line arguments and/or a
+configuration YAML file. The default configuration yaml file is located at `/etc/logdna/config.yaml`. The following
```
jorgebay, line 123: The config yaml file usage still has rough edges: I've filed LOG-9482, but there are more usability issues.

docs/README.md
```diff
@@ -119,10 +119,12 @@ The compiled binary will be built to `./target/release/logdna-agent`.
 
 ### Options
 
-The agent accepts configuration from two sources, environment variables and a configuration YAML file. The default configuration yaml file is located at `/etc/logdna/config.yaml`. The following options are available:
+The agent accepts configuration from three different sources: environment variables, command line arguments and/or a
```
jorgebay, line 122: I haven't included enough information about the command line arguments in the docs, I think the whole "--help" output belongs in a new doc "Using the agent on linux" or something like that. We have to reorganize the docs when we add support for Linux.

bin/tests/cmd_line_env.rs
```diff
@@ -0,0 +1,361 @@
+use assert_cmd::cargo::CommandCargoExt;
+pub use common::*; // workaround for unused functions
```
c-nixon, line 2: Is this to keep clippy happy?

common/config-macro/src/lib.rs
```diff
@@ -1,210 +0,0 @@
-#![recursion_limit = "128"]
-
```
c-nixon, line 2: 🎉

common/config/src/argv.rs
c-nixon: There are other ways to write it, but I don't think there's a way to make it much shorter, you have to decide between Option and non-option no matter what

common/config/src/argv.rs
```diff
@@ -0,0 +1,719 @@
...
+        );
+        assert_eq!(
+            config.journald.paths,
+            Some(vec_paths!["/default_journald", "/journal", "/journald"])
+        );
+    }
+
+    #[test]
+    fn parse_deprecated_test() {
+        // Avoid reusing the constants from above on purpose
```
c-nixon, line 692: 👍

docs/README.md
c-nixon: Yeh, we'll need to add in the installation etc, we could move the configuration into a separate md and reference it from both KUBERNETES.md and Linux, or maybe we should have CLI in one place and ENV var based config in another...

common/config/src/argv.rs
```diff
@@ -0,0 +1,719 @@
...
+    pub const EXCLUSION_RULES_DEPRECATED: &str = "LOGDNA_EXCLUDE";
+    pub const EXCLUSION_REGEX_RULES_DEPRECATED: &str = "LOGDNA_EXCLUDE_REGEX";
+    pub const INCLUSION_RULES_DEPRECATED: &str = "LOGDNA_INCLUDE";
+    pub const INCLUSION_REGEX_RULES_DEPRECATED: &str = "LOGDNA_INCLUDE_REGEX";
+}
+
+/// Contains the command and env var options.
+#[derive(StructOpt, Debug, Default)]
+// Using PKG_VERSION as a workaround while we centralize version management of packages
+#[structopt(name = "LogDNA Agent", about = "A resource-efficient log collection agent that forwards logs to LogDNA.", version = unsafe { PKG_VERSION })]
```
c-nixon, line 57: Hmm, I wonder how this macro expands... if it expands the whole way during compile time the PKG_VERSION won't be set, or will be set to something random, I think this warrants an integration test, even just to check that logdna-agent --version successfuly returns something starting with 3

c-nixon (commented):
Minor comment around testing the PKG_VERSION hack still works as expected given it's being processed at compile time now rather than at runtime.

common/config/src/argv.rs
jorgebay: Makes sense, I'll add it.

jorgebay: I've included a test for it.

jorgebay:
Rebased it and pushed a test for `--version`.

c-nixon (approved):
LGTM!

PR #135, opened 2021-04-20
Integration test for retries

jorgebay:
It adds basic integration test coverage for retries by adding artificial delays to mock ingester responses.

I've added minor modifications to the client wrapper and to retry mod to support setting the retry parameters in the test.

### DIFF
--- a/bin/tests/common.rs
+++ b/bin/tests/common.rs
@@ -117,6 +117,7 @@ pub struct AgentSettings<'a> {
     pub use_ssl: bool,
     pub ingester_key: Option<&'a str>,
     pub tags: Option<&'a str>,
+    pub config_file: Option<&'a str>,
     pub state_db_dir: Option<&'a std::path::Path>,
     pub line_exclusion_regex: Option<&'a str>,
     pub line_inclusion_regex: Option<&'a str>,
@@ -190,6 +191,10 @@ pub fn spawn_agent(settings: AgentSettings) -> Child {
         agent.env("LOGDNA_EXCLUSION_REGEX_RULES", rules);
     }
 
+    if let Some(config_file) = settings.config_file {
+        agent.env("LOGDNA_CONFIG_FILE", config_file);
+    }
+
     if let Some(tags) = settings.tags {
         agent.env("LOGDNA_TAGS", tags);
     }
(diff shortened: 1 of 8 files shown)


bin/tests/common.rs
```diff
@@ -234,6 +239,9 @@ where
     let mut lines_buffer = String::new();
     for _safeguard in 0..100_000 {
         reader.read_line(&mut line).unwrap();
+        if line.is_empty() {
```
c-nixon, line 242: Is this where the empty lines in output was coming from?

common/http/src/client.rs
```diff
@@ -43,6 +44,8 @@ impl Client {
     pub fn new(
         template: RequestTemplate,
         state_handles: Option<(FileOffsetWriteHandle, FileOffsetFlushHandle)>,
+        retry_base_delay: Duration,
```
c-nixon, line 47: This is much better, even if we end up hard coding it in the config and never override it

common/test/mock-ingester/src/lib.rs
```diff
@@ -25,6 +25,7 @@ use tokio_stream::wrappers::TcpListenerStream;
 const ROOT: &str = "/logs/agent";
 
 pub type FileLineCounter = Arc<Mutex<HashMap<String, FileInfo>>>;
+pub type ProcessFn = Arc<Mutex<Box<dyn Fn(&IngestBody) + Send>>>;
```
c-nixon, line 28: This _really_ doesn't matter, but just for future reference because a Fn can be called as many times as you like it can't carry a reference to any state, so it _must_ be Sync implicitly (as there's no way 2 Fns can affect each other).

And since Arc is pretty much fat Box that carries an atomic usize around with it it means that this could just be Arc<dyn Fn(&IngestBody)>, no Mutex or 2nd indirection required.

c-nixon (approved):
Nice! minor nits, nothing that needs changed

common/test/mock-ingester/src/lib.rs
jorgebay: Good point, I don't remember why I added the `Mutex`... I'll remove it.

bin/tests/common.rs
jorgebay: yeah, when the process has exited, reading lines is a noop so we were writing empty lines to debug output.

jorgebay:
I've probably used `Mutex` to get around the `Sync` restriction but it doesn't make much sense. Marking the `Fn` as `Sync` did the trick, thanks @c-nixon for the tip!

c-nixon (approved):
Looks Great

PR #136, opened 2021-04-21
Log 9461 - Expose jemalloc profiling feature

c-nixon:
This PR adds a FEATURES env var and threads it through to to the various cargo build commands.

It's used like:

FEATURES=profiling make build-image

It also adds the profiling feature to the binary to expose the jemallocator/profiling feature

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,8 @@ FROM ${BUILD_IMAGE} as build
 ENV _RJEM_MALLOC_CONF="narenas:1,tcache:false,dirty_decay_ms:0,muzzy_decay_ms:0"
 ENV JEMALLOC_SYS_WITH_MALLOC_CONF="narenas:1,tcache:false,dirty_decay_ms:0,muzzy_decay_ms:0"
 
+ARG FEATURES
+
 ARG SCCACHE_BUCKET
 ENV SCCACHE_BUCKET=${SCCACHE_BUCKET}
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -21,8 +23,9 @@ COPY . .
 # Rebuild the agent
 RUN --mount=type=secret,id=aws,target=/root/.aws/credentials \
     if [ -z "$SCCACHE_BUCKET" ]; then unset RUSTC_WRAPPER; fi; \
-    cargo build --release && strip ./target/release/logdna-agent && \
+    cargo build --manifest-path bin/Cargo.toml ${FEATURES} --release && strip ./target/release/logdna-agent && \
     sccache --show-stats
+#${FEATURES}
```
jorgebay, line 28: NIT leftover :)

Makefile
```diff
@@ -76,6 +76,11 @@ LOGDNA_HOST?=localhost:1337
 
 RUST_LOG?=info
 
+space := $(subst ,, )
+comma := ,
+FEATURES?=
+FEATURES_ARG=$(if $(FEATURES),--features $(subst $(space),$(comma),$(FEATURES)))
```
jorgebay, line 82: Nice way to get around quote escaping!

jorgebay (approved):

Makefile
c-nixon: The first version I had used a function that was called everywhere, Took me a while to remember = was lazy in Make

PR #137, opened 2021-04-22
Listen to signals and exit gracefully

jorgebay:
Use `tokio::main` to start a multi thread runtime. Concurrently run the streams and await for the signals.

It currently does not flush the internal state db before exiting, I'll file a ticket for that.


common/k8s/src/middleware/metadata.rs
```diff
@@ -35,79 +35,57 @@ enum Error {
 pub struct K8sMetadata {
     metadata: Mutex<HashMap<(String, String), PodMetadata>>,
     api: Api<Pod>,
-    runtime: Mutex<Option<Runtime>>,
 }
 
 // TODO refactor to use kube-rs Reflector instead of manually managing hashmap
 impl K8sMetadata {
-    pub fn new() -> Result<Self, K8sError> {
-        let runtime = match Builder::new_multi_thread()
```
jorgebay, line 44: This issue was luckily caught by the k8s integration test (we shouldn't create a new runtime from a runtime context :) )

c-nixon: Nice work around, I'd love to refactor the Middlewares so they can be spawned on the main.rs executor rather than running in their own threads, but that's definitely out of scope for this PR

c-nixon (approved):
This look good to me, we don't need to do too much to try and flush buffers etc given the way stateful lookback works, and dropping the runtime should drop all running tasks anyway.

jorgebay:
Thanks! I'll leave this open until we cut a 3.2 branch.

c-nixon (approved):
LGTM

PR #139, opened 2021-04-27
Split up kube events stream logic

c-nixon:
This PR splits up the original kube events logic into separate associated functions


common/k8s/src/event_source.rs
```diff
@@ -261,198 +261,239 @@ impl K8sEventStream {
...
+                if let Some(oldest_pod_name) =
+                    K8sEventStream::get_oldest_pod(pods.clone(), &pod_label).await?
+                {
+                    if oldest_pod_name == pod_name {
+                        info!("begin logging k8s events");
+                        Ok(None)
+                    } else {
+                        info!("watching {}", oldest_pod_name);
+                        let params = ListParams::default()
+                            .timeout(30)
```
jorgebay, line 340: After the timer elapses, shouldn't we call watch again / restart the stream?

common/k8s/src/event_source.rs
```diff
@@ -261,198 +261,239 @@ impl K8sEventStream {
...
+        let events: Api<Event> = Api::all(client.as_ref().clone());
+        let params = ListParams::default();
+
+        let latest_event_time_w = latest_event_time.clone();
+        try_flatten_touched(watcher(events, params))
+            .map_err(K8sEventStreamError::WatcherError)
+            .filter({
+                move |ref event| {
+                    let latest_event_time = latest_event_time.clone();
+                    let earliest = previous_event_logger_delete_time.clone();
```
jorgebay, line 406: If the pod is starting and its the oldest one, Will this be `None`? If that's the case, is that the behaviour we want?

jorgebay (commented):
I went through the changes and overall the refactor looks good, I've left some comments below.

We should add some tests on the `k8s_test` suite, otherwise this part of the codebase will continue to be not covered. The tests could exercise:
- What happens when no oldest pod is found.
- When the oldest is the current pod
- When there's an oldest one and it goes down, after 30s.

common/k8s/src/event_source.rs
```diff
@@ -261,198 +261,239 @@ impl K8sEventStream {
...
+                    (
+                    p.metadata
+                        .labels
+                        .as_ref()
+                        .map(|l| {
+                            l.get("pod-template-generation")
+                                .and_then(|g| g.parse::<u64>().ok())
+                        })
+                        .flatten(),
+                    get_pod_started_at(p),
```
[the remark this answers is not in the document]
c-nixon: The reason I have that in there is to try and be robust against applications service meshes and other utilities that inject containers into pods

common/k8s/src/event_source.rs
c-nixon: Yes, if there is no earliest event then we go back to the start of the event stream and log from there.

The is the equivalent of Lookback::Start for files except that rather than the lookback happening on pod restart it should only happen if the entire agent install is removed and readded

common/k8s/src/event_source.rs
```diff
@@ -261,198 +261,239 @@ impl K8sEventStream {
...
+                                match e {
+                                    Cont::Cont => None,
+                                    Cont::Break => Some(((), ())),
+                                }
+                            });
+                        pin_mut!(stream);
+                        Ok(stream.next().await)
+                    }
+                } else {
+                    Ok(Some(((), ())))
```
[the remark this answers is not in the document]
c-nixon: I have wrapped the waiter in an exponential backoff

common/k8s/src/event_source.rs
jorgebay: gotcha 👍

common/k8s/src/event_source.rs
c-nixon: This is configuration for the kube-rs watcher object, it uses this to configured how it handles timeouts on the kubernetes REST api calls

common/k8s/src/event_source.rs
```diff
@@ -261,198 +261,239 @@ impl K8sEventStream {
...
+                    None
+                }
+            })
+            // Sorts into ascending order
+            .sorted_by(|(gen, started_at, name), (o_gen, o_started_at, o_name)| {
+                // For our purposes later generations should sort earlier than later ones
+                Ordering::reverse(Ord::cmp(gen, o_gen))
+                    .then(Ord::cmp(started_at, o_started_at))
+                    .then(Ord::cmp(name, o_name))
+            })
```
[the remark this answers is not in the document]
c-nixon: Replaced with a reduce rather than a sort followed by a next, that was an old pythonism leaking into my rust :)

common/k8s/src/event_source.rs
```diff
@@ -383,7 +388,17 @@ impl K8sEventStream {
...
         };
-        try_unfold((), waiter).map(|r: Result<(), K8sEventStreamError>| match r {
+
+        let waiter = Arc::new(waiter);
+
+        try_unfold((), {
+            let waiter = waiter.clone();
+            move |_| {
+                let waiter = waiter.clone();
+                backoff::future::retry(ExponentialBackoff::default(), move || waiter(()))
```
jorgebay, line 398: Nice!

bin/src/main.rs
```diff
@@ -164,21 +164,34 @@ fn main() {
...
+                        if pn.is_none() {
+                            warn!("Kubernetes event logging is configured, but POD_NAME env is not set")
+                        }
+                        if n.is_none() {
+                            warn!("Kubernetes event logging is configured, but NAMESPACE env is not set")
+                        }
+                        if pl.is_none() {
+                            warn!("Kubernetes event logging is configured, but POD_APP_LABEL env is not set")
+                        }
+                        warn!("Kubernetes event logging disabled");
```
jorgebay, line 186: Good idea!

jorgebay (approved):
Nice!

PR #140, opened 2021-05-03
Port k8s tests to run real pods

c-nixon:
This PR ports the k8s integration tests over to run the agent inside kind.

There are quite a few moving parts, and the environment setup is a bit arcane, but it allows us to create a kubernetes cluster, deploy the agent and test pods and then testing it from outside the cluster.

To inject logs it uses socat running in a pod to receive lines from the test binary over a tcp socket and write it to the socat pod's stdout. This allows us to inject lines from the test binary into the socat pod, which are picked up by the agent and then sent back to the mock ingester running inside the test binary.

To allow for running multiple sets of the k8s tests in parallel (like in jenkins) the kind_*.sh scripts set kind up on a custom docker network and attach the integration test container to the same network. It exports the kind kubeconfig and a derived config that works within the docker network (it updates the endpoint to the one inside the docker network).


docker/kind_start.sh
```diff
@@ -0,0 +1,45 @@
...
+then
+  >&2 kind create cluster --name $cluster_name \
+      --config=$curpath/kind/kind-config.yaml \
+      --kubeconfig=$curpath/.kind_config_host
+else
+  >&2 echo "Reusing existing cluster"
+  stat $curpath/.kind_config_host > /dev/null || exit 1
+fi
+
+api_server_node_addr=$(KUBECONFIG=$curpath/.kind_config_host \
```
c-nixon, line 31: Grab the socket addr for the kube api server from the kind cluster

docker/kind_start.sh
```diff
@@ -0,0 +1,45 @@
...
+    grep kubeadm.kubernetes.io/kube-apiserver.advertise-address.endpoint | \
+    awk '{ print length, $2}' | \
+    sort -n | \
+    cut -d" " -f2- | \
+    head -n1 | \
+    tr -d ',' | tr -d '"')
+
+api_server_node_port=$(echo $api_server_node_addr | cut -f2 -d":")
+
+sed "s#server: https://.*#server: https://${cluster_name}-control-plane:$api_server_node_port#" $curpath/.kind_config_host > $curpath/.kind_config
```
c-nixon, line 42: Modify the kind_config KUBECONFIG file to use the kind control plane docker container as the api server endpoint. This allows the test binary docker container to access the k8s api

jmoses (commented):
I look forward to maybe learning enough rust to dangerously add tests cases :D

Dockerfile.debian
```diff
@@ -0,0 +1,73 @@
+# syntax = docker/dockerfile:1.0-experimental
```
c-nixon, line 1: This is mostly a copy of our k8s image dockerfile but with the 2nd half ported to debian for ease of use during testing.

Docker doesn't seem to have a mechanism for #include'ing other dockerfiles so I just copy and pasted the first half to avoid having to rewrite large sections of the Makefile

bin/tests/common.rs
```diff
@@ -153,7 +152,18 @@ impl<'a> AgentSettings<'a> {
 }
 
 pub fn spawn_agent(settings: AgentSettings) -> Child {
-    let mut cmd = Command::cargo_bin("logdna-agent").unwrap();
```
c-nixon, line 156: Using escargot instead of just assert_cmd's cargo support gives us a lot more control over the integration test build, specifically it allows us to select the optimisation level of the target

bin/tests/retries.rs
```diff
@@ -36,9 +35,12 @@ async fn test_retry_after_timeout() {
             *counter += 1;
             if *counter < attempts {
                 // Sleep enough time to mark the request as timed out by the client
-                thread::sleep(Duration::from_millis(timeout + 20));
+                return Some(Box::pin(tokio::time::sleep(Duration::from_millis(
```
c-nixon, line 38: returning a future allows us to use tokio to sleep the task without blocking the integration test main thread

common/k8s/Cargo.toml
```diff
@@ -22,14 +22,14 @@ tokio = { version = "1", features = ["rt", "rt-multi-thread"] }
 futures = "0.3"
 thiserror = "1.0"
 parking_lot = "0.11"
-kube = { version = "0.52", default-features = false, features = ["rustls-tls"] }
-kube-runtime = { version = "0.52", default-features = false, features = ["rustls-tls"] }
-k8s-openapi = { version = "0.11", default_features = false, features = ["v1_12"] }
+kube = { version = "0.59", default-features = false, features = ["rustls-tls"] }
```
c-nixon, line 25: Updating to kube >= 0.59 resolves some issues for the agent running in kubernetes clusters that don't expose dns names for their api-servers

common/k8s/src/event_source.rs
```diff
@@ -240,7 +240,7 @@ impl K8sEventStream {
         pod_label: String,
     ) -> Result<Self, K8sError> {
         Ok(Self {
-            client: Client::new(config.try_into()?),
```
c-nixon, line 243: All of the changes in common/k8s are just updating to the new kube-rs api. There are no changes here

common/test/mock-ingester/src/lib.rs
```diff
@@ -25,7 +25,9 @@ use tokio_stream::wrappers::TcpListenerStream;
 const ROOT: &str = "/logs/agent";
 
 pub type FileLineCounter = Arc<Mutex<HashMap<String, FileInfo>>>;
-pub type ProcessFn = Box<dyn Fn(&IngestBody) + Send + Sync>;
+pub type ProcessFn = Box<
+    dyn (Fn(&IngestBody) -> Option<Pin<Box<dyn Future<Output = ()> + Send + Sync>>>) + Send + Sync,
```
c-nixon, line 29: Allow the process fn to optionally return a future to be awaited

docker/kind/agent-dev-build.yaml
```diff
@@ -1,113 +0,0 @@
----
```
c-nixon, line 1: This was inlined into the k8s test module as rust code to allow it to be templated easily

docker/kind/kind-config.yaml
```diff
@@ -4,4 +4,5 @@ featureGates:
   "CRIContainerLogRotation": true
 nodes:
   - role: control-plane
-  - role: worker
\ No newline at end of file
+  - role: worker
```
c-nixon, line 7: This configures kind to use 2 workers so that we can observe the agents handing over event handling duties when one is destroyed

common/test/mock-ingester/src/lib.rs
dhable: Would it make sense to define the return type separately?
```
pub type ProcessRet = Option<Pin<Box<dyn Future<Output = ()> + Send + Sync>>>;
pub type ProcessFn = Box <dyn (Fn(&IngestBody) -> ProcessRet) + Send + Sync>;
```

dhable (approved):

PR #141, opened 2021-05-04
added new regex section to readme.md

tanberry:


docs/README.md
```diff
@@ -1,4 +1,4 @@
-# LogDNA Agent
+git branch# LogDNA Agent
```
jorgebay, line 1: Probably unintended :)

docs/README.md
```diff
@@ -29,6 +29,7 @@ The LogDNA agent is a resource-efficient log collection client that forwards log
   * [Configuring Lookback](#configuring-lookback)
   * [Configuring Journald](#configuring-journald)
   * [Configuring Kubernetes Events](#configuring-events)
+  * [Configuring regex for redaction and exclusion or inclusion](#configuring-regex-for-redaction-and-exclusion-or-inclusion)
```
jorgebay, line 32: The anchor needs updating :)

docs/README.md
```diff
@@ -212,6 +213,22 @@ __Note:__ The default option is `never`.
...
+### Configuring regex for redaction and exclusion/inclusion
+
+You can define rules, using **regex** (regular expressions), to control what log data is ingested:
+ * include *only* specific log lines (with LOGDNA_LINE_INCLUSION_REGEX)
+ * exclude specific log lines (with LOGDNA_LINE_EXCLUSION_REGEX)
+ * redact parts of a log line (with LOGDNA_REDACT_REGEX)
+
+For example, you can identify certain types of logs that are noisy and not needed at all, and then write a regex pattern to match those files and preclude them from ingestion. Conversely, you can use regex expressions to match the log lines that you DO want in to include, and ingest only those log lines. Additionally, you can use the environment variable LOGDNA_REDACT_REGEX to remove certain parts of a log line. Any redacted data is replaced with [REDACTED].
+
+To access our library of common regex patterns, refer to [our regex library documentation](REGEX.md).
```
jorgebay, line 225: Makes sense, I'll send a PR with patterns using the same file name.

docs/README.md
```diff
@@ -212,6 +213,22 @@ __Note:__ The default option is `never`.
 
 > :warning: Due to a ["won't fix" bug in the Kubernetes API](https://github.com/kubernetes/kubernetes/issues/41743), the LogDNA agent collects events from the entire cluster, including multiple nodes. To prevent duplicate logs when running multiple pods, the LogDNA agent pods defer responsibilty of capturing events to the oldest pod in the cluster. If that pod is down, the next oldest LogDNA agent pod will take over responsibility and continue from where the previous pod left off.
 
+### Configuring regex for redaction and exclusion/inclusion
+
+You can define rules, using **regex** (regular expressions), to control what log data is ingested:
+ * include *only* specific log lines (with LOGDNA_LINE_INCLUSION_REGEX)
```
jorgebay, line 219: Maybe we can use backticks for env var names, wdyt? like `LOGDNA_LINE_INCLUSION_REGEX`

docs/README.md
```diff
@@ -212,6 +213,22 @@ __Note:__ The default option is `never`.
...
+ * exclude specific log lines (with LOGDNA_LINE_EXCLUSION_REGEX)
+ * redact parts of a log line (with LOGDNA_REDACT_REGEX)
+
+For example, you can identify certain types of logs that are noisy and not needed at all, and then write a regex pattern to match those files and preclude them from ingestion. Conversely, you can use regex expressions to match the log lines that you DO want in to include, and ingest only those log lines. Additionally, you can use the environment variable LOGDNA_REDACT_REGEX to remove certain parts of a log line. Any redacted data is replaced with [REDACTED].
+
+To access our library of common regex patterns, refer to [our regex library documentation](REGEX.md).
+
+Notes:
+   * Inclusion rules overwrite exclusion rules
+   * To preclude entire log files from being monitored, use the LOGDNA_EXCLUSION_RULES or the LOGDNA_EXCLUSION_REGEX_RULES environment variable.
```
jorgebay, line 229: Good idea to remind the user about file exclusion.

(TIL _preclude_)

docs/README.md
```diff
@@ -212,6 +213,22 @@ __Note:__ The default option is `never`.
...
+ * include *only* specific log lines (with LOGDNA_LINE_INCLUSION_REGEX)
+ * exclude specific log lines (with LOGDNA_LINE_EXCLUSION_REGEX)
+ * redact parts of a log line (with LOGDNA_REDACT_REGEX)
+
+For example, you can identify certain types of logs that are noisy and not needed at all, and then write a regex pattern to match those files and preclude them from ingestion. Conversely, you can use regex expressions to match the log lines that you DO want in to include, and ingest only those log lines. Additionally, you can use the environment variable LOGDNA_REDACT_REGEX to remove certain parts of a log line. Any redacted data is replaced with [REDACTED].
+
+To access our library of common regex patterns, refer to [our regex library documentation](REGEX.md).
+
+Notes:
+   * Inclusion rules overwrite exclusion rules
```
jorgebay, line 228: This is not entirely correct.

For a line to be ingested, it should match all inclusion rules (if any) and not match any exclusion rule.

For example, if we use `"warning"` as inclusion regex and `"debug"` as exclusion regex (as in, "I want only warnings when those are not included in debug messages"), the following will occur:

- "warning about something" -> will match
- "this is a debug message about a warning" -> will not match
- "something" -> will not match (doesn't match inclusion)

If we want to summarize it,  we can say that exclusion rules overwrite inclusion rules.

docs/README.md
```diff
@@ -212,6 +213,22 @@ __Note:__ The default option is `never`.
...
+   * To preclude entire log files from being monitored, use the LOGDNA_EXCLUSION_RULES or the LOGDNA_EXCLUSION_REGEX_RULES environment variable.
+   * Note that we use commas as separators for environment variable values, making it not possible to use the comma character (,) as a valid value. We are addressing this limitation in upcoming versions. If you need to use the comma character in a regular expression, use the unicode character reference: `\u002C`, for example: `hello\u002C world` matches `hello, world`.
```
jorgebay, line 230: We could add a separate bullet point about case sensitivity, something like:

> All regular expressions are case sensitive by default. If you don't want to differentiate between upper and lower-case letters, use non-capturing groups with a flag: `(?flags:exp)`, for example: (?i:my_case_insensitive_regex)

jorgebay (commented):
It's great to have a dedicated section for this.

Sorry about the number of remarks, it's a _meaty_ topic :)

docs/README.md
tanberry: I wasn't sure how to handle the special character (/) in the title...so I just changed the heading to match this link in ToC. But we can make heading whatever we want, let me know if we need to tweak.

docs/README.md
tanberry: yep, thanks, done.

docs/README.md
tanberry: Glad you caught that, thanks. Fixed it.

tanberry: Oh... I wonder if I meant to say "Inclusion rules **override** (not overwrite) exclusion rules." EIther way, I reworded it as you suggested, that's better.

jorgebay (approved):
Looks great! 👍

c-nixon (approved):
This looks great to me!

PR #143, opened 2021-05-05
Fix redaction insert index

jorgebay:
Break early when insert index is defined.

It fixes an edge case when multiple redactions are defined and multiple matches appear in the same line in a specific order.
It includes a unit test that was failing before the fix.


c-nixon (approved):
Nice, that was a subtle one

PR #152, opened 2021-05-12
Regex section docs

jorgebay:


docs/README.md
```diff
@@ -219,23 +219,25 @@ You can define rules, using regex (regular expressions), to control what log dat
...
- * include *only* specific log lines (with `LOGDNA_LINE_INCLUSION_REGEX`)
- * exclude specific log lines (with `LOGDNA_LINE_EXCLUSION_REGEX`)
- * redact parts of a log line (with `LOGDNA_REDACT_REGEX`)
- * preclude entire log files from being monitored `(LOGDNA_EXCLUSION_REGEX_RULES)`
 
-To access our library of common regex patterns, refer to [our regex library documentation](REGEX.md).
+* include *only* specific log lines (with `LOGDNA_LINE_INCLUSION_REGEX`)
+* exclude specific log lines (with `LOGDNA_LINE_EXCLUSION_REGEX`)
+* redact parts of a log line (with `LOGDNA_REDACT_REGEX`)
+* preclude entire log files from being monitored `(LOGDNA_EXCLUSION_REGEX_RULES)`
```
jorgebay, line 231: I would avoid mentioning here `LOGDNA_EXCLUSION_REGEX_RULES` (and `LOGDNA_INCLUSION_REGEX_RULES`) that targets file names. Include and exclude lists for directories and files are usually more comfortable to be specified using glob patterns.

I think it's better if we dedicate this section entirely to exclusion/inclusion/redaction of log lines instead.

tanberry: OK, sounds reasonable.

tanberry (approved):
Thank you Jorge!

docs/README.md
jorgebay: I'll remove it 👍

docs/README.md
```diff
@@ -217,25 +217,26 @@ __Note:__ The default option is `never`.
...
 
 Notes:
-   * Exclusion rules overwrite inclusion rules. That is, for a line to be ingested, it should match all inclusion rules (if any) and **not match** any exclusion rule.
-   * Note that we use commas as separators for environment variable values, making it not possible to use the comma character (,) as a valid value. We are addressing this limitation in upcoming versions. If you need to use the comma character in a regular expression, use the unicode character reference: `\u002C`, for example: `hello\u002C world` matches `hello, world`.
-   * All regular expressions are case sensitive by default. If you don't want to differentiate between upper and lower-case letters, use non-capturing groups with a flag: (?flags:exp), for example: (?i:my_case_insensitive_regex)
-   * LogDNA also provides post-ingestion <a href ="https://docs.logdna.com/docs/excluding-log-lines" target = "_blank">exclusion rules</a>, to control what log data is displayed and stored in LogDNA.
+
+* Exclusion rules overwrite inclusion rules. That is, for a line to be ingested, it should match all inclusion rules (if any) and **not match** any exclusion rule.
```
c-nixon, line 236: Maybe instead of "overwrite" we could have something like "apply after"

c-nixon (commented):
All looks good to me bar a minor wording suggestion.

docs/README.md
tanberry: Exclusion rules are applied after inclusion rules, and thus override inclusion rules. That is, for a line to be ingested, it should match all inclusion rules (if any) and **not match** any exclusion rule.

c-nixon (approved):

PR #153, opened 2021-05-12
Expose metrics in a Prometheus endpoint

jorgebay:
- Optionally expose a Prometheus endpoint.
- Add documentation for metrics.
- Create additional metrics, like a gauge for open files and a latency histogram.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -2,7 +2,6 @@
 extern crate log;
 
 use std::path::PathBuf;
-use std::thread::spawn;
 
 use futures::Stream;
 
@@ -65,8 +64,6 @@ async fn main() {
             std::process::exit(1);
         }
     };
-
-    spawn(Metrics::start);
 
     let mut _agent_state = None;
     let mut offset_state = None;
@@ -256,6 +253,20 @@ async fn main() {
             Either::Right(_) => client.borrow_mut().poll().await,
         }
     });
+
+    tokio::spawn(async {
+        Metrics::log_periodically().await;
+    });
+
+    if let Some(port) = config.log.metrics_port {
+        info!("Enabling prometheus endpoint with agent metrics");
+        tokio::spawn(async move {
+            // Should panic when server exits
+            http::metrics_endpoint::serve(&port)
+                .await
+                .expect("metrics server error");
+        });
+    }
 
     // Concurrently run the line streams and listen for the `shutdown` signal
     tokio::select! {
--- a/bin/tests/cli.rs
+++ b/bin/tests/cli.rs
@@ -419,7 +419,7 @@ async fn test_metrics_endpoint() {
             body_str = std::str::from_utf8(&buf).unwrap().to_string();
 
             // Request duration metrics are the last ones to appear
-            if body_str.contains("logdna_ingest_request_duration") {
+            if body_str.contains("logdna_agent_ingest_request_duration") {
                 break;
             }
         }
@@ -427,13 +427,13 @@ async fn test_metrics_endpoint() {
         tokio::time::sleep(Duration::from_millis(100)).await;
     }
 
-    assert!(body_str.contains("# TYPE logdna_fs_bytes counter"));
-    assert!(body_str.contains("# TYPE logdna_fs_lines counter"));
-    assert!(body_str.contains("# TYPE logdna_ingest_request_size histogram"));
-    assert!(body_str.contains("# TYPE logdna_ingest_request_duration_millis histogram"));
-    assert!(body_str.contains("# TYPE logdna_fs_events counter"));
+    assert!(body_str.contains("# TYPE logdna_agent_fs_bytes counter"));
+    assert!(body_str.contains("# TYPE logdna_agent_fs_lines counter"));
+    assert!(body_str.contains("# TYPE logdna_agent_ingest_request_size histogram"));
+    assert!(body_str.contains("# TYPE logdna_agent_ingest_request_duration_millis histogram"));
+    assert!(body_str.contains("# TYPE logdna_agent_fs_events counter"));
     // One created file
-    assert!(body_str.contains("logdna_fs_events{event_type=\"create\"} 1"));
+    assert!(body_str.contains("logdna_agent_fs_events{event_type=\"create\"} 1"));
 
     common::assert_agent_running(&mut agent_handle);
     agent_handle.kill().expect("Could not kill process");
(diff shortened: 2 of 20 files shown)


docs/INTERNAL_METRICS.md
```diff
@@ -0,0 +1,52 @@
+# LogDNA Agent Internal Metrics
```
jorgebay, line 1: I've used the word "internal" to avoid misleading the user into thinking that the logdna agent gathers metrics from other pods.

common/metrics/src/lib.rs
```diff
@@ -1,21 +1,82 @@
...
+    register_int_counter_vec, register_int_gauge, Histogram, HistogramVec, IntCounter,
+    IntCounterVec, IntGauge,
+};
+use std::time::{Duration, Instant};
+use tokio::time::sleep;
 
 lazy_static! {
     static ref METRICS: Metrics = Metrics::new();
+    static ref FS_EVENTS: IntCounterVec = register_int_counter_vec!(
+        "logdna_fs_events",
```
jorgebay, line 17: I'm starting to think that `logdna_agent_` is a better prefix. Having `company_product_` prefixes avoids collisions in the future.

bin/src/main.rs
```diff
@@ -2,7 +2,6 @@
 extern crate log;
 
 use std::path::PathBuf;
-use std::thread::spawn;
```
c-nixon, line 5: 🎉

common/http/src/metrics_endpoint.rs
```diff
@@ -0,0 +1,39 @@
...
+use prometheus::{Encoder, TextEncoder};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    Server(#[from] hyper::Error),
+}
+
+pub async fn serve(port: &u32) -> Result<(), Error> {
```
c-nixon, line 16: extremely minor, but ports are usually u16s. It's not worth changing though.

common/http/src/metrics_endpoint.rs
```diff
@@ -0,0 +1,39 @@
...
+    let address = format!("0.0.0.0:{}", port).parse().unwrap();
```
c-nixon, line 17: Again extremely minor nit

going through the format and socket addr parse machinery then unwrapping to get a socketaddr feels awkward when we could do something like
`
SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), port)
`
even if the unwrap can't fail

common/metrics/src/lib.rs
c-nixon: I agree on that

common/metrics/src/lib.rs
```diff
@@ -79,44 +123,57 @@ impl Metrics {
...
-                "lines" => fs.read_lines(),
-                "bytes" => fs.read_bytes(),
-                "partial_reads" => fs.read_partial_reads(),
+                "events" => fs_create + fs_delete + fs_write,
+                "creates" => fs_create,
+                "deletes" => fs_delete,
+                "writes" => fs_write,
+                "lines" => FS_LINES.get(),
+                "bytes" => FS_BYTES.get(),
+                "files_tracked" => FS_FILES.get(),
```
c-nixon, line 145: nice addition

docs/INTERNAL_METRICS.md
```diff
@@ -0,0 +1,52 @@
...
+```yaml
+    - env:
+        - name: LOGDNA_METRICS_PORT
+          value: "9881"
+```
+
+Metrics related to the agent are exposed using the prefix `logdna_` and process status information, like memory and
+cpu usage, are exposed with the prefix `process_`.
+
+## Enabling Prometheus target discovery on Kubernetes
```
c-nixon, line 20: It may be worth adding something about PodMonitors iff the customer is running the prometheus operator (they might be confused and try to use a ServiceMonitor)

https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmonitor

https://github.com/prometheus-operator/prometheus-operator/blob/master/example/user-guides/getting-started/example-app-pod-monitor.yaml

docs/INTERNAL_METRICS.md
c-nixon: Maybe we could just have the "exposing Agent Metrics as prometheus endpoint" as the top level title? I'm not sure most customers realise the agent ships metrics as log lines

bin/tests/cli.rs
```diff
@@ -384,6 +385,60 @@ fn test_exclusion_rules() {
...
+    common::wait_for_file_event("tailer sendings lines", &included_file, &mut stderr_reader);
+    let mut body_str = String::new();
+
+    // Wait for all the metrics to be tracked
+    for _ in 0..20 {
+        let client = Client::new();
+        let url = format!("http://127.0.0.1:{}/metrics", port)
+            .parse()
+            .unwrap();
+        if let Ok(response) = client.get(url).await {
```
c-nixon, line 415: Should this just unwrap? it looks like it will skip the results if client.get fails.

c-nixon (commented):
Couple of questions and some minor nits, no changes I think are mandatory, this is really good.

common/http/src/metrics_endpoint.rs
jorgebay: good point... got me thinking about the actual setting exposed.
I'll change it to `u16` everywhere.

bin/tests/cli.rs
jorgebay: yeah, that's expected, it's a way to wait for the endpoint to be ready (I've initially got failures because it was taking ~200ms for endpoint to accept connections).

If it fails several times, the test will fail (`body_str` will be empty).

docs/INTERNAL_METRICS.md
jorgebay: I wanted to include the log lines metrics information here because it changes behaviour from the previous version. It can be unexpected change for users that are parsing that information (hopefully few users), but at least we have a doc page where we can redirect them.

docs/INTERNAL_METRICS.md
jorgebay: I think the default installation of prometheus and vendors have discovery based on `prometheus.io/port` and `prometheus.io/scrape` enabled by default (I've tried vanilla prometheus server and sysdig), so we hopefully we can avoid documenting here prometheus server settings :)

c-nixon (approved):
This all looks good now. I've been running on top of it locally and it all seems good!

PR #156, opened 2021-05-17
Journald: use a warn interval

jorgebay:
On some environments, seeking the tail of journald has no effect. I've only was able to reproduce it on IKS.

For example, these are journald messages 15+ days old:

Screenshot 2021-05-13 at 11 11 47

In those case, we should continue processing the journald messages and avoid spamming the logs with info messages.

This PR prevents spamming the logs, producing only a maximum of a info every 5mins.

Before (more than info messages per second):

Screenshot 2021-05-17 at 15 20 39

After the patch (set to 5s instead of 5m):

Screenshot 2021-05-17 at 15 18 35

### DIFF
--- a/common/journald/src/stream.rs
+++ b/common/journald/src/stream.rs
@@ -3,6 +3,7 @@ use futures::{channel::oneshot, stream::Stream as FutureStream};
 use http::types::body::LineBuilder;
 use log::{info, warn};
 use metrics::Metrics;
+use std::time::Instant;
 use std::{
     mem::drop,
     path::PathBuf,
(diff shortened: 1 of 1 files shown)


c-nixon (approved):
> only was able to reproduce it on IKS

Of course.

Do you know if it duplicates messages or anything on the systems where it fails to seek?

Regardless, I think this is the best we can do.

jorgebay:
> if it duplicates messages or anything on the systems where it fails to seek?

my guess is on seek but don't know for sure.

PR #157, opened 2021-05-19
Use debug level for old move events and invalid path errors

jorgebay:
Move events processed after parents have been removed, along with other path errors, were causing noise in the agent logs.

This patch creates a dedicated message for old move events and logs path errors (with the path) at debug level.


c-nixon (approved):
Aha, another edge case.

Well caught.

Issue #158, opened 2021-05-20
Collect Log from Specific APPs only


moficodes:
I have logdna agent installed on a k8s cluster.

I only want to collect logs for a few specific apps.

1. Do I need to write the log in any specific way? Right now the logs are written to stdout. (I can see the log in /var/log/pods, or /var/data/kubeletlogs) 
2. I do not want to collect logs for any other pods. like kubelet or coredns.  I tried adding this exclusion rule. Then adding inclusion rule for just my app, not sure if i did that right.
```yaml
            - name: LOGDNA_EXCLUSION_RULES
              value: /var/log/**/*
            - name: LOGDNA_LOG_DIRS
              value: /var/log/pods/kubeadmin*/*
```

Any guidance/help is appreciated.

jorgebay:
Thanks for bringing this up @moficodes.

Include only specific application or pods is currently possible using negative exclusion rules, as in "exclude everything that does not match X".

Leaving the rest of the settings unset, only define the `LOGDNA_AGENT_KEY` and the exclusion rules, for example the following will only match "my-app":

```yaml
            - name: LOGDNA_EXCLUSION_RULES
              value: "!(*my-app*)"
```

jorgebay:
I've added some coverage to this type of rules in #159 and filed an internal ticket for better documenting this.

jorgebay:
@moficodes let us know if you want to keep this ticket open until the doc change or if we can go ahead and close it.

jorgebay:
I'll close this one for now, if you have further questions or suggestions, feel free to reopen it.

PR #159, opened 2021-05-20
Test include-only rules

jorgebay:
Adds coverage for "include-only" type of rules, using negative exclusion rules.

Related to #158.


c-nixon (approved):
Very nice, hadn't thought of that combination

PR #165, opened 2021-06-07
Support retries when fetching initial pod metadata

jorgebay:
Sometimes the kube api might not be accessible as network
is not fully initialized (CNI). Retry with backoff and exit
in error when all attempts fail.


c-nixon (approved):
No comments, this looks like a great addition

PR #167, opened 2021-06-09
Add journalctl source

c-nixon:
WIP, still splitting it out from the static build changes


common/journald/src/journalctl/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+     Journal fields consisting only of valid non-control UTF-8 codepoints are serialized as they are:
+         field name, followed by '=', followed by field data, followed by a newline as separator to the next field.
+     Other journal fields are serialized in a special binary safe way:
+         field name, followed by newline, followed by a binary 64bit little endian size value, followed by the binary field data,
+         followed by a newline as separator to the next field.
+     Entry metadata that is not actually a field is serialized like it was a field, but beginning with two underscores.
+     More specifically, __CURSOR=, __REALTIME_TIMESTAMP=, __MONOTONIC_TIMESTAMP= are introduced this way.
+     The order in which fields appear in an entry is undefined and might be different for each entry that is serialized.
+
+     And that's it.
```
jorgebay, line 84: Good source code doc, I had to go over it a few times :)

common/journald/src/journalctl/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+                    // Get the key
+                    recognize(skip_many1(choice((letter(), token(b'_'))))).and_then(
+                        |bytes: &[u8]| {
+                            std::str::from_utf8(bytes)
+                                .map(|s| s.to_string())
+                                .map_err(StreamErrorFor::<Input>::other)
+                        },
+                    ),
+                    // Get the Value
+                    one_of(b"=\n"[..].iter().copied()).then_partial(|&mut c| {
```
jorgebay, line 100: Pretty elegant, using `combine` is a good call.

jorgebay (commented):
It's looking good 👍

common/journald/src/journalctl/mod.rs
c-nixon: The original manual version wasn't too bad until I started trying to support partial reads, then it got very big, ugly and hard to follow.

Combine has it's own awkwardness, the lifetimes held me up for a few hours until I worked out how partial state works, but I agree, I still think it's worth it for the elegance.

common/journald/src/journalctl/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+                    .map(|(key, value, _)| (key, value)),
+            ),
+            // entries are line separated
+            token(b'\n'),
+        )
+            .map(|(kvs, _)| kvs),
+    )
+}
+
+impl JournaldExportDecoder {
```
c-nixon, line 133: This is unfortunately alloc'y I think I know how to get it working with the client's lazy lines support, but it pollutes the decoder logic a bit so it's probably best left until later.

bin/src/main.rs
```diff
@@ -149,7 +151,20 @@ async fn main() {
     );
 
     #[cfg(feature = "libjournald")]
-    let journald_source = create_source(&config.journald.paths);
+    let (journalctl_source, journald_source) = if config.journald.paths.is_empty() {
+        let journalctl_source = create_journalctl_source()
+            .map(|s| s.map(StrictOrLazyLineBuilder::Strict))
+            .map_err(|e| warn!("Error initializing journalctl source: {}", e));
```
jorgebay, line 157: if the user didn't manually defined journald paths, I think this shouldn't be an error log.

We could do something like:
```rust
info!("Journalctl source was not initialized");
debug!("Journalctl source Initialization error was: {}", e);
```

jorgebay (commented):
Looks great, a minor suggestion below.

jorgebay (approved):

PR #168, opened 2021-06-10
Track file changes with and without inotify

jorgebay:
Replace `inotify` dependency with `notify` and handle the behavioural changes.

macOS watchers are default to polling as `FSEvents` missed a lot of changes.

It's now possible to build on macos and linux:

```bash
cargo build --no-default-features
```

and run integration tests:

```bash
cargo test --features integration_tests  --test cli test_tags  --no-default-features
```

I've wrapped `notify` library to simplify migrating from library versions (a new major version is coming). Our `notify_stream` crate doesn't leak any `notify` type.

All `fs` crate tests are passing.
```
cargo test -p fs
```

There's still some work to do to make all the integration tests pass on both linux and macOS (mainly around timing).


jorgebay:
Rebased and squashed.

common/fs/src/cache/mod.rs
```diff
@@ -474,167 +314,192 @@ impl FileSystem {
...
-                        error!("attempted to remove untracked symlink {:?}", path);
-                        return;
-                    }
-                };
-
-                entries.retain(|other| *other != entry_ptr);
-                if entries.is_empty() {
-                    self.symlinks.remove(link);
-                }
+                self.symlinks.remove(link);
```
dkhokhlov: portable? (symlinks & windows)

PR #170, opened 2021-06-14
Musl static build

c-nixon:
This adds the ability to build a completely static agent binary using musl.

The binary is still built in docker using the existing build-image command. to configure it set the STATIC env var to 1 when calling make build-image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN env
 # Rebuild the agent
 RUN --mount=type=secret,id=aws,target=/root/.aws/credentials \
     if [ -z "$SCCACHE_BUCKET" ]; then unset RUSTC_WRAPPER; fi; \
-    export ${BUILD_ENVS}; cargo build --manifest-path bin/Cargo.toml ${FEATURES} --release --target ${TARGET} && \
+    export ${BUILD_ENVS?}; cargo build --manifest-path bin/Cargo.toml ${FEATURES} --release --target ${TARGET} && \
     strip ./target/${TARGET}/release/logdna-agent && \
     cp ./target/${TARGET}/release/logdna-agent /logdna-agent && \
     sccache --show-stats
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,6 @@ RUN --mount=type=secret,id=aws,target=/root/.aws/credentials \
     strip ./target/${TARGET}/release/logdna-agent && \
     cp ./target/${TARGET}/release/logdna-agent /logdna-agent && \
     sccache --show-stats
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -14,18 +15,31 @@ ENV SCCACHE_BUCKET=${SCCACHE_BUCKET}
...
 COPY . .
 
+RUN env
 # Rebuild the agent
 RUN --mount=type=secret,id=aws,target=/root/.aws/credentials \
     if [ -z "$SCCACHE_BUCKET" ]; then unset RUSTC_WRAPPER; fi; \
-    cargo build --manifest-path bin/Cargo.toml ${FEATURES} --release && strip ./target/release/logdna-agent && \
+    export ${BUILD_ENVS}; cargo build --manifest-path bin/Cargo.toml ${FEATURES} --release --target ${TARGET} && \
+    strip ./target/${TARGET}/release/logdna-agent && \
+    cp ./target/${TARGET}/release/logdna-agent /logdna-agent && \
```
jorgebay, line 41: oh, good idea to simplify resulting path here.

jorgebay (approved):

Makefile
```diff
@@ -287,6 +313,10 @@ build-image: ## Build a docker image as specified in the Dockerfile
 		--build-arg SCCACHE_BUCKET=$(SCCACHE_BUCKET) \
 		--build-arg SCCACHE_REGION=$(SCCACHE_REGION)
 
+.PHONY: publish-s3-binary
+publish-s3-binary:
+	aws s3 cp target/$(TARGET)/release/logdna-agent s3://logdna-agent-build-bin/$(TARGET_TAG)/$(TARGET)/logdna-agent
```
jorgebay, line 318: Use --acl public-read to allow public reading

Makefile
```diff
@@ -273,11 +296,14 @@ release: ## Create a new release from the current beta and push to github
 .PHONY:build-image
 build-image: ## Build a docker image as specified in the Dockerfile
 	$(DOCKER) build . -t $(REPO):$(BUILD_TAG) \
-		--pull \
+		$(PULL_OPTS) \
```
jorgebay, line 299: Curious, why the change to pull newer image?

jorgebay (approved):
lgtm!

Just a question below.

Makefile
c-nixon: This just makes it use the PULL_OPTS option that was already in the makefile, it shouldn't change the behaviour except that it means we can override it by setitng PULL_OPTS=0

jorgebay: makes sense

PR #172, opened 2021-06-24
Support conf file format and location from the node.js agent

jorgebay:
For backward-compatibility with agent v1, support the same format (java properties) and location (`/etc/logdna.conf`) for the config file, while maintaining support for the rust agent yaml.

Validation for the format is a little tricky because both yaml and properties formats are permissive.

TODO:

- [x] Integration tests


common/config/Cargo.toml
```diff
@@ -19,6 +19,7 @@ async-compression = "0.3"
 log = "0.4"
 sysinfo = "0.15"
 structopt = "0.3"
+java-properties = "1"
```
c-nixon, line 22: Of course there's a crate for it :)

common/config/src/properties.rs
```diff
@@ -0,0 +1,233 @@
...
+from_env_name!(DB_PATH);
+from_env_name!(METRICS_PORT);
+from_env_name!(USE_K8S_LOG_ENRICHMENT);
+from_env_name!(LOG_K8S_EVENTS);
+from_env_name!(LINE_EXCLUSION);
+from_env_name!(LINE_INCLUSION);
+from_env_name!(REDACT);
+
+enum Key {
+    FromEnv(&'static str),
```
c-nixon, line 49: Given these are both wrapping up &'static str would it make sense to do

```
struct Key(String);

impl FromStr for Key {
    fn parse(...: &'str) -> Self {
        ...
    }
}

```
That would let us parse once on creation of the key and also let us drop the static bounds

c-nixon: Although, reading it, it would require the from_env_name to be lazy statics wouldn't it. We might be able to use a const fn to build the new keys...

c-nixon: Nope! not yet, and the various macro libraries don't seem to make this possible either. Comments retracted as not worthwhile!

jorgebay: yeah, I give const fn a try and couldn't do it :)

common/config/src/raw.rs
```diff
@@ -15,8 +14,42 @@ pub struct Config {
 }
 
 impl Config {
+    /// Tries to parse from java properties format and then using
     pub fn parse<P: AsRef<Path>>(path: P) -> Result<Self, ConfigError> {
-        Ok(serde_yaml::from_reader(File::open(path)?)?)
+        let is_default_path = path.as_ref().to_string_lossy() == argv::DEFAULT_YAML_FILE;
+        let mut file = match File::open(path) {
```
c-nixon, line 20: Style nit, I had some difficulty reading this at first glance but that might be because I've spent too long in FP land.

```
let mut file = File::open(path).or_else(|e|
    if !is_default_path {
        // A path was configured but the file was NOT found
        Err(e)
    } else {
        File::open(argv::DEFAULT_CONF_FILE)
    }
)?;
```

jorgebay: oh, good suggestion! I'll change it.

common/config/src/raw.rs
```diff
@@ -15,8 +14,42 @@ pub struct Config {
...
+            Ok(f) => f,
+            Err(e) => {
+                if !is_default_path {
+                    // A path was configured but the file was NOT found
```
c-nixon, line 24: This seems worth warning about

c-nixon (commented):
Some minor nits if you think they're worth changing, general functionality looks good and the test coverage looks great.

c-nixon (approved):
LGTM!

PR #173, opened 2021-06-29
Support listing the current options

jorgebay:
Print the current merged settings in yaml format and exit.

Useful for troubleshooting on production environments.

### DIFF
--- a/common/config/src/lib.rs
+++ b/common/config/src/lib.rs
@@ -4,7 +4,7 @@ extern crate log;
 use std::convert::{TryFrom, TryInto};
 use std::fs::File;
 use std::io::Read;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::time::Duration;
 use sysinfo::{RefreshKind, System, SystemExt};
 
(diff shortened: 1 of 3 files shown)


common/config/src/lib.rs
```diff
@@ -298,6 +312,25 @@ pub fn get_hostname() -> Option<String> {
...
 }
 
+fn print_settings(yaml: &str, config_path: &Path) {
+    print!("Listing current settings ");
+
+    let is_default_path = config_path.to_string_lossy() == argv::DEFAULT_YAML_FILE;
+
+    if config_path.exists() {
+        print!("from config ({}), ", config_path.display());
+    } else if is_default_path && Path::new(argv::DEFAULT_CONF_FILE).exists() {
```
c-nixon, line 322: Is it necessary to differentiate between the default config file and non-default config file here? Either way it's just "options from the config file"

jorgebay: I think it can be useful for troubleshooting. The agent tries to read from the `/etc/logdna.conf` when the default `/etc/logdna/config.yaml` is not found and it can be surprising.

c-nixon (approved):
This all looks good to me

c-nixon:
It might be worth checking with support if the output contains the info they expect?

PR #178, opened 2021-07-23
docs(linux): dedupe installation instructions and update conf desc

david-ly:
* Added some more detail about the compatibility of the agent with the listed Linux distros
* Deduped the install/upgrade instructions since it's safe for people in either camp to fully follow along with the appropriate instructions for their OS distribution type
*  Added some header notes in the relevant sections to cover our bases in terms of supporting these 3 end-users / scenarios
    * User doing a new / fresh install of the *Linux* agent - notes aren't applicable to them, they can & should run every command in the instructions
    * User upgrading from a older version of `logdna-agent-v2` - they can skip the `install logdna-agent` command if they've ever installed the Linux package for the current agent OR the legacy agent and just update their package. Running through these steps will also ensure that they're pulling the latest version from the correct source repository.
    * User migrating from the legacy client `logdna-agent` - similar to above, they can skip the `install logdna-agent` step if they've ever installed the Linux package and just update the agent. The most recent release of `logdna-agent-v2` automates the teardown and removal process of the legacy client as well.
* Fleshed out the description/explanation of the `logdna-agent` service managed by `systemd` & updated the usage instructions to reflect the standard configuration process for the agent
    * The default filepath for the agent's configuration file is `/etc/logdna.conf` and its expected to be plain text file which stores environment variables as key-value pairs to be passed into the agent at runtime

### DIFF
--- a/docs/LINUX.md
+++ b/docs/LINUX.md
@@ -6,9 +6,9 @@ The LogDNA Agent for Linux collects log data from your Linux environment. The ag
   * [Installation: fresh installs](#Installation)
   * [Upgrading: migrating from legacy Linux agent](#upgrade-migration)
   * [Usage](#usage)
-   * [Enable the agent](#enable-the-agent)
-   * [Configure the agent](#configure-the-agent)
-   * [Run the agent](#run-the-agent)
+    * [Enable the agent](#enable-the-agent)
+    * [Configure the agent](#configure-the-agent)
+    * [Run the agent](#run-the-agent)
 
 ## Installation
 
--- a/docs/LINUX.md
+++ b/docs/LINUX.md
@@ -1,34 +1,27 @@
 # LogDNA Agent on Linux distributions
 
-The agent is available as a Linux package for Debian-based distributions
-using apt and Red Hat-based distributions using yum. If you use a
-different package manager that does not support .deb or .rpm files, you
-can still use the agent by following the steps
-[here](https://github.com/logdna/logdna-agent-v2/blob/eb06d4f3f7c1033b494f1f0439957f96533f9225/docs/README.md#building-agent-binary-on-linux)
-to manually compile its binary.
+The LogDNA Agent for Linux collects log data from your Linux environment. The agent is available as a Linux package for Debian-based distributions using `apt` and Red Hat-based distributions using `yum`. If you use a different package manager that does not support `.deb` or `.rpm` files, you can still use the agent by [manually compiling the binary](https://github.com/logdna/logdna-agent-v2/blob/eb06d4f3f7c1033b494f1f0439957f96533f9225/docs/README.md#building-agent-binary-on-linux).
+
+## Table of Contents
+  * [Installation: fresh installs](#Installation)
+  * [Upgrading: migrating from legacy Linux agent](#upgrade-migration)
+  * [Usage](#usage)
+   * [Enable the agent](#enable-the-agent)
+   * [Configure the agent](#configure-the-agent)
+   * [Run the agent](#run-the-agent)
 
 ## Installation
-└ **Note** [for users upgrading from older versions of the [LogDNA Agent](https://github.com/logdna/logdna-agent-v2) or migrating from the [Legacy LogDNA Agent](https://github.com/logdna/logdna-agent)]:
-> Even if you have previously installed the `logdna-agent` Linux package, it is _still_ recommended to follow the appropriate instructions below to ensure that all users are retrieving the latest version of the package from the correct logdna source repository.
 
+1. To download the agent and resources package, open a host terminal and run the appropriate commands based on the Linux distribution.
 
-### _On Debian-based distributions_
+* Debian-based distributions
 ```bash
-# add logdna source repo → retrieve latest .deb file & manage package via apt
 echo "deb https://assets.logdna.com stable main" | sudo tee /etc/apt/sources.list.d/logdna.list
 wget -qO - https://assets.logdna.com/logdna.gpg | sudo apt-key add -
 sudo apt-get update
-
-# install the package via apt (skip this step if already installed)
-sudo apt-get install logdna-agent
-
-# update the package via apt
-sudo apt-get upgrade logdna-agent
 ```
-
-### _On Red Hat-based distributions_
+* RPM/Red Hat-based distributions
 ```bash
-# import latest .rpm file from logdna source repo → move into yum dir & manage package via yum
 sudo rpm --import https://assets.logdna.com/logdna.gpg
 echo "[logdna]
 name=LogDNA packages
@@ -36,22 +29,60 @@ baseurl=https://assets.logdna.com/el6/
 enabled=1
 gpgcheck=1
 gpgkey=https://assets.logdna.com/logdna.gpg" | sudo tee /etc/yum.repos.d/logdna.repo
+```
 
-# install the package via yum (skip this step if already installed)
+2. To upgrade the agent, use the commands below, based on the distro:
+
+* Debian-base distributions:
+```bash
+sudo apt-get install logdna-agent
+```
+* RPM/Red Hat-based distributions:
+```bash
 sudo yum -y install logdna-agent
+```
 
-# update the package via yum
+## Upgrade/migration
+└ **Note** [for users migrating from the [legacy LogDNA Agent](https://github.com/logdna/logdna-agent)]:
+> If you have previously installed the `logdna-agent` Linux package and have an existing `/etc/logdna.conf` file, it is _still_ recommended to follow the instructions below to ensure that all users are retrieving the latest version of the package from the correct source repository.
+
+1.  To download the agent and resources package, open a host terminal and run the appropriate commands based on the Linux distribution.
+
+* Debian-base distributions:
+```bash
+echo "deb https://assets.logdna.com stable main" | sudo tee /etc/apt/sources.list.d/logdna.list
+wget -qO - https://assets.logdna.com/logdna.gpg | sudo apt-key add -
+sudo apt-get update
+```
+
+* RPM/Red Hat-based distributions:
+```bash
+sudo rpm --import https://assets.logdna.com/logdna.gpg
+echo "[logdna]
+name=LogDNA packages
+baseurl=https://assets.logdna.com/el6/
+enabled=1
+gpgcheck=1
+gpgkey=https://assets.logdna.com/logdna.gpg" | sudo tee /etc/yum.repos.d/logdna.repo
+```
+
+2. To upgrade the agent, use the commands below, based on the distro:
+
+* Debian-base distributions:
+```bash
+sudo apt-get upgrade logdna-agent
+```
+* RPM/Red Hat-based distributions:
+```bash
 sudo yum update logdna-agent
 ```
 
 ## Usage
-
 The agent uses [**systemd**](https://systemd.io/) to run as a Linux daemon. The installed package provides a **systemd** unit file for the `logdna-agent` service, which is defined to execute a daemon process with the compiled agent binary. The process that is started in the service is managed by **systemd** and can be interfaced with using `systemctl`.
 
----
 ### _Enable the agent_
 
-1.  Reload the **systemd** unit files to ensure that the most recent version of the agent is used:
+1.  Reload the **systemd** unit files to obtain the most recent version of the agent:
 ```bash
     sudo systemctl daemon-reload
 ```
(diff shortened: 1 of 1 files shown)


docs/LINUX.md
```diff
@@ -101,25 +32,58 @@ enabled=1
...
+    ```
+    This file stores key-value pairs as environment variables which can be injected into the agent at runtime and manage a variety of configuration options. You can see all the available variable options via `logdna-agent --help` or refer to them [here](README.md#options). If you're coming from the legacy agent, take note of the variables which have been changed, updated, and deprecated when compared to [logdna-agent](https://github.com/logdna/logdna-agent).
+
+- Before starting the `logdna-agent` service, an ingestion key must be provided by setting the `LOGDNA_INGESTION_KEY` variable in `/etc/logdna.conf`:
+  
+  ```edit /etc/logdna.conf
+  LOGDNA_INGESTION_KEY=<YOUR INGESTION KEY HERE>
+  ```
+  The ingestion key is the only required variable, but users can set additional variables in order to meet their desired configuration. For example, to attach a "production" tag to every log line - you could set the `LOGDNA_TAGS` variable in `/etc/logdna.conf`:
+
```
tanberry, line 72: replace "users" with "you."

replace hyphen with a comma

tanberry (changes requested):
Thanks David for doing this work and adding more content. Let me know if you have any questions about my edits.

c-nixon (approved):
The content looks correct to me, I'm good with it once it meet's Tana's standards

docs/LINUX.md
```diff
@@ -101,25 +32,58 @@ enabled=1
...
+#
+
+- By default, `/etc/logdna.conf` is the expected path for the agent's configuration file. Make sure to create this file if it doesn't exist since it's required to start the agent.
+
+    ```bash
+    sudo touch /etc/logdna.conf # it can initalized as an empty file, you will be adding to it in the steps below
+    ```
+    This file stores key-value pairs as environment variables which can be injected into the agent at runtime and manage a variety of configuration options. You can see all the available variable options via `logdna-agent --help` or refer to them [here](README.md#options). If you're coming from the legacy agent, take note of the variables which have been changed, updated, and deprecated when compared to [logdna-agent](https://github.com/logdna/logdna-agent).
```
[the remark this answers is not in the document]
david-ly: In this case, I think we'd want to reference the legacy project for comparison since these docs would imply they're already looking at the `logdna-agent-v2` project.

docs/LINUX.md
```diff
@@ -101,25 +32,58 @@ enabled=1
...
 ```
-
-[ingestion-key]: https://docs.logdna.com/docs/ingestion-key
+#
+## Usage
+The agent uses [systemd](https://systemd.io/) to run as a Linux daemon. The installed package provides a `systemd` unit file for the `logdna-agent` service which is defined to execute a daemon process with the compiled agent binary. The process started in the service is managed by `systemd` and can be interfaced with using `systemctl`.
+
+---
+### _Enabling the agent_
+- Reload the `systemd` unit files to ensure the most recent version of the agent is used & enable the service via `systemctl`:
```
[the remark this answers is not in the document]
david-ly: Instead of replacing the ampersand & with "and to", I moved the latter part into its own bullet since I think there's a slight distinction to be made. Reloading the files isn't a requirement in enabling the service, they are 2 separate commands so I broke them up accordingly.

docs/LINUX.md
```diff
@@ -101,25 +32,58 @@ enabled=1
...
+
+  ```bash
+  sudo systemctl daemon-reload
+  sudo systemctl enable logdna-agent
+  ``` 
+---
+### _Configuring the agent_
+└ **Note** [for users upgrading from older versions of [logdna-agent-v2](https://github.com/logdna/logdna-agent-v2) or migrating from the legacy [logdna-agent](https://github.com/logdna/logdna-agent)]:
```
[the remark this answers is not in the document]
david-ly: I think using the "proper" names here might make it even more confusing. The "LogDNA Agent v2" project name already includes a version but the project itself is also versioned; throwing in the legacy agent and naming it "LogDNA Agent v1" seems like it would only cause more trouble

david-ly:
@tanberry I made most of the changes as suggested but I think referring to the legacy agent `logdna-agent` and the Rust agent `logdna-agent-v2` is probably the least confusing way to distinguish the agent for our users. There are inconsistencies in your comments about how we should be referring to the distinct agents so I left them the same for now.

docs/LINUX.md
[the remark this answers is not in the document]
tanberry: Agree, let's not use the v1 as I suggested above.

docs/LINUX.md
```diff
@@ -101,25 +32,59 @@ enabled=1
...
+### _Configuring the agent_
+└ **Note** [for users upgrading from older versions of the [LogDNA Agent](https://github.com/logdna/logdna-agent-v2) or migrating from the [Legacy LogDNA Agent](https://github.com/logdna/logdna-agent)]: 
+  > Many users will likely already have a configuration file `/etc/logdna.conf` from prior installations. While the config variable names have been changed / updated between the [Legacy LogDNA Agent](https://github.com/logdna/logdna-agent) and the current [LogDNA Agent](https://github.com/logdna/logdna-agent-v2), backwards compatibility has been maintained so older `.conf` files are still valid. If this is applicable to you and there are no changes that need to be made, you can skip this section and start the service using your existing configuration.
+#
+
+- By default, `/etc/logdna.conf` is the expected path for the agent's configuration file. Make sure to create this file if it doesn't exist since it's required to start the agent.
+
+    ```bash
+    sudo touch /etc/logdna.conf # it can initalized as an empty file, you will be adding to it in the steps below
+    ```
```
tanberry, line 64: Missing an "i" in "initialized."

docs/LINUX.md
```diff
@@ -101,25 +32,59 @@ enabled=1
...
 ```
-
-[ingestion-key]: https://docs.logdna.com/docs/ingestion-key
+#
+## Usage
+The agent uses [systemd](https://systemd.io/) to run as a Linux daemon. The installed package provides a `systemd` unit file for the `logdna-agent` service, which is defined to execute a daemon process with the compiled agent binary. The process that is started in the service is managed by `systemd` and can be interfaced with using `systemctl`.
+
+---
+### _Enabling the agent_
+- Reload the `systemd` unit files to ensure that the most recent version of the agent is used
```
tanberry, line 47: Let's rework this section so that it looks like this:

- Reload the `systemd` unit files to ensure that the most recent version of the agent is used.

  ```bash
  sudo systemctl daemon-reload

- Enable the `logdna-agent` service using the `systemctl` command.

  ```bash
  sudo systemctl enable logdna-agent

tanberry (changes requested):

tanberry (approved):
Looks good to me, tho my eyes are certainly no longer "fresh eyes" on this content, ;-)

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
+    * [Configure the agent](#configure-the-agent)
+    * [Run the agent](#run-the-agent)
 
-### On Debian-based distributions
+## Installation
 
-```shell script
+1. To download the agent and resources package, open a host terminal and run the appropriate commands based on the Linux distribution.
+
+* Debian-based distributions
```
alfeng6, line 17: style-wise, with the bullets, it's harder to read `Debian-based distributions` and `RPM/Red Hat-based distributions`. Can we bold it instead?

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
+gpgcheck=1
+gpgkey=https://assets.logdna.com/logdna.gpg" | sudo tee /etc/yum.repos.d/logdna.repo
 ```
 
-After saving the configuration, start the `logdna-agent` service using the following command:
+2. To upgrade the agent, use the commands below, based on the distro:
 
-```shell script
-sudo systemctl start logdna-agent
+* Debian-base distributions:
```
alfeng6, line 71:
```suggestion
* Debian-based distributions:
```

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
+└ **Note** [for users migrating from the [legacy LogDNA Agent](https://github.com/logdna/logdna-agent)]:
+> If you have previously installed the `logdna-agent` Linux package and have an existing `/etc/logdna.conf` file, it is _still_ recommended to follow the instructions below to ensure that all users are retrieving the latest version of the package from the correct source repository.
 
-Specify the [ingestion key][ingestion-key] by setting the `LOGDNA_INGESTION_KEY` environment variable:
+1.  To download the agent and resources package, open a host terminal and run the appropriate commands based on the Linux distribution.
 
-```unit file (systemd)
-[Service]
-Environment="LOGDNA_INGESTION_KEY=<YOUR INGESTION KEY HERE>"
+* Debian-base distributions:
```
alfeng6, line 51:
```suggestion
* Debian-based distributions:
```

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
-The agent uses [systemd](https://systemd.io/) to run as a Linux daemon.
-
-After installing the package, you should enable it on systemd and set the ingestion key:
+2. To upgrade the agent, use the commands below, based on the distro:
 
-```shell script
-sudo systemctl daemon-reload
-sudo systemctl enable logdna-agent
-sudo systemctl edit logdna-agent
+* Debian-base distributions:
```
alfeng6, line 36:
```suggestion
* Debian-based distributions:
```

alfeng6 (changes requested):
Just a few small nits

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
 gpgkey=https://assets.logdna.com/logdna.gpg" | sudo tee /etc/yum.repos.d/logdna.repo
-sudo yum -y install logdna-agent
 ```
 
-## Usage
-
-The agent uses [systemd](https://systemd.io/) to run as a Linux daemon.
-
-After installing the package, you should enable it on systemd and set the ingestion key:
+2. To upgrade the agent, use the commands below, based on the distro:
```
jorgebay, line 34:
```suggestion
2. To install the agent, use the commands below based on the Linux distribution:
```

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
-sudo systemctl daemon-reload
-sudo systemctl enable logdna-agent
-sudo systemctl edit logdna-agent
+* **Debian-based distributions:**
+```bash
+sudo apt-get install logdna-agent
+```
+* **RPM/Red Hat-based distributions:**
+```bash
+sudo yum -y install logdna-agent
```
jorgebay, line 42: We should remove the auto yes flag here.

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
 # LogDNA Agent on Linux distributions
 
-The agent supports 64-bit (`x86_64`) Linux distributions and provides rpm and deb packages.
+The LogDNA Agent for Linux collects log data from your Linux environment. The agent is available as a Linux package for Debian-based distributions using `apt` and Red Hat-based distributions using `yum`. If you use a different package manager that does not support `.deb` or `.rpm` files, you can still use the agent by [manually compiling the binary](https://github.com/logdna/logdna-agent-v2/blob/eb06d4f3f7c1033b494f1f0439957f96533f9225/docs/README.md#building-agent-binary-on-linux).
 
-## Installing
+## Table of Contents
+  * [Installation: fresh installs](#Installation)
+  * [Upgrading: migrating from legacy Linux agent](#upgrade-migration)
```
jorgebay, line 7: Note that the bookmark is broken, it should be `#upgrademigration`

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
 # LogDNA Agent on Linux distributions
 
-The agent supports 64-bit (`x86_64`) Linux distributions and provides rpm and deb packages.
+The LogDNA Agent for Linux collects log data from your Linux environment. The agent is available as a Linux package for Debian-based distributions using `apt` and Red Hat-based distributions using `yum`. If you use a different package manager that does not support `.deb` or `.rpm` files, you can still use the agent by [manually compiling the binary](https://github.com/logdna/logdna-agent-v2/blob/eb06d4f3f7c1033b494f1f0439957f96533f9225/docs/README.md#building-agent-binary-on-linux).
 
-## Installing
+## Table of Contents
```
jorgebay, line 5: To simplify reading for new users, I think the order should be:

1. Installing
2. Usage
3. Upgrading from legacy Linux Agent

That way, new users can read the required steps in order (1 & 2) without jumping through the page.

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
 # LogDNA Agent on Linux distributions
 
-The agent supports 64-bit (`x86_64`) Linux distributions and provides rpm and deb packages.
+The LogDNA Agent for Linux collects log data from your Linux environment. The agent is available as a Linux package for Debian-based distributions using `apt` and Red Hat-based distributions using `yum`. If you use a different package manager that does not support `.deb` or `.rpm` files, you can still use the agent by [manually compiling the binary](https://github.com/logdna/logdna-agent-v2/blob/eb06d4f3f7c1033b494f1f0439957f96533f9225/docs/README.md#building-agent-binary-on-linux).
```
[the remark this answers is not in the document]
jorgebay: NIT: I think Red Hat-based distribution should be replaced with RPM-based distributions, as a way to note "Linux distributions that use the RPM Package Manager".

https://en.wikipedia.org/wiki/Category:RPM-based_Linux_distributions

Using Red Hat distros in this context leaves SUSE family out :)

jorgebay (changes requested):
I've added some suggestions below, sorry for the noise  :)

docs/LINUX.md
tanberry: @jorgebay I do see your point, but after a ton of conversation we decided to separate the install from upgrade (as you originally had it); then I decided to move the usage section come last since that was accurate for all audiences.

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
+```
+* **RPM/Red Hat-based distributions:**
+```bash
+sudo yum -y install logdna-agent
 ```
 
-The last command above, `systemctl edit logdna-agent`, will start your default text editor with an empty systemd
-configuration file for the LogDNA Agent.
+## Upgrade/migration
+└ **Note** [for users migrating from the [legacy LogDNA Agent](https://github.com/logdna/logdna-agent)]:
```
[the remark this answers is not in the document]
tanberry: The format you suggest is in our other docs, so yes, let's stick with it.

docs/LINUX.md
```diff
@@ -1,125 +1,141 @@
...
 
-The legacy LogDNA Agent for Linux used initd to start as a daemon and `.conf` files to define the settings.
 
-To upgrade, make sure you use the new repository on `https://assets.logdna.com` and run the appropriate upgrade command
-for your specific package manager.
+### _Configure the agent_
 
-### On Debian-based distributions
+└ Note \[for users upgrading /migrating from the [legacy LogDNA
+Agent](https://github.com/logdna/logdna-agent)\]: Many users will likely already have a configuration file `/etc/logdna.conf` from prior installations. The LogDNA Agent 3.3 does not utilize the legacy `/etc/logdna.conf` file, but instead uses a **systemd** unit file `/etc/logdna.env`.
```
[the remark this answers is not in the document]
tanberry: Ah yes, I shouldn't have said it isn't used, for sure.

However, I *think*we came to the decision that everyone, even people upgrading, should still create the `logdna.env` file, for consistency's sake. So the following steps apply to everyone.  But we can certainly change it again if we want. ;)

docs/LINUX.md
jorgebay: ok, sounds good.

jorgebay (approved):

PR #180, opened 2021-08-02
Get arm64 musl build working

c-nixon:
This gets Aarch64 builds working.

Unfortunately I had to drop PCRE2 from config and switch kube-runtime over to rustls to get around compiler errors, so this is a bigger change than I'd hoped. Functionality changes should be minimal to zero though.


jorgebay (commented):
Looks good, the dependency changes are pretty straightforward, the problem is that it changes behaviour: `regex` crate doesn't support look-ahead and look-behind (test failures are related to regex support change).

Maybe we could switch to something like https://github.com/fancy-regex/fancy-regex ?

c-nixon:
> Maybe we could switch to something like https://github.com/fancy-regex/fancy-regex ?

Good shout, will give it a go this morning

c-nixon:
Didn't work unfortunately, had to fork pcre2 to fix up the build

jorgebay (approved):
Nice! :shipit:

PR #186, opened 2021-08-06
Fix retry test flakiness

c-nixon:
Adding an additional sleep seems to stabilise the test.

### DIFF
--- a/bin/tests/retries.rs
+++ b/bin/tests/retries.rs
@@ -66,6 +66,9 @@ async fn test_retry_after_timeout() {
         ))
         .await;
 
+        // Sleep for long enough for retry to tick
+        tokio::time::sleep(tokio::time::Duration::from_millis(5000)).await;
+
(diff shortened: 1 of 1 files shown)


bin/tests/retries.rs
```diff
@@ -66,14 +66,17 @@ async fn test_retry_after_timeout() {
...
+
         let map = received.lock().await;
         let file_info = map.get(file_path.to_str().unwrap()).unwrap();
         // Received it multiple times
         assert!(file_info.values.len() >= file_lines.len());
 
         let attempts_made = attempts_counter.lock().unwrap();
         // It retried multiple times
-        assert_eq!(*attempts_made, attempts);
+        assert_eq!(*attempts_made, attempts + 1);
```
c-nixon, line 79: It tries once, then retries 10 times

jorgebay (approved):

PR #187, opened 2021-08-06
Log 10521 lookback defaults

c-nixon:
This changes the default lookback from smallfiles to none and updates the k8s yaml so that the smallfiles behaviour is preserved for users updating using our yamls


bin/src/main.rs
```diff
@@ -71,25 +70,30 @@ async fn main() {
...
-                Ok(agent_state) => {
-                    let _offset_state = agent_state.get_offset_state();
-                    let offsets = _offset_state.offsets();
-                    _agent_state = Some(agent_state);
-                    offset_state = Some(_offset_state);
-                    match offsets {
-                        Ok(os) => {
-                            initial_offsets =
-                                Some(os.into_iter().map(|fo| (fo.key, fo.offset)).collect());
+    match config.log.db_path {
```
c-nixon, line 73: Would it be cleared to change this to an if let given that one of the match branches is {} ?

jorgebay: +1

jorgebay (approved):
I think its good to go, we should make sure it's documented & communicated correctly :)

PR #190, opened 2021-08-09
Update client to include Read impl fix

c-nixon:
This updates the client library to a version including the Read fix and also changes the mock ingester to panic when it receives invalid payloads so that our integration tests can catch bad http requests


jorgebay (approved):
👍

PR #196, opened 2021-08-10
tweak flakey retry test again

c-nixon:
There's still some timing flakiness with the retry tests, sometimes it's 10 attempts within the window, sometimes it's 11. This checks that a sufficient number of retries occur.


jorgebay (approved):
Make sense, we shouldn't be strict about timing in tests :)

PR #197, opened 2021-08-10
Docs delete section

tanberry:
Edits per Chris' suggestions for removing the "Enable the agent"  section, and adding the `enable` step to be the last part of "Starting the agent" section.


alfeng6 (approved):
These changes look good to me

PR #198, opened 2021-08-11
Docs fix toc

tanberry:
adjust the ToC to match new arrangement of steps.


alfeng6 (approved):
looks great

docs/LINUX.md
```diff
@@ -4,10 +4,9 @@ The LogDNA Agent for Linux collects log data from your Linux environment. The ag
 
 ## Table of Contents
   * [Considerations](#considerations)
-  * [Installation: fresh installs](#Installation)
```
jorgebay, line 7: The anchor links are incorrect, it should be:
- Instead of `#Installation` should be `#installation-first-time-installations` 
- Instead of `#upgrademigration`, it should be `#upgrading-migrating-from-legacy-linux-agent`

jorgebay (commented):
I've added a note below about the anchor links.

Also, `docs/.DS_Store` file (hidden macOS file) was included, it should be removed. Our bad for not having it in the ignore list :)

@c-nixon should we add it?

c-nixon:
Yeh, the .DS_Store file should already have been in our .gitignore, I've added it, and cleaned up the one that snuck in on this branch. It will disappear in the squash.

@tanberry you just need to do a `git pull` before addressing @jorgebay's comment about the anchor links

docs/LINUX.md
tanberry: Thank you, @jorgebay !

PR #201, opened 2021-08-18
Doc final 33 beta edits

tanberry:


tanberry:
@c-nixon and @alfeng6 this is the last of the edits from Support. After this is approved and merged docs are done for 3.3 Beta.

docs/LINUX.md
```diff
@@ -91,9 +91,11 @@ The agent uses [**systemd**](https://systemd.io/) to run as a Linux daemon. The
 ### _Configure the agent_
 
 ---
-**NOTE** for users upgrading /migrating from the [legacy LogDNA
+**NOTES**
+* For users upgrading /migrating from the [legacy LogDNA
 Agent](https://github.com/logdna/logdna-agent)\: You might already have a configuration file `/etc/logdna.conf` from prior installations. The LogDNA Agent 3.3 does support the legacy `/etc/logdna.conf` file by default, and additionally uses a **systemd** unit file `/etc/logdna.env`.
 
+* For users on older Linux distributions (Centos 7, Amazon Linux 2, RHEL 7): If the default variables directory `var/lib/logdna` does not exist after following the above installation or upgrade steps, you will need to manually create the directory.
```
c-nixon, line 98: I think this is grand, if we wanted to add more explanation we could have something like

> For users on older Linux distributions (Centos 7, Amazon Linux 2, RHEL 7): The user should manually ensure that the /var/lib/logdna directory exists as the older versions of systemd packaged with these systems will not automatically create it.

But I think the current version is good

c-nixon (approved):
LGTM!

docs/LINUX.md
teeohhem: I love the way you put it Chris. It gives technical insight, but is also very informative.

docs/LINUX.md
```diff
@@ -91,9 +91,11 @@ The agent uses [**systemd**](https://systemd.io/) to run as a Linux daemon. The
 ### _Configure the agent_
 
 ---
-**NOTE** for users upgrading /migrating from the [legacy LogDNA
+**NOTES**
+* For users upgrading /migrating from the [legacy LogDNA
```
teeohhem, line 95: nit: remove the space before the slash or just say "or"

tanberry: Thanks for catching that extra space!

PR #203, opened 2021-08-23
Build static binaries with 64k page size

c-nixon:
RHEL 8 on ARM configures the kernel with a 65536 byte page size rather than the more usual 4k page size. By default Jemalloc picks up the page size on the build system and hard codes that into the binary.

Previously Jemalloc required it's compiled in page size to be the same as the system's in order to maintain alignment, however as of a year or so ago it now supports using a larger than native page size, at the cost of a small amount of performance.

This explicitly sets the jemalloc page size to 64k, I've tested the generated binary on both RHEL and Debian and it runs correctly on both.


jorgebay (approved):
Good findings!

PR #209, opened 2021-08-24
3.1.1

c-nixon:
Bump dependencies to clear cargo audit reports.

Explicitly allow RUSTSEC-2021-0078 and RUSTSEC-2021-0079 as they do not effect our configuration.

RUSTSEC-2021-0078 is non-applicable as we do not compile hyper's server component.

RUSTSEC-2021-0079 is non-applicable as we do not support explicit http proxying nor do we use application level proxying in front of any of our APIs

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -10,7 +10,7 @@
 
 [package]
 name = "logdna-agent"
-version = "3.1.0"
+version = "3.1.1"
 authors = ["CJP10 <connor.peticca@logdna.com>"]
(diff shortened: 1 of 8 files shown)


bin/Cargo.toml
```diff
@@ -10,7 +10,7 @@
 
 [package]
 name = "logdna-agent"
-version = "3.1.0"
+version = "3.1.1"
```
jorgebay, line 13: No predicates bump to 2?

PR #212, opened 2021-08-25
Tweak buffer sizes and timeouts and raise some logging thresholds

c-nixon:
Bump Timeout to 25 seconds, default gzip ratio to "best" (9), and buffer size to 2.5MB which seemed to be about the right size.

The numbers above were obtained from testing against logs.logdna.com while the environment was under peak load.

### DIFF
--- a/common/config/src/raw.rs
+++ b/common/config/src/raw.rs
@@ -98,15 +98,15 @@ impl Default for HttpConfig {
             host: Some("logs.logdna.com".to_string()),
             endpoint: Some("/logs/agent".to_string()),
             use_ssl: Some(true),
-            timeout: Some(10_000),
+            timeout: Some(25_000),
             use_compression: Some(true),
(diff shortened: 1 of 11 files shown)


common/config/src/raw.rs
```diff
@@ -98,15 +98,15 @@ impl Default for HttpConfig {
...
             use_compression: Some(true),
-            gzip_level: Some(2),
+            gzip_level: Some(9),
             ingestion_key: None,
             params: Params::builder()
                 .hostname(get_hostname().unwrap_or_default())
                 .build()
                 .ok(),
-            body_size: Some(2 * 1024 * 1024),
+            body_size: Some(2 * 1024 * 1024 + 5 * 1024),
```
jorgebay, line 109: Probably unintended

c-nixon: Yup!

PR #215, opened 2021-08-26
Add timeout to Jenkinsfile

charlesthomas:
* we've seen jobs hanging on Jenkins, so add a 1 hour timeout to prevent it
  from continuing.
* looking at the recent jobs that passed, they took between 35 and 50 minutes,
  so 1 hour should be enough, but we can bump this up if we see too many jobs
  timeout


jorgebay (approved):
Makes sense, thanks @charlesthomas !

PR #217, opened 2021-09-07
Log 10699 request prebuffering

c-nixon:
This extracts the buffering and offset tracking logic out into a stream combinator and refactors the existing http client module to be much less stateful, buffer creation, and offsets for buffers and timed batching are now all managed by the batching combinator.

The client can theoretically now do concurrent/parallel uploads however this would corrupt the offset state with the current impl as slow uploads could result in stale state data.

The retry integration tests fail as retries are locked into step with new lines in this PR. This will be corrected in a future PR


common/http/src/batch.rs
```diff
@@ -0,0 +1,838 @@
...
+
+impl<'a, St: Stream> TimedRequestBatcher<'a, St>
+where
+    St::Item: IngestLineSerialize<String, bytes::Bytes, HashMap<String, String>, Ok = ()>
+        + GetOffset
+        + std::marker::Send
+        + std::marker::Sync
+        + 'a,
+{
+    pub fn new(stream: St, capacity: usize, duration: Duration) -> TimedRequestBatcher<'a, St> {
```
c-nixon, line 95: I'm reasonably certain this can be refactored into a much more concise "unfold"

common/http/Cargo.toml
```diff
@@ -9,23 +9,31 @@ edition = "2018"
 metrics = { package = "metrics", path = "../metrics" }
 state = { package = "state", path = "../state" }
 #http
-logdna-client = { git = "https://github.com/logdna/logdna-rust.git", branch="0.5.x", version = "0.5" }
+logdna-client = { git = "https://github.com/logdna/logdna-rust.git", branch="len", version = "0.5" }
```
jorgebay, line 12: Friendly reminder to avoid forgetting to change it once merged :)

common/http/src/batch.rs
```diff
@@ -0,0 +1,630 @@
...
+
+            let mut offsets: OffsetMap = OffsetMap::default();
+
+            // Set the initial timeout
+            let mut timeout = Delay::new(state.duration);
+
+            loop {
+                let item = select(state.stream.next(), &mut timeout).await;
+
+                match item {
```
jorgebay, line 149: NIT: Isn't `tokio::select!()` macro better suited for this?

for example:

```rust
loop {
    let mut delay = time::delay_for(DURATION_SAMPLE);

    loop {
        tokio::select! { /*delay or task*/}
    }
}
```

common/http/src/batch.rs
```diff
@@ -0,0 +1,630 @@
...
+                        timeout.reset(state.duration);
+                    }
+                }
+            }
+        },
+    )))
+}
+
+#[cfg(test)]
+mod tests {
```
jorgebay, line 208: Good coverage!

common/http/src/client.rs
```diff
@@ -45,145 +30,51 @@ impl Client {
...
-        };
-        let buffer = self.buffer.replace(buffer).unwrap();
-        let buffer_size = self.buffer_bytes as u64;
-        self.buffer_bytes = 0;
-        Metrics::http().add_request_size(buffer_size);
-        let body = buffer.end().expect("Failed to close ingest buffer");
-        self.make_request(IngestBodyBuffer::from_buffer(body)).await;
-        self.last_flush = Instant::now();
+        }
+        self.make_request(body, file_offsets.as_deref()).await
```
jorgebay, line 74: Shouldn't we mark the offsets after the request?

common/http/src/batch.rs
```diff
@@ -0,0 +1,630 @@
...
+            .map(OffsetLine::new)
+            .collect();
+
+        assert_eq!(expected, lines1);
+    }
+
+    proptest! {
+        #[test]
+        fn roundtrip(
+            inp in (0..512usize)
```
jorgebay, line 602: This property test takes around 3 mins on my local machine, can we use a smaller set?

common/http/src/limit.rs
```diff
@@ -199,7 +199,7 @@ mod tests {
             let tmp = limiter.clone();
             let join_handle = spawn(move || {
                 let mut rng = rand::thread_rng();
-                for _ in 0..10000 {
+                for _ in 0..1000 {
```
jorgebay, line 202: +1

jorgebay (commented):
It's looking great, I like the new batcher logic is unified in one place.

It would be nice if we could it in dev/staging envs asap and compare metrics (data+memory usage).

I've added a few questions below.

common/http/Cargo.toml
c-nixon: Yeh, need to PR that before we can merge these. Thanks!

common/http/src/client.rs
c-nixon: This queues up the new offsets, but it doesn't actually get flushed until later in the file after we get a Response::Sent

jorgebay: oh! true!

common/http/src/batch.rs
c-nixon: We can yeh, tokio is already a dependency in the http module so we don't need to stick to futures here

common/http/src/batch.rs
c-nixon: Yeh, I've been running it in release mode, 3 mins is excessive. I'll change the number of iterations

bin/src/main.rs
```diff
@@ -102,11 +104,9 @@ async fn main() {
         config.http.template,
         handles,
         config.http.retry_base_delay,
-        config.http.retry_step_delay,
+        // TODO: use config.http.retry_step_delay,
```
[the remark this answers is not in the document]
c-nixon: I wanted to discuss whether we want to keep it in or not. It's function was to artificially delay the retry stream, but we have exponential backoff... so is it necessary?

bin/src/main.rs
```diff
@@ -272,34 +272,46 @@ async fn main() {
...
+                if executor.process(&mut line).is_some() {
+                    Some(StrictOrLazyLines::Lazy(line))
+                } else {
+                    None
                 }
-            },
-            Either::Right(_) => client.borrow_mut().poll().await,
+            }
+        },
+        Either::Right(_) => None,
```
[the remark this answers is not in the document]
c-nixon: Good catch, I've stripped it out

bin/src/main.rs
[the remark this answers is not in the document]
jorgebay: Good question... I don't think we should keep the added retry delay.

Still, we should make sure that after a long outage, retry files don't take the whole traffic bandwidth and leave some for new data. I haven't thought how we can do that, though.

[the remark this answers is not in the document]
c-nixon: Since we select over the streams at the very top level the it _should_ try and pull one then the other, so new lines should get through. We'd need to test it though, I'll add a ticket

jorgebay (approved):
Looks great 👍

PR #218, opened 2021-09-09
Log 10787 separate out retry driver

c-nixon:


common/http/src/retry.rs
```diff
@@ -44,88 +55,169 @@ struct DiskRead {
...
+                if state.waiting.is_empty() {
+                    if let Err(e) = state.fill_waiting().await {
+                        return Some((Err(e), state));
+                    };
+                    // If there are still no objects then sleep for a while
+                    if state.waiting.is_empty() {
+                        Delay::new(Duration::from_secs(
+                            state.retry_base_delay_secs.try_into().unwrap_or(2),
+                        ))
+                        .await;
```
jorgebay, line 125: should we `continue` here?

common/http/src/retry.rs
```diff
@@ -44,88 +55,169 @@ struct DiskRead {
...
+                    }
+                }
+
+                if let Some(path) = state.waiting.pop() {
+                    match Retry::read_from_disk(&path).await {
+                        Ok((offsets, ingest_body)) => {
+                            match IntoIngestBodyBuffer::into(ingest_body).await {
+                                Ok(body_buffer) => {
+                                    return Some((Ok((body_buffer, offsets)), state))
```
jorgebay, line 134: Should we throttle, e.g. in the order of 50-100ms, after a successful read to prioritize new logs?

jorgebay (commented):
New retry logic looks great, having them in a separate stream is a good call.

common/http/src/retry.rs
c-nixon: We could reintroduce the retry step delay config?

I accidentally pushed to the wrong branch so I've reopened the PR over at https://github.com/logdna/logdna-agent-v2/pull/222

PR #222, opened 2021-09-20
Log 10787 separate out retry driver

c-nixon:


common/http/src/retry.rs
```diff
@@ -103,95 +106,118 @@ impl Retry {
...
-                    .await;
+                    if let Err(e) = state.fill_waiting().await {
+                        return Some((Err(e), state));
+                    };
+                    // If there are still no objects then sleep for a while
+                    if state.waiting.is_empty() {
+                        Delay::new(Duration::from_secs(
+                            state.retry_base_delay_secs.try_into().unwrap_or(2),
+                        ))
+                        .await;
```
jorgebay, line 125: Should we call `continue` after the wait to start back at the loop?

common/http/src/retry.rs
```diff
@@ -103,95 +106,118 @@ impl Retry {
...
-                    Ok((offsets, ingest_body)) => {
-                        match IntoIngestBodyBuffer::into(ingest_body).await {
-                            Ok(body_buffer) => return Some((Ok((body_buffer, offsets)), state)),
-                            Err(e) => return Some((Err(e.into()), state)),
+                if let Some(path) = state.waiting.pop() {
+                    match Retry::read_from_disk(&path).await {
+                        Ok((offsets, ingest_body)) => {
+                            match IntoIngestBodyBuffer::into(ingest_body).await {
+                                Ok(body_buffer) => {
+                                    return Some((Ok((body_buffer, offsets)), state))
```
jorgebay, line 134: I think having an insignificant delay after each retry read (in the order of 20 to 100ms) will help prioritize the current logs.

c-nixon: Do you think we should reintroduce the step delay config option?

jorgebay: I think the previous delay was in the order of seconds, I'm in favor for something in the order of millis (non-configurable).
Having few tens milliseconds delay will cause the tailed logs to win over this and to send in a lower rate (trying to avoid thundering herd when the service come back online)

jorgebay (approved):

PR #224, opened 2021-09-21
Log 10787 cleanup retry files

c-nixon:
This lifts http error logic up to the top level and adds logic to clean up the retry files if the retry succeeds


jorgebay (approved):
lgtm!

PR #226, opened 2021-09-23
fix: create state dir if possible

dhable:
Attempt to create the configured state directory (db_path)  if it doesn't exist, handling invalid permissions by logging and skipping the creation. 

Ref: LOG-10638

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -26,7 +26,7 @@ use middleware::line_rules::LineRules;
 use middleware::Executor;
 
 use pin_utils::pin_mut;
-use state::{AgentState, StateError};
+use state::AgentState;
 use std::cell::RefCell;
 use std::path::PathBuf;
 use std::rc::Rc;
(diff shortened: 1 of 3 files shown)


bin/tests/http.rs
```diff
@@ -53,7 +53,6 @@ async fn test_http_buffer_size() {
         // Ensure that the agent sent each log message individually because each log line
         // filled the configured buffer capacity.
         let calls_made = call_counter.lock().unwrap();
-        println!("total call count = {}", calls_made);
```
dhable, line 56: I noticed that I forgot a println! in my prior PR and decided to clean it up.

bin/src/main.rs
```diff
@@ -396,6 +393,24 @@ async fn main() {
     }
 }
 
+fn load_agent_state(db_path: &DbPath) -> Result<Option<AgentState>, StateError> {
```
c-nixon, line 396: Since it's returning AgentState/StateError it feels like this could be over in the state module, perhaps as a constructor on the AgentState type?

It's just a helper function though so I don't have a strong opinon.

Any thoughts?

dhable: Yup, I was so focused on breaking out the functionality and testing it I just left it in the main file. Commit https://github.com/logdna/logdna-agent-v2/pull/226/commits/aa5313aa74261d210d5f24d1ff37dc19b0469a47 moves things around a bit.

common/state/src/lib.rs
```diff
@@ -34,60 +34,73 @@ pub struct AgentState {
     offset_cf_opt: Options,
 }
 
+// This was the old new fn implementation. It could probably be broken up
+// into a few smaller, private functions.
+fn _construct_agent_state(path: &Path) -> Result<AgentState, StateError> {
```
dhable, line 39: This was the original implementation of `AgentState::new` and was only being called by the main function. It seemed to make sense to move it into a module private function so we might be able to tackle breaking into some smaller bits at some point in the future. I was trying to avoid heaping more code into an already giant function.

c-nixon (approved):
LGTM!

PR #227, opened 2021-10-04
Log-11033 Ensure skip ahead doesn't skip too far ahead

c-nixon:


jorgebay:
The patch looks good to me, I have some questions:
- Before this patch: In case there was a later write event, after 250ms, he data the event was not skipped an. In short, the data was eventually going to get through when there were further writes, right?
- After this patch, what's the expected behaviour for large batched writes followed by a delete event? Is there any mechanism to try to get all through the wire for deletes?

c-nixon:
> Before this patch: In case there was a later write event, after 250ms, the data the event was not skipped an. In short, the data was eventually going to get through when there were further writes, right?

The timing logic is the same, the "previous event time" is recorded _after_ the tail call stream yields a result, so the "previous event time" was the time at which the tailer stopped reading last. In an ideal world we could just skip any event that had a time less than previous_event_time however to prevent any issues that could be caused by processing delays inside the tailer between it issuing it's last read and when it actually returns I added the 250ms delay.

Thinking about it, it would be more precise to have the tailer track the last timestamp and return that from the tail call. However looking at the code I don't see anything that would cause a delay, so it might not be worth polluting the tailer.

> After this patch, what's the expected behaviour for large batched writes followed by a delete event? Is there any mechanism to try to get all through the wire for deletes?

Large batched writes followed by a delete should still have all the writes processed, the idea with this skipping mechanism is that a write is only skipped if a previous write was actually processed after the one being skipped was generated.

c-nixon:
> After this patch, what's the expected behaviour for large batched writes followed by a delete event? Is there any mechanism to try to get all through the wire for deletes?

Ahh, is the question here what happens if there is a write large enough to fill the request buffer and cause the tailer to return early?

Yeh, that could be a problem, I need to think about that one.

jorgebay:
re: deletes

> write large enough to fill the request buffer

yes!

c-nixon:
@jorgebay How's that latest change look to you?

common/fs/src/cache/mod.rs
```diff
@@ -203,10 +215,26 @@ impl FileSystem {
...
+            _fs.resume_events_recv
+                .clone()
+                .map({
+                    let fs = fs.clone();
+                    move |(inode, event_time)| {
+                        fs.try_lock()
+                            .expect("couldn't lock filesystem cache")
+                            .wd_by_inode
+                            .get(&inode)
+                            .map(|wd| (Ok(WatchEvent::Modify { wd: wd.clone() }), event_time))
```
jorgebay, line 229: This is a cool idea.

What would happen for deletes with lots of writes?

I imagine that the event order would be the following:

- Modify
  - Produces a Resume Modify event to the tail of the queue
- Modify (ignored)
- Modify (ignored)
- Delete
- Resume Modify event

Issue #240, opened 2021-10-21
Prometheus metrics are millis based where Prometheus best practices dictate seconds


PeterGrace:
The Prometheus documentation states that the base unit of time [should be the second.](https://prometheus.io/docs/practices/naming/#base-units)

The current histogram for ingestion timing is implemented as millis:
```
 static ref INGEST_REQUEST_DURATION: HistogramVec = register_histogram_vec!(
        "logdna_agent_ingest_request_duration_millis",
        "Latency of the requests made to http ingestion service",
        &["outcome"],
        // Buckets ranging from 0.1ms to 26s
        exponential_buckets(0.1, 4.0, 10).unwrap()
    )
    .unwrap();
 <---snip--->
 fn elapsed(start: Instant) -> f64 {
    start.elapsed().as_micros() as f64 / 1_000.0
}
```

Due to the fact that observe() is receiving an f64 anyway, there's plenty of resolution to keep these measurements seconds-based as opposed to milliseconds-based.  `as_secs_f64` natively will return the float64 value without requiring a division step.

Would the LogDNA team be tolerant of a PR to switch this metric to seconds, or a PR that creates a second metric that implements `_seconds` to be compatible with Prometheus best practice?

PR #241, opened 2021-10-21
add seconds metric to compliment millis metric for ingest histogram

PeterGrace:
This PR is an initial attempt to implement issue #240 in the most backwards-compatible way, however I'd be just as happy to remove _millis from the existing metric and make it seconds-based.

### DIFF
--- a/common/metrics/src/lib.rs
+++ b/common/metrics/src/lib.rs
@@ -44,6 +44,13 @@ lazy_static! {
         exponential_buckets(500.0, 2.0, 13).unwrap()
     )
     .unwrap();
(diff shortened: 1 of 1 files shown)


SuperQ:
I would recommend dropping the millis version. It's pretty easy to do backwards-compatible queries in PromQL for users doing transitioning to the new metric. For example:

```
rate(logdna_agent_ingest_request_duration_seconds_bucket[1m]) or rate(logdna_agent_ingest_request_duration_millis_bucket[1m])
```

c-nixon:
@logdnabot test again please

c-nixon (approved):
This all looks good to me, Thanks very much, for the contribution!

PR #243, opened 2021-11-02
feat: Restart file tailer on inotify overflows

dhable:
Refactoring and a stream wrapper designed to trap inotify queue overflow errors and restart the file tailer instead of panicing. This should be safe now that a tailer maintains state of the last batch of logs sent to the API and thus we shouldn't miss events. 

Ref: LOG-10322


bin/src/main.rs
```diff
@@ -177,29 +169,48 @@ async fn main() {
...
+        |item| match item {
+            Err(fs::cache::Error::WatchOverflow) => {
+                warn!("overflowed kernel queue, restarting stream");
+                return true;
             }
-        });
+            _ => false,
+        },
+        |params| {
+            let watched_dirs = params.0.clone();
```
dhable, line 189: This is the core of the change. It creates new copies of the params that are then owned by the tailer. The `process` method now actually consumes the tailer but generates a stream of events.

bin/src/main.rs
```diff
@@ -177,29 +169,48 @@ async fn main() {
...
+    let ds_source_params = (
+        config.log.dirs.clone(),
+        config.log.rules.clone(),
+        config.log.lookback.clone(),
+        initial_offsets.clone(),
+    );
+
+    let fs_source = tail::RestartingTailer::new(
+        ds_source_params,
+        |item| match item {
```
dhable, line 181: This predicate indicates the overflow error should restart the tailer.

common/fs/src/cache/mod.rs
```diff
@@ -260,14 +260,12 @@ impl FileSystem {
 
     pub fn stream_events(
         fs: Arc<Mutex<FileSystem>>,
-        buf: &mut [u8],
-    ) -> Result<impl Stream<Item = (Result<Event, Error>, EventTimestamp)> + '_, std::io::Error>
-    {
+    ) -> Result<impl Stream<Item = (Result<Event, Error>, EventTimestamp)>, std::io::Error> {
```
dhable, line 263: Dropping the mutable buffer threaded through this code is required for the async blocks.

common/fs/src/cache/watch.rs
```diff
@@ -75,29 +75,27 @@ impl Watcher {
...
         Ok(WatchEventStream {
             event_stream: self.inotify.event_stream(buffer)?,
         })
     }
 }
 
-pub struct WatchEventStream<'a> {
-    event_stream: inotify::EventStream<&'a mut [u8]>,
+pub struct WatchEventStream {
+    event_stream: inotify::EventStream<Box<[u8]>>,
```
dhable, line 87: Instead of threading this buffer through all the code, it's owned here.

common/fs/src/rule.rs
```diff
@@ -1,20 +1,47 @@
...
-
 /// A trait for implementing a rule, see GlobRule/RegexRule for an example
-pub trait Rule: Debug {
+pub trait Rule {
     /// Takes a value and returns true or false based on if it matches
     fn matches(&self, value: &Path) -> bool;
 }
 
+#[derive(Debug, Clone)]
+pub enum RuleDef {
```
dhable, line 15: To make it easier to clone the values, I turned them into an enum from a trait object.

common/fs/src/tail.rs
```diff
@@ -79,330 +83,396 @@ pub struct Tailer {
...
-        rules: Rules,
-        lookback_config: Lookback,
-        initial_offsets: Option<HashMap<FileId, u64>>,
-    ) -> Self {
-        Self {
-            lookback_config,
-            fs_cache: Arc::new(Mutex::new(FileSystem::new(watched_dirs, rules))),
-            initial_offsets,
-            event_times: Arc::new(Mutex::new(HashMap::new())),
+fn get_file_for_path(fs: &FileSystem, next_path: &std::path::Path) -> Option<EntryKey> {
```
dhable, line 86: All of these functions were moved from being associated with the `Tailer` struct and are now just functions.

common/fs/src/tail.rs
```diff
@@ -79,330 +83,396 @@ pub struct Tailer {
...
-            .flatten())
+            }
+        })
+        .filter_map(|x| async move { x })
+        .flatten())
+}
+
+impl Tailer {
+    /// Creates new instance of Tailer
+    pub fn new(
```
dhable, line 380: The `Tailer` is not just a state struct.

common/fs/src/tail.rs
```diff
@@ -79,330 +83,396 @@ pub struct Tailer {
...
+            fs_cache: Arc::new(Mutex::new(FileSystem::new(watched_dirs, rules))),
+            initial_offsets,
+            event_times: Arc::new(Mutex::new(HashMap::new())),
+        }
+    }
+}
+
+pin_project! {
+    #[must_use = "streams do nothing unless polled"]
+    pub struct RestartingTailer<P, C, F, S: Stream, Fut> {
```
dhable, line 397: Beginning of the restart stream copy that's specific to passing state in to the closure.

common/fs/src/cache/watch.rs
c-nixon: I'm not sure why this was ever threaded up, probably a misreading of the type of the inotify.event_stream(buf: T) type signature as requiring a &[u8] rather than as something that can be AsRef'd as [u8]

common/fs/src/rule.rs
c-nixon: The branch to pick an implementation is also probably cheaper than chasing the pointer on every element when iterating the rules. 👍

common/fs/src/tail.rs
c-nixon: If anything they feel like methods on the FileSystem type, but that thing's busy enough as is...

common/fs/src/tail.rs
```diff
@@ -79,330 +83,396 @@ pub struct Tailer {
...
-                                if let Some((_, Some((prev_event_idx, previous_event_time)))) =
-                                    key_and_previous_event_time
+        }
+    };
+
+    debug!("Tailer starting with lookback: {:?}", state.lookback_config);
+
+    Ok(events
+        .enumerate()
+        .then({
```
c-nixon, line 312: I'm pretty sure if this .then was a .unfold we wouldn't need to Arc<Mutex<T>> the fs and then clone/capture it anymore, we could just pass it along as state... We might even be able to get rid of the RefCells around the data...

It would make the whole thing a lot less painful. I'll put in a ticket to investigate it

common/fs/src/tail.rs
c-nixon: not or now?

common/http/src/batch.rs
```diff
@@ -263,6 +263,7 @@ mod tests {
         assert_eq!(input, lines);
     }
 
+    #[ignore]
     #[tokio::test(flavor = "multi_thread", worker_threads = 3)]
```
c-nixon, line 267: Bumping the duration of this to something like 5000 makes it a lot less fragile.

c-nixon (approved):
Looks great to me!

The lifetimes really were a nightmare to work out with the old threaded &mut. I couldn't work out any way to get it back from the inotify object to pass back up and reuse it, nor could I work out a way to dynamically check it since the inotify method takes the AsRef<[u8]> + AsMut<[u8]>. The only thing I could think of needed unsafe to have multiple references to a global &'static [u8] or a memory leak (Box::leak would have done it).

common/fs/src/tail.rs
dhable: I meant to say now. Just too excited that it finally compiles.

PR #244, opened 2021-11-08
Implement concurrent upload

c-nixon:


bin/tests/cli.rs
```diff
@@ -1079,8 +1079,9 @@ async fn test_lookback_restarting_agent() {
 }
 
 #[tokio::test]
-#[cfg_attr(not(feature = "integration_tests"), ignore)]
-#[cfg_attr(not(target_os = "linux"), ignore)]
+#[ignore]
+//#[cfg_attr(not(feature = "integration_tests"), ignore)]
```
dhable, line 1083: Is commenting out these conditions intentional?

dhable (approved):
Looks good. I would ask that either remove the commented out unit test conditions in `cli.rs` or uncomment them if they still apply.

bin/tests/cli.rs
c-nixon: If I leave them uncommented clippy complains as the blanket ignore means they're never hit. I'm working on a fix for the processing of symlinks on init now

PR #245, opened 2021-11-08
Log 10775 segmented initial offsets

c-nixon:
This moves the lookback logic down into the tailer and fully adds batch support to the OffsetMap

### DIFF
--- a/bin/tests/cli.rs
+++ b/bin/tests/cli.rs
@@ -7,7 +7,7 @@ use std::thread::{self, sleep};
 use std::time::Duration;
 use wait_timeout::ChildExt;
 
-use crate::common::{consume_output, AgentSettings};
+use crate::common::{consume_output, start_ingester, AgentSettings};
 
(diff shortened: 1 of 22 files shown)


bin/tests/cli.rs
```diff
@@ -1079,8 +1079,9 @@ async fn test_lookback_restarting_agent() {
 }
 
 #[tokio::test]
-#[cfg_attr(not(feature = "integration_tests"), ignore)]
-#[cfg_attr(not(target_os = "linux"), ignore)]
+#[ignore]
```
c-nixon, line 1082: Due to an ordering issue with during startup initial lines and it's interaction with lookback tracking lines aren't currently being uploaded for the symlink and the underlyng file, only one or the other.

The lines are still being uploaded, but not for all paths. This only effects the initial read from disk, not changes that happen during agent execution.

I'm going to fix it before release

dhable (approved):

PR #247, opened 2021-11-10
Log 11479 agent hang under heavy load

c-nixon:
This resolves a problem where the agent was hanging under heavy load due to the runtime threads being put to sleep when the rate limited kicked in.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -30,6 +30,7 @@ use state::{AgentState, FileId, SpanVec};
 use std::collections::HashMap;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::Arc;
 use std::time::Duration;
 use tokio::signal::*;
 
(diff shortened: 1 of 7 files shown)


bin/src/main.rs
```diff
@@ -348,40 +359,57 @@ async fn main() {
...
-                        Ok(s) => match s {
-                            SendStatus::Sent => {
-                                debug!("cleaned up retry file");
-                                if let Err(e) = std::fs::remove_file(path) {
-                                    error!("couldn't clean up retry file {}", e)
-                                }
+        .for_each_concurrent(None, move |body_offsets| {
+            let client = client.clone();
+            async {
+                tokio::spawn(async move {
```
c-nixon, line 387: spawning these futures onto the tokio runtime allows them to run in other threads.

common/http/src/limit.rs
```diff
@@ -112,43 +110,39 @@ impl<T> Drop for InnerSlot<T> {
...
-            step: Cell::new(0),
+            step: AtomicU32::new(0),
             base: 2,
             multipler: 10,
         }
     }
 
-    pub fn snooze(&self) {
-        let step = self.step.get();
-        sleep(Duration::from_millis(self.base.pow(step) * self.multipler));
```
c-nixon, line 131: Sleeping worker threads in an async runtime for ever increasing durations turns out to be a bad idea.

dhable (approved):
Nice find!

PR #249, opened 2021-11-11
feat: Adding metrics to around ingest retry

dhable:
This ticket adds metrics for the following retry data points:

* pending files in the retry queue that have not been resent yet
* counts of successful retry ingest API calls
* counts of failed retry ingest API calls
* disk space used by all the current retry files

This is in addition to counts around the HTTP API calls around ingest API attempts, successes and failures. Tested via the included integration test that makes assertions around the metric properties instead of specific metric values.

Ref: LOG-10294

### DIFF
--- a/common/http/src/retry.rs
+++ b/common/http/src/retry.rs
@@ -17,7 +17,7 @@ use metrics::Metrics;
 use serde::Deserialize;
 use thiserror::Error;
 
-use tokio::fs::{read_dir, remove_file, rename, File, OpenOptions};
+use tokio::fs::{metadata, read_dir, remove_file, rename, File, OpenOptions};
 use tokio::io::{AsyncReadExt, AsyncWriteExt, BufReader, BufWriter};
 
 use uuid::Uuid;
@@ -89,6 +89,7 @@ impl Retry {
 
     async fn fill_waiting(&self) -> Result<(), Error> {
         let mut files = read_dir(&self.directory).await?;
+        let mut retry_disk_used = 0u64;
         while let Some(file) = files.next_entry().await? {
             let path = file.path();
             if path.is_dir() {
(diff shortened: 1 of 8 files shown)


common/http/src/retry.rs
```diff
@@ -110,13 +111,28 @@ impl Retry {
...
+
                 if Utc::now().timestamp() - timestamp < self.retry_base_delay_secs {
                     continue;
                 }
+                Metrics::retry().inc_pending();
                 self.waiting.push(path);
             }
         }
 
+        Metrics::retry().report_storage_used(retry_disk_used);
```
jorgebay, line 135: Instead of updating the gauge when the file system is polled, would it be more dynamic to call `add()` and `sub()` on the gauge every time a file is added / removed from the file system?

jorgebay (commented):
Looks great!

I've added a small suggestion below.

common/http/src/retry.rs
dhable: @jorgebay I initially had coded the implementation that way but ran into two issues with that implementation:
1. There are multiple paths in our code where a retry file may be removed. When a retry block is sent again but fails, the retry code writes out a new retry file and cleans up the prior version. This was causing double counting of disk space. 
2. If the agent should crash with files in the retry directory, I wanted to ensure that the agent emits accurate disk stats on restart.

In both cases, it seemed easier to compute the full disk usage metric and allow the metric to correct itself in case we miss a retry cleanup or if the agent is trying to recover after a crash.

jorgebay: Sounds good, in any case, it's a great improvement to have the size of the queue in the metrics 👍 
We can revisit in the future in case the gauge shows stale data.

jorgebay (approved):

PR #251, opened 2021-11-15
build: Adding build-rpm target alongside the build-deb target

dhable:


Makefile
```diff
@@ -349,6 +350,36 @@ build-deb: build-release
 				"/build/target/${TARGET}/release/logdna-agent=/usr/bin/logdna-agent" \
 				"packaging/linux/logdna-agent.service=/lib/systemd/system/logdna-agent.service"'
 
+RPM_VERSION=1
+RPM_ARCH_NAME_x86_64=amd64
+RPM_ARCH_NAME_aarch64=arm64
```
c-nixon, line 355: I believe RPM uses x86_64 and aarch64 for it's packages, so we don't need to lookup here

jmoses:
built an installable rpm for me

Makefile
dhable: Good to know. Addressed in commit https://github.com/logdna/logdna-agent-v2/pull/251/commits/ceb1134e57dba99b72072e4c35c339fd78e343b8.

c-nixon (approved):

PR #255, opened 2021-11-19
build: ignore chrono CVE; awaiting long term fixes upstream

dhable:
Fix Jenkins builds to ignore the chrono CVE for now while we wait for long term fixes from upstream packages as well as fixes to our code base.


charlesthomas (approved):
approving because there's an open ticket to address this, and it's blocking a thing i'm trying to work on

[LOG-11559](https://logdna.atlassian.net/browse/LOG-11559)

PR #263, opened 2021-11-24
build: Make image pub option Jenkins param

dhable:
One of the difficulties encountered when trying to manually update and push out container images was constraint that one needed to manually confirm the publish step within the 5 minute timeout window from the Jenkins console. This PR updates the Jenkinsfile to include two boolean parameters to replace the two timeout prompts, with the default value for each prompt being `false`. The scheduled triggers set each parameter to `true` to maintain the weekly rebuild/publish schedule we had before.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,9 @@ pipeline {
     }
     triggers {
         issueCommentTrigger(TRIGGER_PATTERN)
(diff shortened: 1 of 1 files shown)


jmoses (approved):
lgtm as long as they work :D

c-nixon (approved):

PR #265, opened 2021-11-29
fix: bump tokio version

dhable:
Used `cargo update -p tokio` to bump the tokio version in the cargo.lock file to pick up the fix for RUSTSEC-2021-0124.

Ref: LOG-11559


c-nixon (approved):
LGTM

PR #271, opened 2021-12-01
Add a shutdown signal to allow agent to exit on error

c-nixon:
The primary purpose of this PR are the changes to src/bin/main.rs to add a shutdown signal that can be triggered in the case of invalid configuration.

There are additional changes to fix the flakey journalctl tests and the build on more recent linux kernels which were causing jenkins and local development problems.


dhable (approved):
LGTM!

PR #272, opened 2021-12-01
fix: bump tokio version to address CVE

dhable:
This PR bumps the tokio version to 1.14.0, which is not affected by RUSTSEC-2021-0124. Given the age of this branch, this is a little more involved than bumping tokio version numbers. The first commit shows all the dependencies that also needed to be bumped with the move from tokio 0.2 to 1. The second commit shows all of the breaking API changes that needed modifications. The third commit is a small change to get unit tests to work.

Ref: LOG-11559

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,6 @@
 version = 3
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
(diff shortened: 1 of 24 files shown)


c-nixon (approved):
Compiles, passes tests, looks good to me.

PR #274, opened 2021-12-02
Add hook to process requests to Mock ingester

c-nixon:
This adds a hook to the mock ingester so that we can process requests and then adds an example use by checking that protocol negotiation via ALPN is being observed by the agent


dhable (commented):
+1

PR #276, opened 2021-12-02
fix: bump tokio version to address CVE

dhable:
This PR bumps the tokio version to 1.14.0, which is not affected by RUSTSEC-2021-0124. Given the age of this branch, this is a little more involved than bumping tokio version numbers. The first commit shows all the dependencies that also needed to be bumped with the move from tokio 0.2 to 1. The second commit shows all of the breaking API changes that needed modifications. The third commit is a small change to get unit tests to work.

Ref: LOG-11559

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,7 @@ pipeline {
     }
     environment {
         RUST_IMAGE_REPO = 'us.gcr.io/logdna-k8s/rust'
-        RUST_IMAGE_TAG = 'buster-stable'
+        RUST_IMAGE_TAG = 'buster-1-stable'
         SCCACHE_BUCKET = 'logdna-sccache-us-west-2'
         SCCACHE_REGION = 'us-west-2'
         CARGO_INCREMENTAL = 'false'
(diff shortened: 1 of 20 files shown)


Jenkinsfile
```diff
@@ -15,7 +15,7 @@ pipeline {
     }
     environment {
         RUST_IMAGE_REPO = 'us.gcr.io/logdna-k8s/rust'
-        RUST_IMAGE_TAG = 'buster-stable'
+        RUST_IMAGE_TAG = 'buster-1-stable'
```
dhable, line 18: @c-nixon Without this, I was getting ``feature `resolver` is required`` from the Jenkins build when it was trying to download the time crate. I assume it is something to do with the version of cargo installed in the older image but not sure what the downstream impact of using the new build image would be. Is there another way to fix this?

c-nixon: I think we are ok bumping the build image version, it shouldn't have any downstream impact as it isn't shipped or anything like that, it only generates the binaries

c-nixon (approved):
This all looks good to me

PR #277, opened 2021-12-06
feat: expose retry dir as conf option

dhable:
Exposes the retry directory as a configurable element, allowing for users to change where they would like retry files stored. Defaults to the current hard coded value of `/tmp/logdna`.

Ref: LOG-10295


c-nixon (approved):
This all looks great

PR #279, opened 2021-12-16
feat: Retry disk space can be limited

dhable:
New configuration parameter now allows for limiting the amount of disk space the agent will use to store retry files. The usage is tracked across the sending and consuming retry structs and the disk usage metric now references that counter. Also updates the existing retry integration tests to separate out the retry directory used so each test now has a clean state on execution.

Ref: LOG-10296

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -99,6 +99,7 @@ async fn main() {
         config.http.retry_dir,
         config.http.retry_base_delay,
         config.http.retry_step_delay,
+        config.http.retry_disk_limit,
     );
 
     let concurrency_limit = Some(100);
@@ -390,32 +391,11 @@ async fn main() {
                             let RetryItem {
                                 body_buffer,
                                 offsets,
-                                path,
+                                path: _,
                             } = item;
                             match client.send(body_buffer, offsets).await {
                                 Ok(s) => match s {
-                                    SendStatus::Sent => {
-                                        Metrics::http().increment_retries_success();
-                                        match std::fs::remove_file(path.clone()) {
-                                            Ok(()) => debug!(
-                                                "cleaned up retry file {}",
-                                                path.to_string_lossy()
-                                            ),
-                                            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
-                                                debug!(
-                                                    "cleanup {}: not found err, skipping",
-                                                    path.to_string_lossy()
-                                                );
-                                            }
-                                            Err(e) => {
-                                                error!(
-                                                    "couldn't clean up retry file {}, reason={}",
-                                                    path.to_string_lossy(),
-                                                    e
-                                                )
-                                            }
-                                        }
-                                    }
+                                    SendStatus::Sent => Metrics::http().increment_retries_success(),
                                     _ => {
                                         Metrics::http().increment_retries_failure();
                                         handle_send_status(s)
--- a/common/config/src/argv.rs
+++ b/common/config/src/argv.rs
@@ -1,6 +1,7 @@
 use crate::raw::{Config as RawConfig, Rules};
 use fs::lookback::Lookback;
 use http::types::params::{Params, Tags};
+use humanize_rs::bytes::Bytes;
 use k8s::K8sTrackingConf;
 use std::env::var as env_var;
 use std::path::PathBuf;
@@ -210,9 +211,10 @@ pub struct ArgumentOptions {
     retry_dir: Option<String>,
 
     /// When set, limits the amount of disk space the agent will use to store log lines that
-    /// need to be resent to the ingestion API.
+    /// need to be resent to the ingestion API. Values can be defined with units of KB, MB, GB,
+    /// etc. Numbers need to be integer values.
     #[structopt(long, env = env::RETRY_DISK_LIMIT)]
-    retry_disk_limit: Option<u64>,
+    retry_disk_limit: Option<Bytes<u64>>,
 }
 
 impl ArgumentOptions {
@@ -287,8 +289,8 @@ impl ArgumentOptions {
             raw.http.retry_dir = self.retry_dir.map(PathBuf::from);
         }
 
-        if self.retry_disk_limit.is_some() {
-            raw.http.retry_disk_limit = self.retry_disk_limit;
+        if let Some(disk_limit) = self.retry_disk_limit {
+            raw.http.retry_disk_limit = Some(disk_limit.size());
         }
 
         if !self.log_dirs.is_empty() {
@@ -490,6 +492,7 @@ mod test {
     use super::*;
 
     use crate::raw::{Config as RawConfig, Rules};
+    use humanize_rs::bytes::Unit;
     use std::env::set_var;
 
     static EXCLUSION_GLOB_DEFAULT: &str = "/var/log/wtmp,/var/log/btmp,/var/log/utmp,/var/log/wtmpx,/var/log/btmpx,/var/log/utmpx,/var/log/asl/**,/var/log/sa/**,/var/log/sar*,/var/log/tallylog,/var/log/fluentd-buffers/**/*,/var/log/pods/**/*";
@@ -646,7 +649,7 @@ mod test {
             ingest_timeout: Some(1111111),
             ingest_buffer_size: Some(222222),
             retry_dir: some_string!("/tmp/argv"),
-            retry_disk_limit: Some(123456),
+            retry_disk_limit: Some(Bytes::new(123456, Unit::Byte).unwrap()),
             ..ArgumentOptions::default()
         };
         let config = argv.merge(RawConfig::default());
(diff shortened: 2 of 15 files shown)


common/config/src/argv.rs
```diff
@@ -207,6 +208,11 @@ pub struct ArgumentOptions {
     /// Defaults to /tmp/logdna.
     #[structopt(long, env = env::RETRY_DIR)]
     retry_dir: Option<String>,
+
+    /// When set, limits the amount of disk space the agent will use to store log lines that
+    /// need to be resent to the ingestion API.
+    #[structopt(long, env = env::RETRY_DISK_LIMIT)]
+    retry_disk_limit: Option<u64>,
```
teeohhem, line 215: What do you think about naming it with the unit?
`retry_disk_limit_mb` for example

bin/src/main.rs
```diff
@@ -390,32 +391,11 @@ async fn main() {
                             let RetryItem {
                                 body_buffer,
                                 offsets,
-                                path,
+                                path: _,
                             } = item;
                             match client.send(body_buffer, offsets).await {
                                 Ok(s) => match s {
-                                    SendStatus::Sent => {
```
dhable, line 397: After refactoring and tracing down some failure on the disk used metric, I concluded that the file in question has already been deleted at this point.

common/config/src/properties.rs
```diff
@@ -172,6 +173,13 @@ fn from_property_map(map: HashMap<String, String>) -> Result<Config, ConfigError
         result.http.retry_dir = Some(PathBuf::from(value));
     }
 
+    if let Some(value) = map.get(&RETRY_DISK_LIMIT) {
+        let limit = u64::from_str(value).map_err(|e| {
```
c-nixon, line 177: What about using something like https://crates.io/crates/humansize to allow users to put in friendly strings like "1GB" or "200MB" etc?

common/config/src/argv.rs
jorgebay: Should we default to a super super large value, like 2 or 4GiB.
I think having unlimited storage usage by default is unexpected and during long outages can lead to infra issues on the user side.

common/config/src/argv.rs
```diff
@@ -281,6 +287,10 @@ impl ArgumentOptions {
             raw.http.retry_dir = self.retry_dir.map(PathBuf::from);
         }
 
+        if self.retry_disk_limit.is_some() {
+            raw.http.retry_disk_limit = self.retry_disk_limit;
```
jorgebay, line 291: Does setting it to `0` mean storing retry files won't be allowed? 
If that's the case, it's a nice little feature that would be useful to document.

common/http/src/retry.rs
```diff
@@ -154,6 +155,9 @@ impl Retry {
             }
         }
 
+        self.disk_used.fetch_sub(file_size, SeqCst);
+        Metrics::retry().report_storage_used(self.disk_used.load(SeqCst));
```
jorgebay, line 159: Minor: we could use the atomic result of `fetch_sub` instead of reloading it.

jorgebay (commented):

jorgebay:
Should we document it on the README?

common/config/src/argv.rs
dhable: Correct. It acts as a disable flag.

common/config/src/properties.rs
dhable: humansize seems to only convert int values into human readable strings. https://crates.io/crates/humanize-rs looks like it's the popular choice for parsing strings back into ints. I'm going to plan on going with that .

common/http/src/retry.rs
jorgebay: Yes. Atomic ops are expensive, I was suggesting using a single one (minor improvement in perf, bad for readability) 🙂 

I don't feel strongly about, tho.

dhable: Addressed in commit https://github.com/logdna/logdna-agent-v2/pull/279/commits/9d0881cda62a8c0e2b50802810b6ada6f20ff1e6.

common/config/src/argv.rs
dhable: > Should we default to a super super large value, like 2 or 4GiB.
> I think having unlimited storage usage by default is unexpected and during long outages can lead to infra issues on the user side.

We discussed this at standup today and agree that filling the disk space up is problematic but that enforcing a default might lead to data loss if we pick incorrectly. Customers who are sensitive to their disk usage can monitor the disk used as well as set limits for their configurations.

dhable:
> Should we document it on the README?
Documentation updated in commit https://github.com/logdna/logdna-agent-v2/pull/279/commits/e403bc8c6cc5827155bfe456bce05b6ff8cad6c3.

common/config/src/properties.rs
dhable: Commit https://github.com/logdna/logdna-agent-v2/pull/279/commits/8965a89b0ecfa3ac842b3db4e1d5eb5426b0d4f6 adds humanize_rs as a parser for command line argument and environment values. Getting that same logic to work with serde and the YAML format has been more challenging and isn't yet done in the PR.

common/config/src/argv.rs
dhable: > Should we default to a super super large value, like 2 or 4GiB.
> I think having unlimited storage usage by default is unexpected and during long outages can lead to infra issues on the user side.

We discussed this at standup today and agree that filling the disk space up is problematic but that enforcing a default might lead to data loss if we pick incorrectly. Customers who are sensitive to their disk usage can monitor the disk used as well as set limits for their configurations. Tom felt that running the risk of consuming the disk is less bad than failing to ingest logs.

dhable: > What do you think about naming it with the unit?

@teeohhem Commit https://github.com/logdna/logdna-agent-v2/commit/8965a89b0ecfa3ac842b3db4e1d5eb5426b0d4f6 is adding support for using humanize. Does that address your concerns?

common/config/src/properties.rs
dhable: I updated the parsing logic for the yaml file so it also understand a humanized string in commit https://github.com/logdna/logdna-agent-v2/pull/279/commits/0535082caaa4326836d35c2a9120bd7fd1be8ee8.

jorgebay (commented):
Last changes look great! 👍 

(I'll leave the approval for the agent team)

teeohhem (approved):

common/config/src/raw.rs
```diff
@@ -1,12 +1,75 @@
...
+use humanize_rs::bytes::Bytes;
+use serde::de::{Deserializer, Error, Unexpected, Visitor};
 use serde::{Deserialize, Serialize};
+use std::fmt;
 use std::fs::File;
 use std::io::{Seek, SeekFrom};
 use std::path::{Path, PathBuf};
 use std::vec::Vec;
 
+pub fn filesize_deser<'de, D>(d: D) -> Result<Option<u64>, D::Error>
```
c-nixon, line 13: Ouch, manually implementing serde serializers/deserializers is never fun.

common/http/src/retry.rs
```diff
@@ -262,22 +284,75 @@ impl RetrySender {
...
+            self.disk_used.load(SeqCst)
+        };
+
+        Metrics::retry().report_storage_used(new_disk_used);
+
         let mut new_file_name = self.directory.clone();
         new_file_name.push(format!("{}_{}.retry", fn_ts, fn_uuid));
 
-        Ok(rename(file_name, new_file_name).await?)
+        return Ok(rename(file_name, new_file_name).await?);
```
c-nixon, line 338: Style nit, unnecessary return?

c-nixon (approved):
Couple of minor nits

common/config/src/raw.rs
dhable: I tried using https://crates.io/crates/serde-humanize-rs but it makes the assumption that you're going to represent sizes as `usize` but the fs APIs use u64 for all file sizes. It was either roll our own serde deserializer or put in a bunch of code to handle loss of precision in converting from `usize` to `u64`. This seemed less problematic.

common/http/src/retry.rs
dhable: D'OH! Old C habits die hard. Addressed in commit https://github.com/logdna/logdna-agent-v2/pull/279/commits/5795585b171c8a2ad48b1e840ff438a5b0aa72cf.

common/config/src/raw.rs
c-nixon: Huh, interesting, usize to u64 shouldn't be lossy on any modern system, can't think of anything with a 128bit pointer. We could probably handle it with TryFrom and a LossOfPrecision error variant, but this is as tidy or tidier.

PR #282, opened 2022-01-08
feat: added compression to retry files

dkhokhlov:
LOG-11062


common/http/src/retry.rs
```diff
@@ -132,9 +134,22 @@ impl Retry {
     }
 
     async fn read_from_disk(&self, path: &Path) -> Result<(Option<OffsetMap>, IngestBody), Error> {
-        let mut file = BufReader::new(File::open(path).await?);
         let mut data = String::new();
-        file.read_to_string(&mut data).await?;
+        {
```
dhable, line 138: Can we remove this block around the code? The only reason I could see for it would be ensuring that the buffers and decoder are deallocated before the remainder of the function is called and it doesn't seem to be that much need for the slightly earlier deallocation.

It just looks odd and not like idomatic rust code that I've seen.

common/http/src/retry.rs
```diff
@@ -245,44 +260,42 @@ impl RetrySender {
...
-                    // Get the std::fs::File out of the tokio BufWriter
-                    file.flush().await?;
-                    let mut std_file = std::io::BufWriter::new(file.into_inner().into_std().await);
-                    move || -> Result<std::fs::File, std::io::Error> {
-                        // Serialise the offsets to the std::io::BufWriter
-                        serde_json::to_writer(&mut std_file, &offsets)?;
-                        std_file.flush()?;
-                        Ok(std_file.into_inner()?)
+        {
+            let mut plain_file = BufWriter::new(
```
dhable, line 264: Most idomatic rust code would keep the name as `file` so it's shaddowed on line 272 since you never refer to `plain_file` anywhere else in the method. Coming from Java/Scala/etc is does seem strange but does have the benefit that it's very clear that you do not intend to ever use the plain file object anywhere else other than in the creation of the gzip file. 

No change required, just FYI.

dhable (changes requested):
There is one comment around adding some logging that I'd like to see before approving. Otherwise, looking good.

common/http/src/retry.rs
dkhokhlov: It is needed to grantee that file is closed. Otherwise size calcs and file stats later in the same function may be affected.

common/http/src/retry.rs
```diff
@@ -132,9 +134,22 @@ impl Retry {
...
+        {
+            let file = BufReader::new(File::open(path).await?);
+            // try to read as gzip
+            let mut zip = GzipDecoder::new(file);
+            if let Err(e) = zip.read_to_string(&mut data).await {
+                if e.kind() != std::io::ErrorKind::InvalidData {
+                    debug!("GzipDecoder failed: {}", e);
+                    return Err(Error::from(e));
+                }
+                // now read as text
```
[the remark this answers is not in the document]
dkhokhlov: uploaded the change.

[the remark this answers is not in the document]
dhable: I was thinking that the log would be between line 147 and 148 like
```
if let Err(e) = zip.read_to_string(&mut data).await {
  if e.kind() != std::io::ErrorKind::InvalidData {
      let result = Err(Error::from(e));
      return result;
  }

  warn!("GzipDecoder failed: {} for file [{}], trying as plain text", e, path.display());

  // now read as text  
  let mut inner_file = zip.into_inner();
  inner_file.rewind().await?;
  inner_file.read_to_string(&mut data).await?;
}
```
That way if there is then a serde error later on or some other unusual error, we would have a record that the gzip failed with kind of `InvalidData` and we were going to try to read it as plain text.

common/http/src/retry.rs
dhable: I could easily see someone removing that block in the future and commit it because all the CI tests pass. So we should either build a test that will fail when changed or comment it to communicate the intent. We've never seen a failure from any of our tests that are checking that files are cleaned up so it would be easy for someone to undo that without understanding the intent.

common/http/src/retry.rs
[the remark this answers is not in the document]
dkhokhlov: +1. uploaded.

common/http/src/retry.rs
dkhokhlov: added comments to scope blocks.

dhable (approved):

PR #283, opened 2022-01-10
build: Picking main line linting fixes over

dhable:
After updating the Jenkins build container images, we have some new linting rule failures that have been addressed in master as part of PR #279. This PR applies those changes to the 3.4 branch so we can build it again.


c-nixon (approved):
LGTM

PR #284, opened 2022-01-19
Benchmarking tweaks

c-nixon:
This tweaks the benchmark to work with the new base image and to make it more verbose to make it slightly more debuggable.

### DIFF
--- a/bench/bin/throughput.rs
+++ b/bench/bin/throughput.rs
@@ -215,9 +215,7 @@ async fn main() -> Result<(), std::io::Error> {
                 &format!("{}", agent_handle.id()),
                 "-o",
                 "/tmp/flamegraph.svg",
-            ])
-            .stdout(Stdio::null())
-            .stderr(Stdio::null());
+            ]);
 
         Some(flamegraph_cmd.spawn().unwrap())
     } else {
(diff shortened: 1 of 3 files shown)


bench/bin/throughput.rs
```diff
@@ -232,11 +230,55 @@ async fn main() -> Result<(), std::io::Error> {
 
         let line_counter = line_counter.clone();
         async move {
+            tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
c-nixon, line 233: This adds an additional writer so that the agent has to read from more than one file.

bench/bin/throughput.rs
```diff
@@ -251,7 +293,7 @@ async fn main() -> Result<(), std::io::Error> {
                     for word in words.iter().cycle().take(line_count / 20) {
                         count += 1;
                         if count % 10_000 == 0 {
-                            std::thread::sleep(std::time::Duration::from_millis(50));
+                            std::thread::sleep(std::time::Duration::from_millis(12));
```
c-nixon, line 296: This is still a terrible way to control the rate, but tokio doesn't have sub millisecond timers to allow us to easily control lines written per second.

stsantilena: Just for my understanding, it's dropping the wait time so that both writers are writing files at a faster rate?

c-nixon: Yeh. we should really refactor it so that it uses some sort of limiter so that it attempts to write a certain number of lines per millisecond, but it's more logic and the original purpose of this bench was mostly to saturate the agent for debugging purposes

PR #285, opened 2022-01-20
test: hardening up the metrics integration test

dhable:
Building off the retry metrics work, this PR updates the existing metrics test case to not just check that metrics data points are defined. but also makes assertions about the captured values to ensure they are working as we intended. The test case also expands the number of metrics looked at, bringing the coverage to 12/15 metrics. 

I did find one metric that was never emitted in the code base and thus removed its definition.  

Ref: LOG-11063

### DIFF
--- a/bin/tests/common.rs
+++ b/bin/tests/common.rs
@@ -1,6 +1,7 @@
 use core::time;
 
 use hyper::{Client, StatusCode};
+use prometheus_parse::{Sample, Scrape};
 use rand::seq::IteratorRandom;
 
 use std::fs;
@@ -11,10 +12,13 @@ use std::process::{Child, Command, Stdio};
 use std::sync::mpsc;
 use std::sync::mpsc::TryRecvError;
 use std::thread;
+use std::time::Duration;
 
 use std::net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener};
 
-use futures::Future;
+use futures::future::{AbortHandle, Abortable};
+use futures::stream::{Stream, StreamExt};
+use futures::{stream, Future};
 use log::debug;
 use logdna_mock_ingester::{
     http_ingester, http_ingester_with_processors, https_ingester_with_processors, FileLineCounter,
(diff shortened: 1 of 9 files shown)


bin/tests/common.rs
```diff
@@ -481,3 +485,83 @@ pub async fn fetch_agent_metrics(
...
+            }
+        };
+
+        state.pop().map(|metric| (metric, state))
+    })
+}
+
+// The compiler/linter believes this function isn't used anywhere but it is currently
+// used in the retries and http integration tests. This flag disables that false positive.
+#[allow(dead_code)]
```
stsantilena, line 533: Are these `dead_code` attributes used just for compiler/linter errors? Looks like it's complaining about functions and structs?

dhable: Yes, the compiler/linter think this code is not used because the only place it is used is in some test code that is conditional to the integration_test flag. It's only really a concern for common code in the `bin/tests/*` modules.

common/k8s/src/middleware/metadata.rs
```diff
@@ -205,6 +205,7 @@ impl Middleware for K8sMetadata {
         });
     }
 
+    #[allow(clippy::question_mark)]
```
dhable, line 208: Not sure why clippy is now complaining about this block of code but it occurs both locally and on Jenkins so I added the allow tag.

common/metrics/src/lib.rs
```diff
@@ -310,30 +303,33 @@ impl Http {
     }
 
     pub fn add_request_success(&self, start: Instant) {
+        let duration = start.elapsed();
```
dhable, line 306: Before, we would emit slightly different timestamp values since each `observe` call computed its own timestamp. Now we use the same duration for both metrics so they are consistent with each other.

common/middleware/src/line_rules.rs
```diff
@@ -125,6 +125,7 @@ impl LineRules {
             redacted.extend_from_slice(&value[index..]);
         }
 
+        #[allow(clippy::question_mark)]
```
dhable, line 128: Not sure why clippy is now complaining about this block of code but it occurs both locally and on Jenkins so I added the allow tag.

c-nixon (approved):
Sorry this took me so long to get through, tests are hard to review!

It all looks good to me, great to have coverage of these and nice catch on the disused Metric

PR #286, opened 2022-01-31
Win rebase

c-nixon:
Current state: test_read_file_appended_in_the_background is still flakey on Jenkins. This current wip has the rules logic stripped out as I thought that might be the cause, but it doesn't seem to be the case.

Current hypothesis is that it's debouncer related...


bin/src/main.rs
```diff
@@ -296,19 +320,14 @@ async fn _main() {
...
     let fs_source = tail::RestartingTailer::new(
         ds_source_params,
-        |item| match item {
-            Err(fs::cache::Error::WatchOverflow) => {
-                warn!("overflowed kernel queue, restarting stream");
-                true
-            }
-            _ => false,
-        },
+        // TODO check for any conditions that require the tailer to restart
```
dkhokhlov, line 323: does not look good to drop error processing.

dkhokhlov (changes requested):

bin/src/main.rs
c-nixon: Yup, that's one of the parts that's still WIP, need to propagate the rescan event from the notify stream.

bin/src/main.rs
```diff
@@ -261,7 +281,12 @@ async fn _main() {
 
     executor.init();
 
-    #[cfg(feature = "libjournald")]
+    // Use an internal env var to support running integration test w/o additional delays
+    let event_delay = std::env::var("LOGDNA_INTERNAL_FS_DELAY")
```
dkhokhlov, line 285: hardcoded string needs to be replaced by a const from config::env_vars

dkhokhlov (changes requested):

bin/src/main.rs
c-nixon: Good catch, there may be other things like this where the change predates the new convention, I missed this one

dkhokhlov (approved):

PR #288, opened 2022-02-02
feat: windows MSI installer

dkhokhlov:
- WiX project
- MSI installer
- signed EXEs & MSI with dev cert for now

LOG-10066


packaging/windows/logdna.conf
```diff
@@ -0,0 +1,23 @@
+http:
+  host: logs.logdna.com
+  endpoint: /logs/agent
+  use_ssl: true
+  timeout: 10000
+  use_compression: true
+  gzip_level: 2
+  body_size: 2097152
+  ingestion_key: <YOUR_INGESTION_KEY>
```
jiangao1, line 9: Should we add the params attribute below the ingestion_key as part of the default template config?
 For example:
      params:
        hostname: 
        now: 0

dkhokhlov: no. Chris fixed the default hostname bug already.

Makefile
```diff
@@ -475,6 +476,28 @@ publish-s3-binary:
...
+	fi;
+
+define MSI_RULE
+.PHONY: msi-$(1)
+msi-$(1):  ## create signed exe(s) and msi in $(BUILD_DIR)/signed
+	$(eval BUILD_DIR := target/$(TARGET)/$(1))
+	$(eval CERT_NAME := "logdna_dev_cert.pfx")
+	aws s3 cp "s3://ecosys-vault/$(CERT_NAME)" "$(BUILD_DIR)" && \
+	aws s3 cp "s3://ecosys-vault/$(CERT_NAME).pwd" "$(BUILD_DIR)" && \
+	$(BENCH_COMMAND) "--env BUILD_DIR=/build/$(BUILD_DIR) --env CERT_NAME=$(CERT_NAME) --env BUILD_VERSION=$(BUILD_VERSION)" "cd /build/packaging/windows && ./mk_msi" && \
```
c-nixon, line 495: Out of curiosity why BENCH_COMMAND instead of UNCACHED_RUST_COMMAND?

packaging/windows/mezmo-agent.wxs
```diff
@@ -0,0 +1,57 @@
...
+            </Directory>
+            <Directory Id="CommonAppDataFolder">
+                <Directory Id="DATAFOLDER" Name="logdna" />
+            </Directory>
+        </Directory>
+    </Fragment>
+    <Fragment>
+        <ComponentGroup Id="ProductComponents" Directory="INSTALLFOLDER">
+            <Component Id="ProductComponent" Win64="$(var.Win64)" Guid="56CD2588-B976-4198-B815-FAB7E1E57CD7">
+                <File Id="AgentSvc" Name="logdna-agent-svc.exe" DiskId="1" Source="$(env.AGENT_SVC_EXE_PATH)" KeyPath="yes" />
```
c-nixon, line 41: I think we might run into upgrade issues if we're not specifying the FileVersion, afaik cargo doesn't set windows resource info in binaries by default, so the MSI will never have any way to infer get ProductVersion unless we provide it or compile it into the agent

packaging/windows/mk
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -eu
+VER=$(sed -nE "s/^version = \"(.+)\"\$$/\1/p" ../../bin/Cargo.toml)
```
c-nixon, line 3: This means we're stuck with just the Major.Minor.Patch versions rather than full Major.Minor.Patch.Build versions, which may be insufficient for upgrade purposes if we do a rebuild to pick up something like a security patch to the compiler without bumping the patch version.

We have some logic over in the cdn build to handle this on the linux packages. https://github.com/answerbook/logdna-cdn/blob/48c23808f088c539c2d849c6ba2af6505149e0a3/deployment/docker/Dockerfile#L105

jiangao1:
Saw the don't override logdna.conf commit. Will there a new MSI being build so that I'll fetch the latest msi from S3?

Makefile
dkhokhlov: now replaced by TOOLS_COMMAND.

packaging/windows/mezmo-agent.wxs
dkhokhlov: Current MSI is configured in such way that new build will overwrite old files. 
Your Windows resources PR will be good for user to see version and icon of exe. Currently version and icon are visible in Add/Remove Progs.

packaging/windows/mk
dkhokhlov: Current MSI allows patch updates in any direction. Works.

james-t-bloom (approved):

Makefile
```diff
@@ -62,6 +67,7 @@ BUILD_TIMESTAMP := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 BUILD_VERSION := $(shell sed -nE "s/^version = \"(.+)\"\$$/\1/p" bin/Cargo.toml)
 BUILD_TAG ?= $(VCS_REF)
 IMAGE_TAG := $(BUILD_TAG)-$(ARCH)
+BUILD_NUMBER := 0
```
c-nixon, line 70: Is it intended that BUILD_NUMBER is always 0? Or do we want to use `?=` here?

c-nixon (approved):
Just one question, but this looks sensible to me based on my limited experience with msi's

Makefile
dkhokhlov: thanks!

PR #289, opened 2022-02-07
k8s: added LOGDNA_HOST to deployment yamls

dkhokhlov:
- added  LOGDNA_HOST to allow to specify dev endpoints


c-nixon:
This is deliberately absent, as it's templated in as part of the CDN initialisation. If we add this here it will break the CDN

dkhokhlov:
Abandoned. 
Superseded by
https://github.com/logdna/logdna-agent-v2/pull/307

PR #290, opened 2022-02-09
Feat: Adds Metrics to Benchmark

stsantilena:
This PR extends to current benchmark project by collecting various metrics during the benchmark runtime. It then writes a summary of these metrics to STDOUT and  all gathered metrics to a `metrics_output.log` file.

Tests:
Ran on local machine as well as EC2 dev/test instance.

TODO: 
- [x] Merge #294

### DIFF
--- a/bench/bin/throughput.rs
+++ b/bench/bin/throughput.rs
@@ -571,7 +571,7 @@ async fn main() -> Result<(), std::io::Error> {
         fs_line_metrics.0, fs_line_metrics.1
     );
     println!(
-        "\nINGESTION METRICS\n . Totol Time (sec): {:?}\n . Average Duration (sec): {:?}\n . Average Request Size (bytes): {:?}",
+        "\nINGESTION METRICS\n . Total Time (sec): {:?}\n . Average Duration (sec): {:?}\n . Average Request Size (bytes): {:?}",
         ingest_time_metrics.0, ingest_time_metrics.1, ingest_size_metrics
     );
     println!(
--- a/bench/bin/throughput.rs
+++ b/bench/bin/throughput.rs
@@ -111,7 +111,7 @@ fn is_agent_metric(sample: &prometheus_parse::Sample, metric_name: &str) -> bool
     sample.metric.as_str() == full_metric_name
 }
 
-// Sample number of lines
+// Sample number of file system lines
 fn calculate_fs_line_metrics(samples: &[Sample]) -> (f64, f64) {
     let metric_name = "fs_lines";
     let fs_sample_data = samples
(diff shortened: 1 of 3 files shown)


bench/bin/throughput.rs
```diff
@@ -92,6 +99,190 @@ fn start_ingester(
...
+    }
+}
+// **** IMPORTED **************************
+
+// Sample number of lines
+fn calculate_fs_line_metrics(samples: &[Sample]) -> (f64, f64) {
+    let fs_sample_data = samples
+        .iter()
+        .filter_map(|s| match s.value {
+            Value::Counter(raw) if s.metric.as_str() == "logdna_agent_fs_lines" => Some(raw),
```
dhable, line 212: A good helper function here might encapsulate this string manipulation and property reference so you can write something like:
```
Value::Counter(raw) if is_agent_metric(s, "fs_lines") => Some(raw)
```
since we know that we're always going to prefix "logdna_agent", we always need to do this string/str conversion dance and we're always going to use the metric property. Might make some of these logic blocks less busy but would only work for metrics in the agent.

bench/bin/throughput.rs
```diff
@@ -92,6 +99,190 @@ fn start_ingester(
...
+            _ => None,
+        })
+        .zip(samples.iter().filter_map(|t| match t.value {
```
dhable, line 215: The previous `filter_map` can could also output the tuple which would reduce the number of iterations required on the samples.
```
Value::Counter(raw) if s.metric.as_str() == "logdna_agent_fs_lines" => 
    Some((raw, t.timestamp.timestamp_millis()))
```
The extra `(...)` pair creates the tuple without zipping.

bench/bin/throughput.rs
```diff
@@ -92,6 +99,190 @@ fn start_ingester(
...
+    let mut max_value: f64 = 0.0;
+    let private_virtual_memory = samples
+        .iter()
+        .filter_map(|m| match m.value {
+            Value::Gauge(raw) if m.metric.as_str() == "process_virtual_memory_bytes" => Some(raw),
+            _ => None,
+        })
+        .collect::<Vec<f64>>();
+
+        for &val in private_virtual_memory.iter() {
```
dhable, line 241: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max

bench/bin/throughput.rs
```diff
@@ -92,6 +99,190 @@ fn start_ingester(
...
+        .iter()
+        .filter_map(data_pair_for(
+            "logdna_agent_ingest_request_duration_seconds_sum",
+        ))
+        .zip(samples.iter().filter_map(data_pair_for(
+            "logdna_agent_ingest_request_duration_seconds_count",
+        )))
+        .collect::<Vec<((i64, f64), (i64, f64))>>();
+
+    let ingest_total_time = (ingest_duration_sample.last().unwrap().0.0 - ingest_duration_sample[0].0.0)/1000;
```
dhable, line 262: FYI: You could use the rust destructuring operation to pull apart the tuples and give them more meaningful names than `.0`. 
```
let ((last_sum_ts, last_sum_val), (last_count_ts, last_count_val)) = ingest_duration_sample.last().unwrap();
```
The destructuring construct is pretty powerful and works on deeply nested data. It's one of the cooler features of rust that is missing in langs like Java or C++.

bench/bin/throughput.rs
stsantilena: Implemented the max method. It apparently doesn't handle `f34` or `f64` well but, thankfully, there was a workaround listed in the docs.

bench/bin/throughput.rs
stsantilena: This was very helpful and makes the code a lot more readable. Great suggestion!

c-nixon (approved):

dhable (approved):

PR #291, opened 2022-02-10
Fix: Add Default Inclusion Rule to Include Extention-less Files

stsantilena:
The default behavior of the agent was to include only files with a `.log` extention. However, the documentation states that this it should include extention-less file as well. This PR adds a new default glob that will include extention-less files by default.

Tests:
Ran locally and confirmed that extention-less files were not being included by default, after the fix, I could see extention-less files being consumed. Also, tested that the behavior is the same regardless of glob order.

### DIFF
--- a/common/config/src/raw.rs
+++ b/common/config/src/raw.rs
@@ -376,7 +376,10 @@ impl Default for LogConfig {
             db_path: None,
             metrics_port: None,
             include: Some(Rules {
-                glob: vec!["*.log".parse().unwrap()],
+                glob: vec![
+                    "*.log".parse().unwrap(),
+                    "!(*.*)".parse().unwrap(),
+                ],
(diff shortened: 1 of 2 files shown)


common/config/src/raw.rs
```diff
@@ -463,6 +466,23 @@ mod tests {
         assert_eq!(config, new_config);
     }
 
+    #[test]
+    fn test_default_glob() {
```
stsantilena, line 470: There were two other units tests that caught this change, but they were located in `argv` so not sure if this test was necessary but added it anyway.

dkhokhlov: does the order matter in this vec?

stsantilena:
As per discussion with @alfeng6 and @c-nixon I'm closing this PR due to concerns around usage spikes.

PR #293, opened 2022-02-14
Updated Defaults in README and Include Common Glob Pattern

stsantilena:
This removes the extention-less files from the default values list and makes a note about including that pattern via environment variable.


alfeng6 (approved):
looks great!

PR #294, opened 2022-02-14
Refactor Metrics Recorder

stsantilena:
The PR is a prerequisite to #290. It pulls the existing `metrics-recorder` out of `bin/tests` and puts it into its own package so that it can be used in tests as well as bench.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,8 @@ members = [
     "common/middleware",
     "common/journald",
     "common/state",
-    "bench"
+    "bench",
+    "utils/metrics-recorder"
 ]
 
 [profile.release]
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -9,7 +9,8 @@ members = [
     "common/middleware",
     "common/journald",
     "common/state",
-    "bench"
+    "bench",
```
c-nixon, line 12: If we drop bench and utils/metrics-recorder from the workspace here we may not need the dead code annotations.

c-nixon (approved):
This looks good to me!

Cargo.toml
stsantilena: I removed the annotations without removing the packages from the workspace and I didn't get any compile errors; perhaps because it's in a separate package, the compiler does't complain.

dhable (approved):

PR #297, opened 2022-02-24
feat: added make targets for multi-arch images

dkhokhlov:
- added make targets for multi-arch images (arm64, amd64)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@ def publishDockerhubICRImages = false
 pipeline {
     agent any
     options {
-        timeout time: 2, unit: 'HOURS'
+        timeout time: 8, unit: 'HOURS'
         timestamps()
         ansiColor 'xterm'
     }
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -134,12 +157,30 @@ pipeline {
...
+                                echo "[default]" > ${PWD}/.aws_creds_static_x86_64
+                                echo "AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}" >> ${PWD}/.aws_creds_static_x86_64
+                                echo "AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}" >> ${PWD}/.aws_creds_static_x86_64
+                                ARCH=x86_64 STATIC=1 FEATURES= make build-release AWS_SHARED_CREDENTIALS_FILE=${PWD}/.aws_creds_static_x86_64
+                                rm ${PWD}/.aws_creds_static_x86_64
+                            '''
+                        }
+                    }
+                }
+                stage('Build static release binary aarch64') {
```
c-nixon, line 169: Why do we need to split this isn't 2 separate stages? The cross compile path works fine for the static builds and doesn't have fall back through the emulator

Makefile
```diff
@@ -397,12 +398,26 @@ publish-s3-binary:
...
+	$(eval TARGET_VERSIONS := $(TARGET_TAG) $(shell if [ "$(BETA_VERSION)" = "0" ]; then echo "$(BUILD_VERSION)-$(BUILD_DATE).$(shell docker images -q $(REPO):$(IMAGE_TAG)) $(MAJOR_VERSION) $(MAJOR_VERSION).$(MINOR_VERSION)"; fi))
 	@set -e; \
 	arr=($(TARGET_VERSIONS)); \
 	for version in $${arr[@]}; do \
 		echo $(1):$${version}; \
-		$(DOCKER) tag $(REPO):$(BUILD_TAG) $(1):$${version}; \
+		$(DOCKER) manifest create $(1):$${version} \
+			--amend $(1):$${version}-arm64 \
+			--amend $(1):$${version}-amd64; \
 		$(DOCKER) push $(1):$${version}; \
```
c-nixon, line 421: I believe this needs to be `docker manifest push`

c-nixon (approved):
This is great! LGTM

PR #300, opened 2022-03-04
Fix: Add Integration Test to Catch Directory Creation Bug

stsantilena:
When the agent is configured to watch a directory but that directory doesn't exist on startup, the agent drops the directory and throws a warning. This test (ported from a branch created by @jorgebay) will fail if that the agent exhibits this behavior. I've set an ignore attribute tag until the underlying issue is resolved.


bin/tests/cli.rs
```diff
@@ -1580,6 +1580,43 @@ async fn test_line_redact() {
     test_line_rules(None, None, redact, to_write, expected).await;
 }
 
+#[tokio::test]
+#[ignore]
```
stsantilena, line 1584: Adding ignore tag so test won't break pipeline.

stsantilena, line 1584: Adding ignore tag so test won't break pipeline.

c-nixon: LGTM, and reading back through the commit/CI history suggests it works :)

c-nixon (approved):

PR #301, opened 2022-03-10
K8s: Refactor K8sMetadata to Use Reflectors

stsantilena:
Draft PR for refactoring the way the `K8sMetadata` is managed using reflectors instead of manually managing.


stsantilena:
Closing this as it was addressed in #302.

PR #302, opened 2022-03-14
Update kube-rs and port to reflector

c-nixon:


dkhokhlov (changes requested):

docker/kind_dispatch.sh
```diff
@@ -24,17 +24,19 @@ _term() {
...
 
+set -x
 echo "Building Agent Image"
 DOCKER_BUILDKIT=1 docker build $curpath/.. \
   -f Dockerfile.debian \
   -t "$image" \
   --pull \
   --progress=plain \
-  --build-arg BUILD_IMAGE=docker.io/logdna/build-images:rust-buster-stable \
+  --build-arg BUILD_IMAGE=docker.io/logdna/build-images:rust-buster-1-stable-x86_64 \
```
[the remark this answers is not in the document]
c-nixon: Ahh, yeh, I'll need to pass that from the makefile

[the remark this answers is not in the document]
c-nixon: Once https://github.com/logdna/build-images/pull/26/files is working this should work on either platform

common/k8s/src/middleware/metadata.rs
```diff
@@ -1,281 +1,204 @@
...
-                    }
-                })
-                .for_each(|p| async {
-                    self.handle_pod(p)
-                        .unwrap_or_else(|e| log::warn!("unable to process pod event: {}", e));
-                })
-                .await;
-        });
-    }
+    fn run(&self) {}
```
stsantilena, line 112: I assume the reason the whole run function is able to go away is because the reflector is handling all the data management internally?

bin/src/main.rs
```diff
@@ -120,8 +121,10 @@ async fn main() {
     if config.log.use_k8s_enrichment == K8sTrackingConf::Always
         && PathBuf::from("/var/log/containers/").exists()
     {
-        match K8sMetadata::new().await {
-            Ok(v) => {
+        let node_name = std::env::var("NODE_NAME").ok();
+        match K8sMetadata::new(user_agent, node_name.as_deref()).await {
+            Ok((driver, v)) => {
+                tokio::spawn(driver);
```
stsantilena, line 127: Is the driver is spawned in a separate thread because it's continuously updating pod metadata in the background?

c-nixon: Yeh, it's a future containing the reflector stream. Spawning it onto the tokio runtime means that it will continually be polled which keeps it's store up to date

common/k8s/src/middleware/metadata.rs
c-nixon: Yeh, it's just here to fulfill the old synchronous Middleware interface. The driver returned from new will "run" it, so this doesn't need to do anything, but it needs to exist to avoid having to refactor `Middleware`.

PR #303, opened 2022-03-18
Add build image suffix support

c-nixon:
This adds the ability to specify a suffix for our build images in the Makefile and also paramaterise builds in Jenkins using it. This allows us to pin to a specific build image if we need to.

Suffixed images are implemented by https://github.com/logdna/build-images/pull/26


dkhokhlov (approved):
lgtm

PR #306, opened 2022-03-22
Log 12327 ensure agent passes user agent to k8s api

c-nixon:
This finishes off the refactor at #302 so that the k8s event api client ill also have it's user-agent set

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -119,7 +119,7 @@ async fn main() {
 
     let mut executor = Executor::new();
 
-    match create_k8s_client_default_from_env(user_agent) {
+    let k8s_event_stream = match create_k8s_client_default_from_env(user_agent) {
         Ok(k8s_client) => {
             if config.log.use_k8s_enrichment == K8sTrackingConf::Always
                 && PathBuf::from("/var/log/containers/").exists()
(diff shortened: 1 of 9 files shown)


bin/src/main.rs
```diff
@@ -118,26 +118,65 @@ async fn main() {
...
-                    "The agent could not access k8s api after several attempts: {}",
-                    e
-                );
-                error!("{}", message);
-                panic!("{}", message);
+
+    let k8s_event_stream = match create_k8s_client_default_from_env(user_agent) {
+        Ok(k8s_client) => {
+            if config.log.use_k8s_enrichment == K8sTrackingConf::Always
+                && PathBuf::from("/var/log/containers/").exists()
```
dkhokhlov, line 125: using hardcoded /var/log/containers, and few other places
side note: 
files in /var/log/containers are symlinks to files in /var/log/pods. /var/log/pods is primary location. would it be better to switch to /var/log/pods? yes, it will require patching the log file name parsing in k8s enrichment, but overall will more stable solution going forward.

c-nixon: I don't have a strong opinion, We've always using /var/log/containers as the agent doesn't really distinguish between symlinks and files, they're all reduced down to the actual inode containing the data

stsantilena (approved):
This LGTM.

PR #307, opened 2022-03-24
feat: omit/override metadata LOG-11095

dkhokhlov:
- overriding/omitting line metadata fields
- new configuration options:
  *  LOGDNA_META_APP
  *  LOGDNA_META_HOST
  * LOGDNA_META_ENV
  * LOGDNA_META_FILE
  * LOGDNA_META_K8S_FILE
  * LOGDNA_META_JSON
  * LOGDNA_META_ANNOTATIONS
  * LOGDNA_META_LABELS

[LOG-11095]

[LOG-11095]: https://logdna.atlassian.net/browse/LOG-11095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


common/middleware/src/meta_rules.rs
```diff
@@ -0,0 +1,516 @@
...
+        line: &'a mut dyn LineBufferMut,
+    ) -> Status<&'a mut dyn LineBufferMut> {
+        if !self.is_active() {
+            return Status::Ok(line);
+        }
+        //
+        // create map
+        //
+        let mut meta_map: HashMap<String, String> = self.env_map.clone();
+        if line.get_annotations().is_some() {
```
c-nixon, line 128: Anywhere we're using `is_some` followed by an unwrap we could refactor to use if let, eg

```rust
        if let Some(annotations) = line.get_annotations().as_ref() {
            for (k, v) in annotations.iter() {
                meta_map.insert(k.clone(), v.clone());
            }
        }
```

common/middleware/src/meta_rules.rs
```diff
@@ -0,0 +1,516 @@
...
+    pub env: Option<String>,
+    pub file: Option<String>,
+    pub k8s_file: Option<String>, // for k8s lines, applied after "file"
+    pub meta: Option<String>,
+    pub annotations: Option<String>,
+    pub labels: Option<String>,
+}
+
+impl MetaRulesConfig {
+    pub fn default() -> Self {
```
c-nixon, line 29: Since it's all None's we could get rid of this function and add the 

`#[derive(Default)]` annotation to the MetaRulesConfig struct and the compiler will add an `impl Default for MetaRulesConfig` for us

common/middleware/src/meta_rules.rs
```diff
@@ -0,0 +1,516 @@
...
+            env_map: os_env_hashmap(),
+            over_app: cfg.app,
+            over_host: cfg.host,
+            over_env: cfg.env,
+            over_file: cfg.file,
+            over_k8s_file: cfg.k8s_file,
+            over_meta: cfg.meta,
+            over_annotations: cfg
+                .annotations
+                .map_or_else(|| None, |str| serde_json::from_str(str.as_str()).unwrap()),
```
c-nixon, line 88: `.map_or_else(|| None,...)` is just `map(...)`, right?

Unwrapping on user provides data seems dangerous. Could we change these lines to something like the below in order to propagate the serde error?

```
            over_annotations: cfg
                .annotations
                .map(|data| serde_json::from_str(data.as_str()))
                .transpose()?
```

or even

```
            over_annotations: cfg
                .annotations.as_deref()
                .map(serde_json::from_str)
                .transpose()?
```
If we wanted to do it points free :)

k8s/agent-resources-satellite-debug.yaml
```diff
@@ -0,0 +1,202 @@
+---
```
c-nixon, line 1: wrt all 3 of these manifests, are we planning to distribute these on the cdn or are they just here for testing?

dkhokhlov: probably not, instead document new params with examples. these are temps.

common/middleware/src/meta_rules.rs
dkhokhlov: +1. updating.

common/middleware/src/meta_rules.rs
dkhokhlov: +1. updating.

common/middleware/src/meta_rules.rs
dkhokhlov: +1. updating.

docs/META.md
```diff
@@ -0,0 +1,38 @@
+# Mangling Line Metadata  
+
+You can use `LOGDNA_META_XXXXX` environment variables to override or ommit various metadata fields before they get sent to server.
```
c-nixon, line 3: > ommit

Small type, this should be `omit`

common/middleware/src/meta_rules.rs
```diff
@@ -0,0 +1,710 @@
...
+static LOGDNA_META_ANNOTATIONS: &str = "LOGDNA_META_ANNOTATIONS";
+static LOGDNA_META_LABELS: &str = "LOGDNA_META_LABELS";
+
+static K8S_LOG_DIR: &str = "/var/log/containers/";
+
+static MAP_PREFIX_ENV: &str = "env.";
+static MAP_PREFIX_ANN: &str = "annot.";
+static MAP_PREFIX_LAB: &str = "label.";
+
+lazy_static! {
```
c-nixon, line 26: For reference [Once Cell](https://docs.rs/once_cell/latest/once_cell/) is a more modern replacement for the lazy_static macro, it's a little more generic and avoids a macro for the same functionality (It's also what's likely to be upstreamed into std eventually)

common/middleware/src/meta_rules.rs
```diff
@@ -0,0 +1,710 @@
...
+            over_env: cfg.env,
+            over_file: cfg.file,
+            over_k8s_file: cfg.k8s_file,
+            over_meta: cfg.meta,
+            over_annotations: cfg.annotations.map_or_else(
+                || None,
+                |str| match serde_json::from_str(str.as_str()) {
+                    Ok(kvp) => kvp,
+                    Err(err) => {
+                        panic!(
```
c-nixon, line 97: If possible I'd prefer to propagate an error up to main. Users aren't used to seeing "panics" in logs and don't necessarily have the experience to parse them, which leads to more support calls.

If we pass it up to main as a normal error value then the existing code will already log it and exit the process

c-nixon (changes requested):
This all looks good. I think I have a proposal to reduce some of the

common/middleware/src/meta_rules.rs
dkhokhlov: +1. updated.

c-nixon (approved):
This looks good to me!

PR #308, opened 2022-03-24
WIP: Start parallelising agent CI

c-nixon:


c-nixon:
Weirdly I couldn't reproduce the issue with dash, only with _ash_ which is _ancient_. But yeh, having the jenkins user's default shell be bash or zsh would probably be less surprising for people coming from macs

PR #309, opened 2022-03-29
K8s: Use Leases to Throttle Agent Startup

stsantilena:
Adds option to use Kubernetes Lease on startup. This ensures that the agent much claim a lease from the K8's cluster before it can establish an event stream.


bin/src/main.rs
```diff
@@ -178,6 +180,82 @@ async fn main() {
...
+                                std::env::var("POD_NAME").unwrap(),
+                                &k8s_lease_api,
+                                &mut k8s_claimed_lease,
+                            )
+                            .await;
+                            break;
+                        }
+                        None => {
+                            info!("No lease availabe at this time. Waiting 1 second...");
+                            thread::sleep(Duration::from_millis(1000));
```
c-nixon, line 211: We definitely don't want to use thread::sleep here, tokio has it's own async sleep mechanism which we should use

common/k8s/src/lease.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+struct LeasePatchSpec {
+    spec: LeasePatchValue,
+}
+
+// Non snake case for JSON parsing reasons
+#[allow(non_snake_case)]
+#[derive(Debug, Serialize, Deserialize)]
+struct LeasePatchValue {
+    holderIdentity: Option<String>,
```
c-nixon, line 21: We can use serde's rename mechanism so that we can use snake case here

common/k8s/src/lease.rs
```diff
@@ -0,0 +1,123 @@
...
+            None => {
+                println!("Lease {} NOT OWNED...", lease_name);
+                return Some(lease_name);
+            }
+        }
+    }
+    None
+}
+
+pub async fn claim_lease(
```
c-nixon, line 42: We could refactor this into a 

```rust
struct Lease {
    lease_name: String,
    ...
}

impl Lease {
    fn acquire(&mut self) -> LeaseGuard {
        // Current contents of claim lease
    }
}

struct LeaseGuard {
    ...
}

impl Drop for LeaseGuard {
    fn drop(self) {
        tokio::spawn(async move {
            // current contents of release_lease
        }
    }
}
```

That way it would be a bit more rustic

bin/src/main.rs
stsantilena: Updated to use tokio.

common/k8s/src/lease.rs
stsantilena: Updated.

common/k8s/src/lease.rs
stsantilena: Thinking we handle this in a follow ticket. Thoughts @c-nixon?

bin/src/main.rs
```diff
@@ -143,7 +153,7 @@ async fn main() {
             }
 
             let k8s_event_stream = match config.log.log_k8s_events {
-                K8sTrackingConf::Never => None,
+                K8sTrackingConf::Never => None, // Agent won't release it this is set to None.
```
jakedipity, line 156: Believe there's a typo here: `it => if`

bin/src/main.rs
```diff
@@ -468,6 +494,68 @@ async fn main() {
...
+                    tokio::time::sleep(Duration::from_millis(1000)).await;
+                }
+            };
+        }
+    } else {
+        info!(
+            "Kubernetes cluster initialised, K8s startup lease set to: {:?}",
+            start_option,
+        )
+    }
```
jakedipity, line 556: The code duplication isn't great. Can you use a while with a looping condition built from the value of the `start_option`? Incoming rust psuedocode

```rust
let max_attempts = match start_option {
  Some("on") => { info!("Getting agent-startup-lease (making limited attempts)"); K8S_STARTUP_LEASE_RETRY_ATTEMPTS },
  Some("always") => { info!("Getting agent-startup-lease (trying forever)"); -1 },
  _ =>{ info!("Initialised..."); return }
}

let attempts = 0
while (max_attempts == -1) || (attempts < max_attempts) {
  // the get lease code
}
```

bin/src/main.rs
stsantilena: Good catch. I removed this comment as it no longer applies.

bin/src/main.rs
stsantilena: Thanks for the feedback @jakedipity. I've refactored base on your suggestion.

docs/KUBERNETES.md
```diff
@@ -68,6 +68,11 @@ kubectl apply -f k8s/agent-namespace.yaml
 kubectl create secret generic logdna-agent-key -n logdna-agent --from-literal=logdna-agent-key=<YOUR LOGDNA INGESTION KEY>
 kubectl apply -f k8s/agent-resources.yaml
 ```
+__Optionally:__ If you intend to use the K8's Lease option in the agent startup process, your cluster must be configured before starting the agent. To do this, run the following command in addition the commands listed above in step #2: 
+```console
+kubectl apply -f k8s/agent-startup-lease.yaml
```
jiangao1, line 73: This command can run before applying the the agent-resources.yaml and after the create secret.

docs/KUBERNETES.md
```diff
@@ -68,6 +68,11 @@ kubectl apply -f k8s/agent-namespace.yaml
...
+```
+By default, there are 5 leases created. The `agent-startup-lease.yaml` file can be modified to include additional startup leases if needed.
 3. Monitor the pods for startup success:
 ```console
 foo@bar:~$ kubectl get pods -n logdna-agent --watch
```
jiangao1, line 78: may remove the foo@bar:~$ for consistency?

docs/README.md
```diff
@@ -173,6 +173,7 @@ options are available:
 |`LOGDNA_REDACT_REGEX`|Comma separated list of regex patterns used to mask matching sensitive information (such as PII) before sending it in the log line.||
 |`LOGDNA_JOURNALD_PATHS`|Comma separated list of paths (directories or files) of journald paths to monitor||
 |`LOGDNA_LOOKBACK`|The lookback strategy on startup|`none`|
+|`LOGDNA_K8S_STARTUP_LEASE`|Determines whether or not to use K8 leases on startup||
```
jiangao1, line 176: should we add "on" or always as possible values here.

docs/README.md
```diff
@@ -228,6 +229,21 @@ The valid values for this option are:
 * If you configure the LogDNA Agent to run as non-root, review the [documentation](KUBERNETES.md#enabling-file-offset-tracking-across-restarts) about enabling "statefulness" for the LogDNA Agent.
 * When upgrading from LogDNA Agent version 3.0 to 3.1, the state file will initially be empty, so the lookback setting will be used for existing files. After that (i.e. on process restart), the state file will be present and will be used.
 
+### Configuring Lease Startup
+
+The lease startup configuration uses Kubernetes Leases to limit the number of agents that can start at one time on a cluster. When enabled, the agent will "claim" a lease before starting. Once started, the agent will then release the lease. If no leases are available, the agent will wait for one to become available. This feature would only be needed if running the agent on a cluster large enough that you'd risk crashing `etcd` if all the the agents tried to connect at once.
```
jiangao1, line 234: Should we give a number of what large means? We know 500 is a large number.

k8s/agent-startup-lease.yaml
```diff
@@ -0,0 +1,73 @@
...
+apiVersion: coordination.k8s.io/v1
+kind: LeaseList
+items:
+  - apiVersion: coordination.k8s.io/v1
+    kind: Lease
+    metadata:
+      name: logdna-agent-startup-lease-0
+      namespace: logdna-agent
+      labels:
+        process: logdna-agent-startup
```
jiangao1, line 35: We had to remove the logdna- prefix for the process value. Is it something to be fixed?

docs/README.md
stsantilena: I left a bit vague just because I don't think we know the exact number at which issue will occur and it would probably depend on a number of factors.

Perhaps @alfeng6 has some thoughts on this?

docs/KUBERNETES.md
stsantilena: I restructured, let me know if what I did makes sense.

docs/KUBERNETES.md
stsantilena: Removed.

docs/README.md
```diff
@@ -228,6 +229,21 @@ The valid values for this option are:
...
+
+The valid values for this option are:
+* When set to **`None`** (default):
+  * The agent will start normally and will not attempt to obtain a lease before starting.
+* When set to **`limited`**:
```
stsantilena, line 239: @c-nixon I changed this from "on" to "limited" because when @jiangao1 and I were running some test it looked like "on" has some special meaning in yaml on it need to be double quoted, otherwise it threw an error when you tried to apply it. I changed to remove any confusion but let me know if you think another option name would be better?

docs/README.md
dkhokhlov: As Agent  consumes some resources here would be useful to stress-test this. Like on empty cluster with N nodes to know the limits and base line.

docs/README.md
c-nixon: Hmm, yeh that's annoying about "on". Although I think "limited" might carry the wrong emphasis... I think we really we mean "opportunistic" but that's also quite long... maybe "IfAvailable"? Or maybe "attempt" and "require" instead of "always"?

@alfeng6 What do you think?

c-nixon (commented):
Logic looks good, just some minor bikeshedding around the configuration names

docs/README.md
stsantilena: I like "attempt" and "require" but let's see what @alfeng6 says.

alfeng6: I like "attempt" and "require" as well — the word gives the notion that we will actually try something.

alfeng6: "attempt" makes me think it will try a few times. "Always" makes me think it'll always keep trying.

stsantilena: Updated.

docs/README.md
stsantilena: As per the discussion with @alfeng6 we've decided to leave the option off by default.

docs/README.md
stsantilena: As per discussion, we're going to perform additional scale testing after merge.

PR #316, opened 2022-04-08
Add extra docker pull to make sure images are fixed up

c-nixon:
Docker has strange behaviour where it seems to remember the last image used for a particular tag. This can cause very strange behaviour if you use a non-host image with an explicit --platform arg and then go back to using a bare docker run


dkhokhlov (approved):
tested it using "./scripts/mk.test" on x86 host - the error has been resolved.

PR #318, opened 2022-04-13
Explicitly pass cli args to Config so that it doesn't try and parse the args for it's unit test binary

c-nixon:
This is a very small change so that we now pass the cli args explicitly to Config rather than it reading from the environment itself.

This allows us to pass arguments to the unit test binaries again eg `cargo test -- --nocapture` without Config attempting to parse them and exiting when it finds an arg it doesn't understand.

### DIFF
--- a/common/config/src/lib.rs
+++ b/common/config/src/lib.rs
@@ -3,6 +3,7 @@ extern crate log;
 extern crate humanize_rs;
 
 use std::convert::{TryFrom, TryInto};
+use std::ffi::OsString;
 use std::fs::File;
 use std::io::Read;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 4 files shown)


common/config/src/lib.rs
```diff
@@ -103,8 +104,12 @@ pub struct JournaldConfig {
 }
 
 impl Config {
-    pub fn new() -> Result<Self, ConfigError> {
-        let argv_options = ArgumentOptions::from_args_with_all_env_vars();
+    pub fn new<I>(args: I) -> Result<Self, ConfigError>
```
stsantilena, line 107: Will this change the existing behavior or just adding an additional way to pass configs?

c-nixon: It doesn't change the behaviour, it just extracts the implicit call to std::env::args_os() from inside 

```rust
    ArgumentOptions::from_args()
```
and gives us the ability to pass in an arg list ourselves. (which we now explicitly get from std::env::args_os() in main.rs)

stsantilena (approved):

PR #326, opened 2022-04-22
Cache CargoBuild in integration tests

c-nixon:
The integration tests have to build the agent in the background in order to be able to test it.

Locally cargo caching seems to handle this fine and doesn't do any extra work, on jenkins it is slower (possibly slower disks when validating the build cache?).

This updates the spawn_agent function to cache the cargo build commands in a global so that only one build can execute at once per test binary and each subsequent build with a particular feature set will reuse the CargoRun command

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -56,7 +56,6 @@ libjournald = ["journald/libjournald"]
 journald_tests = ["journald/journald_tests"]
 
 [dev-dependencies]
-lazy_static = "*"
 assert_cmd = "1"
 escargot = "0.5"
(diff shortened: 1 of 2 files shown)


c-nixon:
On the PR build this has brought the lint/test/integration-test step down from ~2 hours to ~30 mins. The qemu aarch64 build still takes a long time, but this certainly seems to help

c-nixon:
On the PR build this has brought the lint/test/integration-test step down from ~2 hours to ~30 mins. The qemu aarch64 build still takes a long time, but this certainly seems to help

dkhokhlov (approved):
I assume internally initiated builds will have zero interference now.

lgtm

PR #327, opened 2022-04-25
Feat: Added RocksDB Cache Size

stsantilena:
This adds an options to the `db_opts` that limits the cache size to 500 KB as opposed to the default 8 MB.


common/state/src/lib.rs
```diff
@@ -54,9 +57,14 @@ fn _construct_agent_state(path: &Path) -> Result<AgentState, StateError> {
 
     let path = path.join("agent_state.db");
 
+    let cache = Cache::new_lru_cache(ROCKSDB_CACHE_SIZE);
+    let mut block_options = BlockBasedOptions::default();
+    block_options.set_block_cache(&cache.unwrap());
```
c-nixon, line 62: I think we should be able to ? rather than .unwrap here. Otherwise this looks good to me. If you run the benchmarks a couple of times (to warm the page cache) with this setting is it any different than running the benchs on master?

common/state/src/lib.rs
```diff
@@ -23,6 +25,7 @@ pub use offsets::{Offset, OffsetMap};
 pub use span::{Span, SpanError, SpanVec};
 
 const OFFSET_NAME: &str = "file_offsets";
+const ROCKSDB_CACHE_SIZE: usize = 1024 * 500;
```
dkhokhlov, line 28: config option? tunable

stsantilena: I believe there is an additional ticket to expose more of these options via configuration. For this task, the scope was just reduce the default.

common/state/src/lib.rs
stsantilena: Updated the `.unwrap`.

In regards to the bench tests, I ran a total of four local tests, two from this branch and two from master. The only difference I noticed was that the branch tests (smaller cache) had a "warm up" period. For example:

```
Branch test results (lines received and lines per/sec):
- 8450969/10000000 451059/s (1st run)
- 9832146/10000000 527983/s (2nd run)

Master test results
- 9548160/10000000 510464/s (1st run)
- 9561652/10000000 512820/s (2nd run)
```
There was a larger difference between the first and second run with the smaller cache.

dkhokhlov (approved):
lgtm

c-nixon (approved):

PR #330, opened 2022-04-28
Log 12443 not found

dkhokhlov:


common/config/src/raw.rs
```diff
@@ -215,7 +215,13 @@ impl Config {
         let (final_conf, mut error_list) = merge_all_confs(indiv_confs);
 
         for conf_err in &error_list {
-            error!("error encountered loading configuration: {:?}", conf_err);
+            let mut log_err = true;
+            if let ConfigError::Io(err) = conf_err {
```
c-nixon, line 219: We can use matches! to inline the check, eg

```
if matches!(conf_err, ConfigError::Io(ref err) if !is_default_path || err.kind() != ErrorKind::NotFound){
    error!("error encountered loading configuration: {:?}", conf_err);
}
```

It's a bit more busy on the line, but it avoids the awkward "declare a bool, test a different bool, assign the first bool and check it again" dance

dkhokhlov: less readable, compression in one line not always good

c-nixon: Ahh, sorry, misread the negative logic need to flip the logic around so that it preserves the  "log the error unless it is a default path and the error is File:NotFound"

```rust
if !matches!(conf_err, ConfigError::Io(ref err) if is_default_path && err.kind() == ErrorKind::NotFound){
    error!("error encountered loading configuration: {:?}", conf_err);
}
```

dkhokhlov: there are two cases when we need to log:
1) match  Io + guard by kind
2) no match Io - w/o the guard
the flip still doe snot cover both  - specifically #2 w/o the guard

c-nixon: Hmm, I'm missing something then, I thought the logic was to log if

```
!(err.kind == NotFound  && is_default_path)
```
ie to log in any case where the  filepath is not present, but _is_ configured (non-default) and also to log on any other error.

Am I missing a case where we would not want it to log?

dkhokhlov: case 2), line 222, when error is not ConfigError::Io
basically with your change you need to have two log statement. mine - only one.

c-nixon: https://github.com/logdna/logdna-agent-v2/pull/330/files/1725bd31bfba92fc683a848c88b249c325b11d78#r861558563

That should still match if the ConfigError isn't IO as it will fail to match? so !matches!(...) will be true, which will then log?

c-nixon (approved):
LGTM

PR #332, opened 2022-05-03
Parallelise and optimise integration tests

c-nixon:
This PR does... various things to harden and speed up the integration tests, especially when running on linux


Jenkinsfile
```diff
@@ -73,35 +114,31 @@ pipeline {
...
                             """
                         }
                     }
-                    post {
-                        success {
-                            sh "make clean"
-                        }
-                    }
                 }
                 stage('Run K8s Integration Tests') {
```
stsantilena, line 122: Is there an issue with running the K8s IT's as a separate thread?

Dockerfile
```diff
@@ -43,8 +43,8 @@ RUN --mount=type=secret,id=aws,target=/root/.aws/credentials \
     export ${BUILD_ENVS?};  \
     if [ -z "$SCCACHE_ENDPOINT" ]; then unset SCCACHE_ENDPOINT; fi; \
     cargo build --manifest-path bin/Cargo.toml --no-default-features ${FEATURES} --release $TARGET_ARG && \
-    strip ./target/${TARGET}/release/logdna-agent && \
-    cp ./target/${TARGET}/release/logdna-agent /logdna-agent && \
+    llvm-strip ./target/${TARGET}/release/logdna-agent; \
```
stsantilena, line 46: Curious why the change from `strip` to `llvm-strip`?

c-nixon: That's a leftover, from the rest of the changes this was based on, will revert the non-essential stuff

Jenkinsfile
c-nixon: The side-by-side stages means they are run in parallel to the integration tests

stsantilena (approved):
This LGTM.

Jenkinsfile
```diff
@@ -51,20 +51,61 @@ pipeline {
...
-                stage('Lint, Unit and Integration Tests'){
+                stage('Integration Tests'){
                     steps {
                         script {
                             def creds = readJSON file: CREDS_FILE
                             // Assumes the pipeline-e2e-creds format remains the same. Chase
                             // refer to the e2e tests's README's authorization docs for the
                             // current structure
                             LOGDNA_INGESTION_KEY = creds["packet-stage"]["account"]["ingestionkey"]
+                            TEST_THREADS = sh (script: 'echo "$(nproc)/4"| bc', returnStdout: true).trim()
```
dkhokhlov, line 108: $ echo "$(nproc)/10"| bc
0

for
$ nproc
8

TEST_THREADS=0 works?

Makefile
```diff
@@ -48,7 +48,9 @@ BENCH_COMMAND = CACHE_TARGET="false" $(DOCKER_DISPATCH) $(BENCH_IMAGE)
 HADOLINT_COMMAND := $(DOCKER_DISPATCH) $(HADOLINT_IMAGE)
 SHELLCHECK_COMMAND := $(DOCKER_DISPATCH) $(SHELLCHECK_IMAGE)
 
-INTEGRATION_TEST_THREADS ?= 1
+TEST_THREADS ?= $(shell echo "$$(nproc)/4" | bc)
```
dkhokhlov, line 51: see line 108

Jenkinsfile
c-nixon: Nice catch, I added an extra bit of logic to make sure it's never 0. I couldn't just add 1 as it meant machines with 4 cores like ec2-fleet would try to use 2 threads, which I found was prone to breakage, but I added a slightly ugly ternary

PR #333, opened 2022-05-04
Log 12405 update k8s builds to link against target image binaries

c-nixon:
Fix the docker container build so that the build links against the libraries from the target/distribution image rather than the build image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ ARG BUILD_ENVS
 
 ARG TARGET
 
+ARG UBI_VERSION
 ENV SYSROOT_PATH="/sysroot/ubi-${UBI_VERSION}"
 
 ENV RUST_LOG=rustc_codegen_ssa::back::link=info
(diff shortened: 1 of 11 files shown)


scripts/ubi8-sysroot.sh
```diff
@@ -0,0 +1,38 @@
...
+
+TARGET_ARCH="$1"
+UBI_VERSION="$2"
+
+SYSROOT_PATH="/sysroot/ubi-$UBI_VERSION"
+
+ubi_packages="systemd-libs systemd-devel glibc glibc-devel gcc libstdc++-devel libstdc++-static kernel-headers"
+
+apt-get update 1>&2 && apt-get install --no-install-recommends -y dnf 1>&2
+dnf install --releasever=8 --forcearch="${TARGET_ARCH}" \
```
stsantilena, line 11: This is interesting. Is this using a Debian package manager to down load a Red Hat package manager?

Dockerfile
```diff
@@ -40,38 +41,42 @@ WORKDIR /opt/logdna-agent-v2
...
-RUN printf "/* GNU ld script\n*/\n\
-OUTPUT_FORMAT(elf64-%s)\n\
-GROUP ( /usr/lib64/libgcc_s.so.1  AS_NEEDED ( /usr/lib64/libgcc_s.so.1 ) )" "$(echo ${TARGET_ARCH} | tr '_' '-' )" > $SYSROOT_PATH/usr/lib64/libgcc_s.so
+ENV UBI_PACKAGES="systemd-libs systemd-devel glibc glibc-devel gcc libstdc++-devel libstdc++-static kernel-headers"
 
-# Add the actual agent source files
-COPY . .
+RUN apt-get update && apt-get install --no-install-recommends -y dnf && \
+    dnf install --releasever=8 --forcearch="${TARGET_ARCH}" \
+        --installroot=$SYSROOT_PATH/ --repo=ubi-8-baseos --repo=ubi-8-appstream \
```
dkhokhlov: UBI_VERSION is decoupled from "8..."  here?

scripts/ubi8-sysroot.sh
c-nixon: Yup, dnf supports alternative install roots, so it's a bit like debootstrap or pacstrap from debian and arch respectively in that it will allow you to install packages into a filesystem subtree.

The general approach means we can:
* Grab the actual packaged libraries and headers from pretty much any linux distro and for any ISA/architecture (or even windows)
* Install them under a directory (commonly referred to as a sysroot)
* Configure the C, C++ and Rust compilers to look for the headers/libraries under the sysroot rather than the actual root directory

Once have that (and cross compilers/binutils on the host, ie rustc, clang, lld, llvm-strip etc) you can build for all of your targets from the same base toolchain.

It's very commonly used when doing any sort of embedded work or compiling for systems that you don't have high performance hardware available for

Dockerfile
c-nixon: I have extracted it out as `$UBI_MAJOR_VERSION` I'm not 100% sure package names will remain the same if they ever release a UBI 9, so I've left the script name the same, but it _might_ "just work" if that ever happens.

dkhokhlov (approved):

stsantilena (approved):

PR #337, opened 2022-05-16
Use VCS_REF and Jenkins BUILD_NUMBER to uniquely tag images

c-nixon:
Jenkins sets BUILD_TAG. This was resulting in images being created with a tag like `3.5.0-20220516.jenkin
s-logdna-logdna-agent-v2-3.5-27-amd64`. This PR changes things such that the first 16 characters of the sha256sum of the concatentation of the git short rev and the Jenkins branch build number to ensure uniqueness


c-nixon:
There doesn't seem to be a built in "make me a random string derived from other strings" function, which surprised me too

jmoses:
Oh I mostly meant the embedded shell bits and ifs and execs and all that.  But also what you said.

PR #340, opened 2022-05-19
feat: support for GKE Autopilot

dkhokhlov:
- added agent-resources-no-cap.yaml for GKE Autopilot
- tested on GCP

LOG-12418

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,6 +103,7 @@ RUN microdnf update -y \
     && microdnf install ca-certificates libcap shadow-utils -y \
     && rm -rf /var/cache/yum \
     && chmod -R 777 . \
+    && cp /work/logdna-agent /work/logdna-agent-no-cap \
     && setcap "cap_dac_read_search+eip" /work/logdna-agent \
     && groupadd -g 5000 logdna \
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -103,6 +103,7 @@ RUN microdnf update -y \
     && microdnf install ca-certificates libcap shadow-utils -y \
     && rm -rf /var/cache/yum \
     && chmod -R 777 . \
+    && cp /work/logdna-agent /work/logdna-agent-no-cap \
```
c-nixon, line 106: This is a neat way to do it, does it increase the size of the binary much? Just wondering if it's worth creating a really tiny shim binary that we can add the setcap bit to which just forks off the agent

c-nixon (approved):
This looks good to me, one general interest question but the changes make perfect sense.

Dockerfile
dkhokhlov: it does increase image size. thought about wrapper too. it may break existing customer setups (remote debugging or custom yamls referring to agent exe). the better way is to request caps using `libcap` inside agent itself and add ability to disable that in command line. will do it in another pr after this one settles.

PR #347, opened 2022-06-01
clippy ignore

c-nixon:
Ignore newly introduced clippy lint warning.

Posted a ticket to refactor the fs module to remove the pattern that's provoking it: https://logdna.atlassian.net/browse/LOG-13066


dkhokhlov (dismissed):

dkhokhlov (approved):
safe to ignore under assumptions (per @c-nixon):
- the whole FS stream is a single future
- nothing spawns and the types currently make it impossible for the tailer logic to be spawned

PR #349, opened 2022-06-02
added libcap support

dkhokhlov:
Context: k8s & docker deployments 
- integrated libcap
- Linux capabilities (CAP_DAC_READ_SEARCH) are now optionally set in agent at startup
- default behavior as before: CAP_DAC_READ_SEARCH is enabled.
- second `logdna-agent-no-cap` executable file is now replaced by symbolic link, 
- setting capabilities can be disabled by defining LOGDNA_NO_CAP env var as in [k8s/agent-resources-no-cap.yaml](k8s/agent-resources-no-cap.yaml).


scripts/ubi8-sysroot.sh
```diff
@@ -9,7 +9,7 @@ UBI_VERSION="${UBI_MAJOR_VERSION}.${UBI_MINOR_VERSION}"
 
 SYSROOT_PATH="/sysroot/ubi-$UBI_VERSION"
 
-ubi_packages="systemd-libs systemd-devel glibc glibc-devel gcc libstdc++-devel libstdc++-static kernel-headers"
+ubi_packages="systemd-libs systemd-devel glibc glibc-devel gcc libstdc++-devel libstdc++-static kernel-headers libcap-ng-devel"
```
c-nixon, line 12: Do we still need this with capctl?

bin/src/main.rs
```diff
@@ -51,11 +52,52 @@ pub static PKG_NAME: &str = env!("CARGO_PKG_NAME");
...
+        }
+        let status =
+            std::fs::read_to_string("/proc/self/status").expect("Failed to read /proc/self/status");
+        let re = regex::Regex::new(r"(?m)^((Cap|Cpu|Seccomp|Groups|Uid|Gid).+?)$").unwrap();
+        for cap in re.captures_iter(status.as_str()) {
+            info!("{}", &cap[0]);
+        }
+    }
+
+    let _contents =
```
c-nixon, line 86: Are these couple of lines still necessary?

c-nixon (commented):
This looks good to me. I'm actually wondering if we really need `-no-cap` now or if we could just have the basic agent adopt the caps dynamically? I think that should always be backwards compatible with previous releases, right?


 WRT testing do you think it makes sense to add a `user` block to the k8s integration tests over at https://github.com/logdna/logdna-agent-v2/blob/ef5d424fdfa3d34f39bb633944f4b1dee488217f/bin/tests/it/k8s.rs#L602 so that the agent runs as non-root and maybe with *-no-cap by default?

bin/src/main.rs
dkhokhlov: leftover

scripts/ubi8-sysroot.sh
dkhokhlov: nope

c-nixon (approved):
LGTM!

bin/tests/it/k8s.rs
```diff
@@ -600,6 +600,9 @@ fn get_agent_ds_yaml(
                                 }
                             },
                             "securityContext": {
+                                "runAsUser": 5000,
```
dkhokhlov, line 603: running as non-root during k8s tests

bin/tests/it/k8s.rs
```diff
@@ -600,6 +600,9 @@ fn get_agent_ds_yaml(
...
+                                "runAsGroup": 5000,
+                                "fsGroup": 5000,
```
dkhokhlov, line 605: running as non-root during k8s tests

PR #355, opened 2022-06-14
MZ_ environment variables

dkhokhlov:
- introduced MZ_ environment variables in code only, mirroring LOGDNA_ ones
- LOGDNA_ env vars  are still functional. 
- old yaml files are supported. 
- changing deployment/service/shell/environment files is not required.
- also see header note in README


common/config/src/lib.rs
```diff
@@ -151,12 +151,21 @@ pub struct K8sStartupLeaseConfig {
     pub option: String,
 }
 
+const LOGDNA_PREFIX: &str = "LOGDNA_";
+const MEZMO_PREFIX: &str = "MZ_";
+
 impl Config {
+    pub fn new_from_env() -> Result<Self, ConfigError> {
+        Config::process_logdna_env_vars();
```
stsantilena, line 159: A bit confused why this new constructor wrapper calls `process_logdna_env_vars()` when it looks like the old `new()` constructor calls it as well?

dkhokhlov: leftover. thanks!

stsantilena:
In regards to tests that use the legacy prefix, are we okay leaving those as is and only using new prefix going forward? For example, in `..test/it/k8s.rs`. This more of a style question since, as you mention, the legacy format will still work.

dkhokhlov:
> In regards to tests that use the legacy prefix, are we okay leaving those as is and only using new prefix going forward? For example, in `..test/it/k8s.rs`. This more of a style question since, as you mention, the legacy format will still work.

we will keep old vars for awhile.  so need to test them too. 
https://github.com/logdna/logdna-agent-v2/tree/LOG-13041-introduce-mz-environment-variables#logdna-agent

common/config/src/lib.rs
```diff
@@ -205,7 +200,7 @@ impl Config {
         Config::try_from(raw_config)
     }
 
-    fn process_logdna_env_vars() {
+    pub fn process_logdna_env_vars() {
```
stsantilena, line 203: Just curious as to why you chose to make this public and explicitly call in main, as opposed to keeping private and calling within the existing `new()` constructor?

dkhokhlov: it was like that before. it delta diff with per commit. full diff in Files changed.

stsantilena (approved):
This LGTM!

common/config/src/lib.rs
```diff
@@ -196,6 +199,21 @@ impl Config {
...
+            .filter(|(n, _)| {
+                n.clone()
+                    .into_string()
+                    .unwrap_or_default()
+                    .starts_with(LOGDNA_PREFIX)
+            })
+            .for_each(|(name, value)| {
+                let new_name = MEZMO_PREFIX.to_string()
+                    + &name.into_string().unwrap_or_default()[LOGDNA_PREFIX.len()..];
+                std::env::set_var(new_name, value);
```
c-nixon, line 214: I tried to work out a way of doing this without mutating the process environment, but I can't see a nice way to extend the StructOpt parsing to have multiple env vars. We could macro it, but it would pretty horrible. So this is probably the lesser evil

c-nixon (approved):
This looks good to me

PR #363, opened 2022-06-27
Feat: K8s Lease Startup Option to Enum

stsantilena:
This PR does two things:

1. It refactors the existing `parse_k8s_tracking_or_warn` to take a generic type so it can be used for parsing both the `K8sTrackingConf` as well as the `K8sLeaseConf`.

2. Changes the existing K8s startup lease configuration option to use an Enum (similar to K8sTrackingConf) to make it more robust.


common/config/src/lib.rs
```diff
@@ -452,13 +442,16 @@ fn print_settings(yaml: &str, config_path: &Path) {
     std::process::exit(0);
 }
 
-fn parse_k8s_tracking_or_warn(
+fn parse_k8s_enum_config_or_warn<T: FromStr + std::fmt::Display + std::fmt::Debug>(
```
stsantilena, line 445: Refactored this helper function to take a generic type so I could use it for both `K8sTrackingConf` and `K8sLeaseConf`.

james-t-bloom (approved):

PR #367, opened 2022-07-06
feat(reporter): Port reporter functionality

james-t-bloom:
Summary:
Port metrics server statistics functionality
Test plan:
Integration Test
Unit Test
Manual Testing via Web App

Ref: LOG-12885


common/k8s/src/kube_stats/helpers.rs
```diff
@@ -0,0 +1,191 @@
+pub fn convert_cpu_usage_to_milli(cpu: &str) -> i32 {
```
dkhokhlov, line 1: returns some number in all cases, no errors. zero could be misleading value.

common/k8s/src/kube_stats/cluster_stats.rs
```diff
@@ -0,0 +1,70 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+pub struct ClusterStats {
+    pub resource: String,
+    pub r#type: String,
+    pub containers_init: i32,
```
dkhokhlov, line 7: signed, no unsigned?

common/k8s/src/kube_stats/cluster_stats.rs
```diff
@@ -0,0 +1,70 @@
...
+impl Default for ClusterStats {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ClusterStats {
+    pub fn new() -> ClusterStats {
+        ClusterStats {
+            containers_init: 0,
```
dkhokhlov, line 42: wrapped in Option? to handle cases when metric is missing and excluded from serialization.

common/k8s/src/kube_stats/container_stats.rs
```diff
@@ -0,0 +1,437 @@
...
+        let running = self.c_state.running.as_ref();
+        let terminated = self.c_state.terminated.as_ref();
+
+        if let Some(..) = running {
+            state = "Running".to_string();
+
+            let started_at = running.unwrap().started_at.as_ref().map(|s| s.0);
+
+            if let Some(..) = started_at {
+                container_age = Local::now()
```
c-nixon, line 135: Is Local the right TZ to be using? Afair all kubernetes timestamps should have timezone included in them (eg "2022-07-01T12:48:58Z" which I just grabbed off a pod and specifies UTC+-0).

If possible I'd prefer to use UTC here to avoid the chrono dependency since there's essentially a WONTFIX security issue around the localtime_r libc function that chrono relies on

common/k8s/src/kube_stats/container_stats.rs
```diff
@@ -0,0 +1,437 @@
...
+
+                started = started_at.unwrap().timestamp_millis();
+            }
+        } else if terminated.is_some() {
+            state = "Terminated".to_string()
+        } else {
+            state = "Waiting".to_string()
+        }
+
+        let last_status_state = self.c_status.last_state.as_ref();
```
c-nixon, line 147: Can we use something like the following to avoid the unwraps on line 150, etc?

```rust
if let Some(last_status_state) = self.c_status.last_state.as_ref() {

}
```

c-nixon: Same for all the other `if let Some(..)` blocks

common/k8s/src/kube_stats/cluster_stats.rs
james-t-bloom: all of the cluster stuff - or just usage/cap/allocatable?

I think zeros on the rest of the stats except for usage/allocatable/limits make sense

common/k8s/src/kube_stats/helpers.rs
james-t-bloom: really good call

common/k8s/src/kube_stats/cluster_stats.rs
```diff
@@ -0,0 +1,70 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
```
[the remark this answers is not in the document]
james-t-bloom: its serialized and sent through along with pod & node stats. 

`fn format_cluster_str(cluster_stats: &ClusterStats) -> String {
    let cluster_string = format!(
        r#"{{"kube":{}}}"#,
        serde_json::to_string(&cluster_stats).unwrap_or_else(|_| String::from(""))
    );
    cluster_string
}`

common/k8s/Cargo.toml
```diff
@@ -11,28 +11,31 @@ http = { package = "http", path = "../http" }
...
 log = "0.4"
-tokio = { version = "1", features = ["rt", "rt-multi-thread"] }
+tokio = { package = "tokio", version = "1", features = ["macros", "process", "rt-multi-thread", "time"] }
 futures = "0.3"
 thiserror = "1.0"
 parking_lot = "0.11"
-kube = { version = "0.69", default-features = false, features = ["rustls-tls", "client", "runtime", "gzip"] }
-k8s-openapi = { version = "0.14", default_features = false, features = ["v1_16"] }
+kube = { version = "0.73.0", default-features = false, features = ["rustls-tls", "client", "runtime", "gzip"] }
+k8s-openapi = { version = "0.15", default_features = false, features = ["v1_24"] }
```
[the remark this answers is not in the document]
james-t-bloom: your limiting my power

1_18 seems to work - it'll be downgraded in the next push

common/k8s/src/kube_stats/cluster_stats.rs
james-t-bloom: changed the ones that aren't mirroring the types from the kube lib.

common/k8s/src/kube_stats/container_stats.rs
james-t-bloom: fixed - good call on this. Local was a weird choice.

common/k8s/src/kube_stats/container_stats.rs
james-t-bloom: cleaned up a bunch of these

common/config/src/lib.rs
```diff
@@ -355,6 +356,11 @@ impl TryFrom<RawConfig> for Config {
                 env_vars::LOG_K8S_EVENTS,
                 K8sTrackingConf::Never,
             ),
+            log_metric_server_stats: parse_k8s_tracking_or_warn(
```
stsantilena, line 359: Just FYI, I refactor this helper function in #363

common/config/src/argv.rs
```diff
@@ -121,6 +121,13 @@ pub struct ArgumentOptions {
     #[structopt(long, env = env_vars::LOG_K8S_EVENTS)]
     log_k8s_events: Option<K8sTrackingConf>,
 
+    /// Determines whether the agent should produce metrics-servers usage logs.
+    /// When set this agent will poll the metrics-server on the cluster to get usage statistics
+    /// combine them with other pod/node statistics for enhancements to the web application.
+    /// Defaults to "off".
+    #[structopt(long, env = env_vars::LOG_METRIC_SERVER_STATS)]
+    log_metric_server_stats: Option<K8sTrackingConf>,
```
stsantilena, line 129: Does having the use the `Options<K8sTrackingConf>` make this dependent on the other two feature that using this option type?

.gitignore
```diff
@@ -8,3 +8,8 @@
 .DS_Store
 *.stamp
 /mk*
+.vscode/launch.json
+.vscode/settings.json
+.aws_creds_static
+docker/.kind_config
+docker/.kind_config_host
```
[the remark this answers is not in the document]
james-t-bloom: good catch

bin/src/main.rs
```diff
@@ -239,6 +241,28 @@ async fn _main() {
         }
     };
 
+    // TODO Leader Election
+    let metric_stats_source = match config.log.log_metric_server_stats {
```
c-nixon, line 245: I think this should be moved up to line ~220 or so, inside the match block before we release the lease.

This would also let us reuse the k8s_client created there, just need to move this block up and change the return from that block to `(k8s_event_stream, metric_stats_source)` then adjust the let binding up at line 162 to unpack it

common/k8s/src/metrics_stats_aggregator.rs
```diff
@@ -0,0 +1,728 @@
...
+
+    for node_stat in node_stats {
+        cluster_stats.containers_init += node_stat.containers_init;
+        cluster_stats.containers_ready += node_stat.containers_ready;
+        cluster_stats.containers_running += node_stat.containers_running;
+        cluster_stats.containers_terminated += node_stat.containers_terminated;
+        cluster_stats.containers_total += node_stat.containers_total;
+        cluster_stats.containers_waiting += node_stat.containers_waiting;
+
+        if node_stat.cpu_usage.is_some() {
```
c-nixon, line 176: This is fairly verbose and repetitive, we could get rid of all of the temporaries and if statements with something like

```rust
        cluster_stats.cpu_usage = cluster_stats.cpu_usage.map_or(node_stat.cpu_usage, |current| {
            node_stat.cpu_usage.map(|new| current + new)
        });

```

That way we wouldn't need the bools or the temporary variables, we'd just accumulate straight into the cluster_stats struct

c-nixon: And if we wanted to macro _that_ away we could rewrite the wholle thing to the following

```rust

macro_rules! aggregate_stat {
    ($acc_name:ident, $var_name:ident, $field_name:ident) => {
        $acc_name.$field_name = $acc_name.$field_name.map_or($var_name.$field_name, |current| {
            $var_name.$field_name.map(|new| current + new)
        });


    }
}

fn build_cluster_stats(node_stats: &Vec<NodeStats>) -> ClusterStats {
    let mut cluster_stats = ClusterStats::new();

    for node_stat in node_stats {
        cluster_stats.containers_init += node_stat.containers_init;
        cluster_stats.containers_ready += node_stat.containers_ready;
        cluster_stats.containers_running += node_stat.containers_running;
        cluster_stats.containers_terminated += node_stat.containers_terminated;
        cluster_stats.containers_total += node_stat.containers_total;
        cluster_stats.containers_waiting += node_stat.containers_waiting;

        aggregate_stat!(cluster_stats, node_stat, cpu_usage);
        aggregate_stat!(cluster_stats, node_stat, memory_usage);
        aggregate_stat!(cluster_stats, node_stat, cpu_allocatable);
        aggregate_stat!(cluster_stats, node_stat, cpu_capacity);
        aggregate_stat!(cluster_stats, node_stat, memory_allocatable);
        aggregate_stat!(cluster_stats, node_stat, memory_capacity);
        aggregate_stat!(cluster_stats, node_stat, pods_allocatable);
        aggregate_stat!(cluster_stats, node_stat, pods_capacity);

        cluster_stats.pods_failed += node_stat.pods_failed;
        cluster_stats.pods_pending += node_stat.pods_pending;
        cluster_stats.pods_running += node_stat.pods_running;
        cluster_stats.pods_succeeded += node_stat.pods_succeeded;
        cluster_stats.pods_total += node_stat.pods_total;
        cluster_stats.pods_unknown += node_stat.pods_unknown;
        cluster_stats.nodes_total += 1;

        if node_stat.ready.unwrap_or(false) {
            cluster_stats.nodes_ready += 1;
        } else {
            cluster_stats.nodes_notready += 1;
        }

        if node_stat.unschedulable.unwrap_or(false) {
            cluster_stats.nodes_unschedulable += 1;
        }
    }
    cluster_stats
}

```

common/k8s/src/kube_stats/cluster_stats.rs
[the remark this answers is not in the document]
james-t-bloom: 🤦

common/config/src/argv.rs
stsantilena: Yeah, I'm not sure if it was meant to be general purpose or if the other two configurations needed to be coupled and that's why it was written that way.

common/k8s/src/metrics_stats_aggregator.rs
```diff
@@ -0,0 +1,672 @@
...
+}
+
+impl MetricsStatsAggregator {
+    pub fn new(client: Client) -> Self {
+        Self { client }
+    }
+
+    pub fn start_metrics_call_task(self) -> impl Stream<Item = LineBuilder> {
+        stream::unfold(self.client, |client| async {
+            sleep(Duration::from_millis(30000)).await;
```
dkhokhlov, line 38: parameterize | named static?

common/k8s/src/metrics_stats_aggregator.rs
```diff
@@ -0,0 +1,672 @@
...
+                        )
+                        .map(|line| {
+                            LineBuilder::new()
+                                .line(line)
+                                .file("logdna-reporter".to_string())
+                        }),
+                    ),
+                    Err(e) => {
+                        error!("Failed To Gather Metrics Server Info {}", e);
+                        None
```
dkhokhlov, line 56: propagate the error up?

common/k8s/src/metrics_stats_aggregator.rs
```diff
@@ -0,0 +1,672 @@
...
+
+    build_pod_metric_map(pod_metrics, &mut pod_usage_map);
+    process_pods(
+        pods,
+        &mut controller_map,
+        pod_usage_map,
+        &mut extended_pod_stats,
+        &mut node_pod_counts_map,
+        &mut node_container_counts_map,
+    );
```
dkhokhlov, line 92: always successful?

common/k8s/src/metrics_stats_aggregator.rs
james-t-bloom: good call

common/k8s/src/metrics_stats_aggregator.rs
james-t-bloom: It can fail - but the error will propagate to the calling function - which handles. 

This being a private function - it won't be called elsewhere.

common/k8s/src/metrics_stats_aggregator.rs
james-t-bloom: this gets expanded on a bit in my next PR, around the leader election. 

My personal preference though as this stream just keeps running is to inform on the error, as its probably the metrics-server not running or a config issue, and keep trying every once in awhile. Rather than propagating up to the main agent flow, this feature is fairly niche and kind of an extra thing. 

The user also gets informed on the UI if something is wrong.

beefcheeks (approved):
While I can't say I can read rust well, it seems like the general logic of this PR is solid. I would normally ask minor questions around where lowercasing occurs for things like SI unit value checks, but i think this is mostly my lack of context and experience with rust. The use of kind to bootstrap integration is tests is great and overall the logic flow seems very similar to the original github.com/answerbook/logdna-reporter repo. Although my approval is definitely not a signal that 'this is correct and will work', the general logic/flow of it looks good to me.

common/k8s/src/metrics_stats_aggregator.rs
james-t-bloom: misunderstood this comment - moved to private chat

docker/kind/metrics-server.yaml
```diff
@@ -0,0 +1,196 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    k8s-app: metrics-server
+  name: metrics-server
+  namespace: kube-system
```
stsantilena, line 7: Just curious why this is living in the kube-system namespace as opposed to say, default?

docs/README.md
```diff
@@ -295,6 +296,16 @@ __Note:__ The default option is `never`.
 
 > :warning: Due to a ["won't fix" bug in the Kubernetes API](https://github.com/kubernetes/kubernetes/issues/41743), the LogDNA agent collects events from the entire cluster, including multiple nodes. To prevent duplicate logs when running multiple pods, the LogDNA agent pods defer responsibilty of capturing events to the oldest pod in the cluster. If that pod is down, the next oldest LogDNA agent pod will take over responsibility and continue from where the previous pod left off.
 
+### Configuring Metrics Reporter Statistics
```
stsantilena, line 299: Do you think it's worth add a little more information here? Maybe explain how this feature differs from other features similar to it, like the K8's events/enrichment? Is this eventually going to replace the current features?

I could see an argument both ways as adding too much information might be confusing to the client.

james-t-bloom: I don't think you're wrong, as I get confused. I'll add a bit more.

docker/kind/metrics-server.yaml
james-t-bloom: excellent question

james-t-bloom: :) I honestly am not sure - I think just because it's in the same category of cluster wide functionality etc on kube. 

This is the actual server that calculates usage statistics across the cluster - and this yaml was ripped directly from their deploy with very minor tweeks.

stsantilena (approved):
This LGTM

c-nixon (approved):

PR #377, opened 2022-07-19
Log 13584 deserialize params as paramsbuilder

c-nixon:
This adds a trivial custom deserializer for the Params field that first attempts to deserialize it as a ParamsBuilder object and ensures that hostname is correctly defaulted rather than failing to parse the config

### DIFF
--- a/common/config/src/raw.rs
+++ b/common/config/src/raw.rs
@@ -739,6 +739,51 @@ startup: {}
     }
 
     #[test]
+    fn test_yaml_file_properties_with_no_hostname_or_now_in_params() -> io::Result<()> {
+        let dir = tempdir()?;
+        let file_name = dir.path().join("test.yml");
+        fs::write(
+            &file_name,
+            "
+http:
+  host: logs.logdna.prod
+  endpoint: /path/to/endpoint1
+  use_ssl: false
+  timeout: 12000
+  use_compression: true
+  gzip_level: 1
+  params:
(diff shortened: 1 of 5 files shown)


dkhokhlov (approved):

common/config/src/raw.rs
```diff
@@ -70,6 +70,19 @@ where
...
+fn params_deser<'de, D>(data: D) -> Result<Option<Params>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let mut params_builder: ParamsBuilder = Deserialize::deserialize(data)?;
+    Ok(params_builder.build().ok().or_else(|| {
+        params_builder
+            .hostname(get_hostname().unwrap_or_default())
+            .build()
+            .ok()
```
stsantilena, line 82: Defaulting only if it can't parse the initial value from file?

stsantilena (approved):
This LGTM

common/config/src/raw.rs
c-nixon: Exactly.

It derualises the params field, if there is one, into a ParamsBuilder and then uses the Builder's built.in validation logic to check if the required fields are populated (ie hostname) if hostname is not provided it sets hostname and then builds a params object from the now fully populated builder.

Any otger implementation I could think of either required hostname in order to populate params, which is bad as you may well want to populate params.tags and default the hostname (which we already do for env var parsing).

PR #380, opened 2022-07-20
[Fix] Change K8s Startup Lease Default Value

stsantilena:
This changes the K8's lease startup default value from `off` -> `never`. The reason being that `off` is a special keyword in yaml so if, for some reason, a user wanted to explicitly set the default to `off`, they would need put the value in quotes to ensure it was handled as a string, i.e. `"off"`. 

With this change the default value is in line with other default values and doesn't need to be quoted if used explicitly.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -194,7 +194,7 @@ options are available:
 |`LOGDNA_REDACT_REGEX`|Comma separated list of regex patterns used to mask matching sensitive information (such as PII) before sending it in the log line.||
 |`LOGDNA_JOURNALD_PATHS`|Comma separated list of paths (directories or files) of journald paths to monitor||
 |`LOGDNA_LOOKBACK`|The lookback strategy on startup|`none`|
-|`LOGDNA_K8S_STARTUP_LEASE`|Determines whether or not to use K8 leases on startup|`Off`|
+|`LOGDNA_K8S_STARTUP_LEASE`|Determines whether or not to use K8 leases on startup|`never`|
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -194,7 +194,7 @@ options are available:
 |`LOGDNA_REDACT_REGEX`|Comma separated list of regex patterns used to mask matching sensitive information (such as PII) before sending it in the log line.||
 |`LOGDNA_JOURNALD_PATHS`|Comma separated list of paths (directories or files) of journald paths to monitor||
 |`LOGDNA_LOOKBACK`|The lookback strategy on startup|`none`|
-|`LOGDNA_K8S_STARTUP_LEASE`|Determines whether or not to use K8 leases on startup|`Off`|
+|`LOGDNA_K8S_STARTUP_LEASE`|Determines whether or not to use K8 leases on startup|`never`|
```
stsantilena, line 197: @alfeng6 and @c-nixon if the user isn't using the feature, ideally we'd want them to NOT include `LOGDNA_K8S_STARTUP_LEASE` at all in there config; rather trying to explicitly set the default. If we want to encourage that, do you think it's better to have the default listed, or leave this blank?

c-nixon: I think we should full doc everything, we can customise the config printing to keep extraneous settings out of the agent's logs, but I think a user reading the docs should be able to find everything there

c-nixon (approved):

docs/README.md
stsantilena: Sounds good. I'll leave default in the docs.

PR #381, opened 2022-07-21
[Bug] Keep Track of Ignored Files to Prevent Verbose Logging

stsantilena:
This PR adds an additional field to the `FileSystem` struct to keep track of ignored file. It then uses the new field to keep from continuously logging about the same ignored file.

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -680,8 +680,8 @@ impl FileSystem {
             if let false = self.ignored_dirs.contains(path) {
                 self.ignored_dirs.insert(path.to_path_buf());
                 info!("ignoring {:?}", path);
-                return Ok(None);
             }
+            return Ok(None);
         }
 
         // If we already know about it warn and return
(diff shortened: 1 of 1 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -673,8 +676,12 @@ impl FileSystem {
         debug!("inserting {:?}", path);
 
         if !self.passes(path, _entries) {
-            info!("ignoring {:?}", path);
-            return Ok(None);
+            // Do not continuously log ignored files
+            if let false = self.ignored_dirs.contains(path) {
+                self.ignored_dirs.insert(path.to_path_buf());
+                info!("ignoring {:?}", path);
+                return Ok(None);
```
c-nixon, line 683: We should still return early, even if we've seen it before, so we should move this return back down a line

c-nixon (changes requested):
Just one tweak, then this lgtm

common/fs/src/cache/mod.rs
stsantilena: Good catch! Fixed.

c-nixon (approved):

PR #382, opened 2022-07-23
Rewrite config params handling to use ParamsBuilder everywhere

c-nixon:
Had to completely rewrite the Params config handling to make it work properly in all cases.


stsantilena:
This looks to be a continuation of #377?

dkhokhlov (approved):
will be nice to use config file w/o specifying defaults. approving as-is to unblock testing.

PR #384, opened 2022-07-27
feat: winres support

c-nixon:
This allows us to use rc.exe to link manifest files in they way visual studio usually does.

Before:
```
peres -v target/x86_64-pc-windows-msvc/release/logdna-agent.exe
WARNING: Resource directory VA is zero [at resources.c:582]
```
After:
```
DEBUG: id=0x10, dataOffset=0x80000018
DEBUG: id=0x1, dataOffset=0x80000030
DEBUG: id=0x409, dataOffset=0x48
DEBUG: CodePage=0, OffsetToData=19087456[0x1234060], Reserved=0[0], Size=472[0x1d8]
File Version:                    65263.1213.1.0
Product Version:                 3.6.0.0
```
This lets us embed fairly arbitrary information in the agent exe in a way that windows tooling expects (for example being able to set an icon that shows up in explorer etc, version information that appears in the file's property pane etc.)

This would also allow us to use something like 

```
<Product [...] Version="!(bind.FileVersion.File1)" ...
 />
```

in the wix config to pull the version straight out of the agent binary so they can't get out of sync rather than having to do a dance around in the makefile

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1149,8 +1149,8 @@ dependencies = [
  "pcre2",
  "pin-project-lite",
  "pin-utils",
- "regex",
  "rate-limit-macro",
+ "regex",
  "serde_json",
  "slotmap",
  "smallvec",
@@ -2084,6 +2084,7 @@ dependencies = [
  "tracing-subscriber",
  "types",
  "wait-timeout",
+ "winres",
  "winservice",
 ]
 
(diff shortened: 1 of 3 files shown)


dkhokhlov:
+1, to track version

dkhokhlov:
@c-nixon 
what would be easy way to add agent version info from bin cargo?

dkhokhlov (approved):
lgtm

PR #385, opened 2022-08-03
[WIP] Log Directories Created After Startup not Tracked

stsantilena:


common/fs/src/cache/dir_path.rs
```diff
@@ -74,3 +75,59 @@ impl From<DirPathBuf> for PathBuf {
...
 }
+
+// Recursive directory back off
+fn find_valid_path(
+    path: Option<PathBuf>,
+    postfix: Option<PathBuf>,
+) -> Result<DirPathBuf, DirPathBufError> {
+    match path {
+        Some(p) => {
+            if p.is_dir() {
```
dkhokhlov, line 86: does it work for:
- it is a file
- permission error

common/fs/src/cache/dir_path.rs
```diff
@@ -74,3 +75,59 @@ impl From<DirPathBuf> for PathBuf {
...
+                Ok(DirPathBuf { inner: p, postfix })
+            } else {
+                warn!("{:?} is not a directory; moving to parent directory", p);
+                let mut postfix_pathbuf = PathBuf::new();
+                let mut pop_pathbuf = PathBuf::new();
+                if let Some(path) = postfix {
```
dkhokhlov, line 92: shadowing fn arg on purpose?

common/fs/src/cache/dir_path.rs
```diff
@@ -74,3 +75,59 @@ impl From<DirPathBuf> for PathBuf {
...
+        None => None,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_level_up() {
```
dkhokhlov, line 123: find_valid_path has various corner cases...

common/fs/src/cache/dir_path.rs
stsantilena: The `postfix` function input is only non-None in the event that a given directory doesn't exist, in which case it saves that piece of the path so it can properly construct the missing directory path.

common/fs/src/cache/dir_path.rs
stsantilena: True. In this case the `level_up` function was easy to cover in a simple unit test. There are other unit and integration test that do seem to cover path creation. Are there any specific cases you suggest testing?

common/fs/src/cache/dir_path.rs
stsantilena: The configuration is meant to be set as directory so a single file wouldn't be expected. I followed same pattern as existing directory check. As for permissions, in theory, I think could fail is for some reason it was running in non-root mode and didn't have directory access.

common/config/src/lib.rs
```diff
@@ -344,6 +344,7 @@ impl TryFrom<RawConfig> for Config {
                 // Filter off paths that are not directories and warn about them
                 .filter_map(|d| {
                     d.clone()
+                        //TODO: Need to fix this iteration as warning now happens in dir_path.rs
```
stsantilena, line 347: @c-nixon I don't think there needs to any change here, perhaps the warning might be different message?

common/fs/src/cache/dir_path.rs
```diff
@@ -27,18 +28,18 @@ impl Deref for DirPathBuf {
...
+            Ok(p) => Ok(p),
+            _ => Err(DirPathBufError::NotADirPath(path)),
         }
 
         #[cfg(windows)]
         {
-            Ok(DirPathBuf { inner: path })
+            Ok(DirPathBuf {
+                inner: path,
+                postfix: None,
```
stsantilena, line 41: Had to add this line to fix the Windows build. This shouldn't affect existing Windows behavior.

bin/tests/it/cli.rs
```diff
@@ -1573,7 +1573,6 @@ async fn test_line_redact() {
     test_line_rules(None, None, redact, to_write, expected).await;
 }
 
-#[ignore]
 #[tokio::test]
 async fn test_directory_created_after_initialization() {
```
stsantilena, line 1577: Turned on bug specific integration test from previous PR.

common/fs/src/cache/dir_path.rs
stsantilena: @dkhokhlov there is also an integration test that was merged before this PR to specifically test, I activated it in the last commit.

common/fs/src/cache/dir_path.rs
dkhokhlov: i was thinking about var name

common/fs/src/cache/dir_path.rs
dkhokhlov: root, w/o any "/", file, invalid path

stsantilena: Added additional code coverage around `find_valid_path` function.

common/fs/src/cache/dir_path.rs
```diff
@@ -74,3 +83,111 @@ impl From<DirPathBuf> for PathBuf {
...
+                let root_pathbuf = Path::new("/");
+                let mut postfix_pathbuf = PathBuf::new();
+                let mut pop_pathbuf = PathBuf::new();
+                if let Some(path) = postfix {
+                    pop_pathbuf.push(path);
+                }
+                let parent = match level_up(&p) {
+                    Some(p) => {
+                        if p == root_pathbuf {
+                            panic!("configured directory recursed to root!");
```
stsantilena, line 107: Set to panic if function recurses back to the root directory. This will watch the entire root file system which is most likely not desirable.

common/fs/src/cache/dir_path.rs
stsantilena: Ah, I see. Updated var name `path` -> `pop`

common/fs/src/cache/dir_path.rs
stsantilena: Changed this so that instead of throwing a panic, it will warn and watch root level for missing dir.

common/config/src/lib.rs
dkhokhlov: if we allow to specify non-existing dirs - why do we filter them here?

stsantilena: This iteration just filters out anything that throws an error. Originally, if a directory didn't exist and error was thrown. Now, there is a more complex process happening down in `find_valid_path` that allows directories that don't yet exist; still propagating errors up but for different reasons. So really just the comment is no longer valid.

c-nixon (approved):
This all looks good to me!

common/fs/src/cache/dir_path.rs
```diff
@@ -74,3 +83,183 @@ impl From<DirPathBuf> for PathBuf {
...
+    postfix: Option<PathBuf>,
+) -> Result<DirPathBuf, DirPathBufError> {
+    match path {
+        Some(p) => {
+            if matches!(std::fs::canonicalize(&p), Ok(_)) {
+                Ok(DirPathBuf { inner: p, postfix })
+            } else {
+                warn!("{:?} is not a directory; moving to parent directory", p);
+                let root_pathbuf = Path::new("/");
+                let mut postfix_pathbuf = PathBuf::new();
```
c-nixon, line 103: I feel like there might be some way to use the std::path `Component` types to implement this, but that might just complicate the logic

stsantilena: There's probably a few different ways to go about it. The tricky part is flipping the path each integration to make sure it comes out correct, and plays nicely with the path rules.

PR #387, opened 2022-08-08
Log 13560 kube client warn

james-t-bloom:


bin/src/_main.rs
```diff
@@ -213,6 +214,7 @@ pub async fn _main(
 
             k8s_event_stream
         }
+        Err(K8sError::K8sNotInClusterError()) => None,
```
c-nixon, line 217: Probably best to have a debug! message here rather than completely swallowing it, just in case someone is doing something weird with the agent

james-t-bloom: done

c-nixon (approved):
Looks good to me!

PR #388, opened 2022-08-09
Log 13434 leader

james-t-bloom:
This adds leader election to the agent - currently it is just wrapped around the metrics reporter.
One leader is chosen in a cluster, and will hold on to that leader indefinitely. The only way a leader should be lost is if a pod goes down, or if for some reason execution hangs long enough that it is unable to update the leader lease.
Pods not elected leader will poll with a duration of the leaseDurationInSeconds. Or 5 minutes if that for whatever reason is not set.
Polling to become the leader has a built in jitter used by a random interval of ten seconds.
See:
leader-election-leases.yaml
for lease definition.
Leases don't have a concept of expire - this must be done with manually checking.
The metrics reporter stream is set to be created, and then blocked until such time it becomes the leader.

Test plan:
Integration Test
Unit Test
Manual Testing via observing leader logs

Ref: LOG-13434


dkhokhlov:
What happens at daemon set startup? could you describe the sequence, agent pods start async one by one and then ....

james-t-bloom:
> 

@dkhokhlov so as pods come online they check the lease - if no one is holding it they attempt to grab it. The update is atomic so only one pod will be able to grab the lease.

dkhokhlov:
what happen next? 
- pod that got lease
- pods w/o lease

james-t-bloom:
> Pods not elected leader will poll with a duration of the leaseDurationInSeconds. Or 5 minutes if that for whatever reason is not set.

The metrics reporter stream is set to be created, and then blocked until such time it becomes the leader.

Pods who get the lease start reporting on metrics.

dkhokhlov:
any other functionality that depends on "leader election"? k8s event?

james-t-bloom:
> any other functionality that depends on "leader election"? k8s events?

Not yet - currently only the usage statistics reporter functionality. 

I am working on a story to wrap k8 events, but won't put that up until this passed code review/testing I think. It'll be fairly small when it comes out.

dkhokhlov:
ic. what happens when leader pod exits right away after startup - expired ingestion key etc?
- for metrics
- for events

james-t-bloom:
> ic. what happens when leader pod exits right away after startup - expired ingestion key etc?
> 
> * for metrics
> * for events
> * log lines loss

That's a good question - 

I guess the first thing to point out is the decision on the product side was made lost metric-server data is not a big deal, and it is acceptable for there to be a gap in the usage data. As there is no point recovering it since it goes stale rather quickly. So this is not made to have zero gaps. 

if a pod spins up grabs the lease and dies - the lease will be stay claimed as far as kubernetes is concerned by that pod until such a time that the lease hold expires. Then another pod will be able to force take the lease. Another thing to keep in mind is leases don't automatically lose their holder even when expired, someone has to take the lease. 

Events - not currently applicable.

dkhokhlov:
>  Then another pod will be able to force take the lease.
what happens exactly?

james-t-bloom:
> 

the other pods check on the lease periodically, if the lease doesn't have a holder (it has never been claimed) or if acquired time + lease duration is in the past - a non leader lease will attempt to claim the lease. Update being atomic - if this operation is successful that pod will become the leader, and start reporting on metrics. If it failed to acquire the lease, even after detecting its expired then the pod will continue on as it was, periodically polling the lease to check for expiration.

dkhokhlov:
> acquired time + lease duration is in the past
"lease duration" stops when the leader disappeared? is not it "no lease state"?

james-t-bloom:
> 

I'm not sure I understand, but the way leases work in this context is they don't lose the leader, the leader still "holds" the lease its just their hold expired and others can take it away. 

This point is acquire time + lease duration.

k8s/leader-election-leases.yaml
```diff
@@ -0,0 +1,37 @@
...
+- kind: ServiceAccount
+  name: logdna-agent
+  namespace: logdna-agent
+roleRef:
+  kind: Role
+  name: logdna-agent-reporter-lease
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: coordination.k8s.io/v1
+kind: LeaseList
```
stsantilena, line 27: Is there a need to use LeaseList here? I looks like there's only one Lease that's needed for this feature?

docs/README.md
```diff
@@ -298,9 +298,11 @@ __Note:__ The default option is `never`.
 
 ### Configuring Metrics Reporter Statistics
 
-With this enabled the agent will pull from the kubernetes metrics-server, this allows the agent to report on CPU/Memory usage statistics for pods and nodes in the cluster. These statistics will be viewable on the web application for individual log lines showing usage for the pod and node associated with that log line. 
+With this enabled the agent will pull from the kubernetes metrics-server, this allows the agent to report on CPU/Memory usage statistics for pods and nodes in the cluster. These statistics will be viewable on the web application for individual log lines showing usage for the pod and node associated with that log line.
 
-Reach out to Mezmo to get this feature enabled on the web application in addition to this configuration. 
+Reach out to Mezmo to get this feature enabled on the web application in addition to this configuration.
+
+Only one pod in a cluster will report metrics statistics for the entire cluster - the leader election process leverages leases. please see the leader-election-leases.yaml file in the k8s folder for the lease specifications, permissions.
```
stsantilena, line 305: Perhaps a bit more detail around the leader algorithm?

Nit: Capital "P" in Please.

bin/src/_main.rs
```diff
@@ -139,7 +146,9 @@ pub async fn _main(
     let mut executor = Executor::new();
 
     let mut k8s_claimed_lease: Option<String> = None;
-    let (k8s_event_stream, metric_stats_source) = match create_k8s_client_default_from_env(
+    let mut metrics_stream_feature_meta: Option<FeatureLeaderMeta> = None;
+
+    let (k8s_event_stream, metric_stats_stream) = match create_k8s_client_default_from_env(
```
stsantilena, line 151: I'm probably missing something but, is this also being defined below as well?

bin/src/_main.rs
```diff
@@ -27,17 +29,22 @@ use middleware::meta_rules::{MetaRules, MetaRulesConfig};
...
 use tokio::signal::*;
 use tokio::sync::Mutex;
+use tokio::time::sleep;
 use tokio::time::Duration;
 
 #[cfg(feature = "dep_audit")]
 use crate::dep_audit;
 use crate::stream_adapter::{StrictOrLazyLineBuilder, StrictOrLazyLines};
 
+pub static REPORTER_LEASE_NAME: &str = "logdna-agent-reporter-lease";
```
stsantilena, line 45: Const?

bin/src/_main.rs
james-t-bloom: lets chat about this more - but one is source, one is stream.

docs/README.md
james-t-bloom: fixed

k8s/leader-election-leases.yaml
james-t-bloom: fixed

bin/src/_main.rs
james-t-bloom: fixed and moved down a bit - its by other vars though, let me know if you have a better spot.

bin/src/_main.rs
stsantilena: Ah, yeah I see what you mean, my mistake. This looks good 👌

c-nixon (approved):

PR #389, opened 2022-08-10
Set TOOLS_IMAGE_TAG in Jenkins environment

c-nixon:
This sets the TOOLS_IMAGE_TAG env var when running under Jenkins. We need to do this as the gcr image tags that Jenkins uses are different from the docker.io tags that are set by default in the Makefile.


dkhokhlov (approved):
thanks!

PR #392, opened 2022-08-16
Address new lint errors

c-nixon:
This addresses the new clippy lints from the next version of the rust toolchian

### DIFF
--- a/bin/src/_main.rs
+++ b/bin/src/_main.rs
@@ -607,11 +607,11 @@ async fn get_signal() -> &'static str {
     let mut quit_signal = unix::signal(unix::SignalKind::quit()).unwrap();
     let mut term_signal = unix::signal(unix::SignalKind::terminate()).unwrap();
 
-    return tokio::select! {
+    tokio::select! {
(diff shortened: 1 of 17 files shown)


bin/src/_main.rs
```diff
@@ -1,621 +1,621 @@
...
-    }
-}
-
-#[cfg(unix)]
-async fn get_signal() -> &'static str {
-    let mut interrupt_signal = unix::signal(unix::SignalKind::interrupt()).unwrap();
-    let mut quit_signal = unix::signal(unix::SignalKind::quit()).unwrap();
-    let mut term_signal = unix::signal(unix::SignalKind::terminate()).unwrap();
-
-    return tokio::select! {
```
stsantilena, line 610: Seems like a small change. Wondering why the large git diff?

stsantilena (approved):
LGTM.

bin/src/_main.rs
c-nixon: Looks like whitespace, the previous version of the file must have had windows line endings, when I changed it and re cargo-fmt'd it they much have been stripped

PR #401, opened 2022-08-18
Log 13962 retry

james-t-bloom:
Fixes retry logging file location on windows specifically
Test plan: Integration Test Unit Test Manual Testing via observing leader logs
Ref: LOG-13962

### DIFF
--- a/common/config/src/lib.rs
+++ b/common/config/src/lib.rs
@@ -319,8 +319,7 @@ impl TryFrom<RawConfig> for Config {
                 .ok_or(ConfigError::MissingField("http.body_size"))?,
             retry_dir: raw.http.retry_dir.unwrap_or_else(|| {
                 if cfg!(windows) {
-                    let temp = std::env::var("temp");
-                    if let Ok(temp) = temp {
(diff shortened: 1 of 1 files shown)


common/config/src/lib.rs
```diff
@@ -317,10 +317,16 @@ impl TryFrom<RawConfig> for Config {
...
-                .http
-                .retry_dir
-                .unwrap_or_else(|| PathBuf::from("/tmp/logdna")),
+            retry_dir: raw.http.retry_dir.unwrap_or_else(|| {
+                if cfg!(windows) {
+                    let temp = std::env::var("temp");
+                    if let Ok(temp) = temp {
+                        let dir = format!("{}/{}", temp, "logdna");
+                        return PathBuf::from(dir);
+                    }
```
dkhokhlov, line 326:
```
 if let Ok(temp) = std::env::var("temp") {
     let dir = format!("{}/{}", temp.into(), "logdna");
     return PathBuf::from(dir);
```
?

james-t-bloom: changed

stsantilena (approved):
LGTM.

PR #403, opened 2022-08-18
Log 13753 event leader refactor

james-t-bloom:
Surrounds k8 events with new leader election system, while removing the old “oldest pod” way of doing things.

Fixed a bug around failure case 

Test plan: Integration Test Unit Test Manual Testing via observing leader logs
Ref: LOG-13753


bin/src/_main.rs
```diff
@@ -148,134 +150,110 @@ pub async fn _main(
...
+                            executor.register(v);
+                            info!("Registered k8s metadata middleware");
+                        }
+                        Err(e) => {
+                            let message = format!(
+                                "The agent could not access k8s api after several attempts: {}",
+                                e
+                            );
+                            error!("{}", message);
+                            panic!("{}", message);
```
stsantilena, line 180: Why the need for the error and the panic?

k8s/event-leader-lease.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: logdna-agent
+roleRef:
+  kind: Role
+  name: logdna-agent-k8-events-lease
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: coordination.k8s.io/v1
+kind: Lease
+metadata:
+  name: logdna-agent-k8-events-lease
```
stsantilena, line 29: Looks like you've switched naming convention in the rest of the PR from `k8-events` -> `reporter`, should probably change here as well?

bin/src/_main.rs
james-t-bloom: It was already there just looks new because of some moving code. 

178 in the diff, that being said.... I don't really know...

bin/src/_main.rs
```diff
@@ -611,6 +585,38 @@ pub async fn _main(
...
+    );
+
+    let is_starting_leader = leader.try_claim_feature_leader().await;
+    let leader_lease = leader.get_feature_leader_lease().await;
+    let mut interval = DEFAULT_CHECK_FOR_LEADER_S;
+    if let Ok(lease) = leader_lease {
+        interval = lease
+            .spec
+            .map(|s| s.lease_duration_seconds.unwrap_or(0i32))
+            .unwrap_or(0i32)
```
dkhokhlov, line 611: this hardcoded "0" is the same as DEFAULT_CHECK_FOR_LEADER_S?

common/k8s/src/feature_leader.rs
```diff
@@ -8,7 +8,7 @@ use crate::lease::{self, get_lease};
 pub struct FeatureLeaderMeta {
     pub interval: i32,
     pub leader: FeatureLeader,
-    pub is_currently_leader: bool,
+    pub is_starting_leader: bool,
```
dkhokhlov, line 11: is_currently_leader vs is_starting_leader

is_starting_leader is not leader?

bin/src/_main.rs
james-t-bloom: it is not, good eye. Going to rework this a bit.

k8s/event-leader-lease.yaml
james-t-bloom: actually two seperate features, kind of confusing.

reporter is kind of reporting on metrics
events is the k8 eventrs ie create, delete, etc...

bin/src/_main.rs
james-t-bloom: added a check where if this comes in 0, A the lease is misconfigured but it'll default back to DEFAULT_CHECK_FOR_LEADER_S

bin/tests/it/k8s.rs
```diff
@@ -870,13 +870,18 @@ async fn create_agent_startup_lease_list(client: Client, name: &str, namespace:
         .unwrap();
 }
 
-async fn create_agent_feature_lease(client: Client, namespace: &str) {
+async fn create_agent_feature_lease(
+    client: Client,
+    namespace: &str,
+    lease_name: &str,
```
stsantilena, line 876: Nit: maybe be a little more specific in the naming here? I'm just thinking about the confusion between the startup leases and these leases.

james-t-bloom: as in the function name or the parameter lease_name?

stsantilena:
In regards to the manual testing has there been any regression tests to ensure that this process plays nicely with the startup lease process? There's a lot of logic changes in `_main.rs` that is in and around the existing startup lease logic.

james-t-bloom:
> 

Outside of the integration tests, and just testing my own stuff with all the changes. Not really, unless @jiangao1 has - I do know once this is buttoned up he has a few smoke tests planned on master/release branch.

james-t-bloom:
> In regards to the manual testing has there been any regression tests to ensure that this process plays nicely with the startup lease process? There's a lot of logic changes in `_main.rs` that is in and around the existing startup lease logic.

@stsantilena on top of that this is using the leader election logic that his been in master for a few weeks now. Nothing _really_ new has been added

stsantilena (approved):
This LGTM.

jiangao1 (approved):
Passed both reporting tests and starting lease regression tests. Will mix and match different features and smoke test more once it is merged into Master and have a master build.

PR #422, opened 2022-09-29
LOG-13046-agent-add-support-for-windows-event-logs

dkhokhlov:
- create new Tailer source
- spawn winevt-tailer as child process (https://github.com/logdna/winevt-tailer). 
- receive JSON log lines from the tailer stdout 
- send log lines to Mezmo using standard agent LineBuilder stream infra
- parameterize the tailer source in agent configuration file using new params:  log.tailer_cmd, log.tailer_args
**Testing:**
- check that  entries from Windows Event logs (System, Application) are correctly sent, parsed and presented in Mezmo Web UI front-end
**Notes:**
- early tailer errors and stack traces go into agent log. any non-recoverable tailer process error is fatal - causes agent to exit (and windows service to restart).
- packaging includes released winevt-tailer exe  into agent MSI alone with the tailer configuration appended to agent config file (yaml).
- install: previous existing conf is now gets saved with timestamp suffix in name, agent uses bundled config to be fully functional. 
- install: if ingestion key provided during MSI or Choco install then agent config get updated with it
- documentation is WIP

LOG-13046
LOG-14553
LOG-14644


gjanco (approved):

api/src/tailer/mod.rs
```diff
@@ -0,0 +1,347 @@
...
+pub struct TailerApiDecoder {
+    state: AnyPartialState,
+    _stdin: ChildStdin,
+    #[cfg(windows)]
+    _job: Job,
+}
+
+type TailerRecord = String;
+
+// The actual parser for the Tailer API format
```
james-t-bloom, line 41: I know we've had issues with encoding in the past, is this inline with current flows?

bin/src/_main.rs
```diff
@@ -326,6 +329,39 @@ pub async fn _main(
         false => None,
     };
 
+    let tailer_source = match (config.log.tailer_cmd, config.log.tailer_args) {
```
james-t-bloom, line 332: So I got kind of turned around with how this is set up - and I feel it requires too much domain knowledge. This could just be a misunderstanding on my part. 

Instead of this being wrapped around two configs that are set in a novel way that is new to the agent (by a brand new outside library) but if we wrapped it around if these tailers are defined for a given platform? This is all basically dead code for linux/macos but there is no indication as such. If possible I'm normally against over commenting but a comment around here explaining this I think would be welcome.

common/config/src/raw.rs
```diff
@@ -303,6 +308,10 @@ pub struct LogConfig {
     pub log_metric_server_stats: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub clear_cache_interval: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
james-t-bloom, line 311: The way this raw config comes in again I think id would be great to have a little documentation around this.

api/src/tailer/mod.rs
```diff
@@ -0,0 +1,347 @@
...
+fn decode_parser<'a, Input>(
+) -> impl Parser<Input, Output = TailerRecord, PartialState = AnyPartialState> + 'a
+where
+    Input: RangeStream<Token = u8, Range = &'a [u8]> + 'a,
+    Input::Error: ParseError<Input::Token, Input::Range, Input::Position>,
+{
+    /*
+    Tailer API log lines are 'etf-8' encoded and separated by a single newline.
```
james-t-bloom, line 49: When you say tailer API lines are encoded this way - but eventually customers will be able to write their own tailers - perhaps we should add some error handling around this as we can't guarantee how an outside source will format their data.

james-t-bloom (commented):
Mostly LGTM 

- most of the issues I had were around understanding what was going on but those have been addressed - 

my other ask I guess is this PR encompasses three seperate stories. I know they are related and two of them are really small but it did add a little confusion and overhead tracking down the three seperate stories.  These could be split up.

api/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "api"
+version = "0.1.0"
+authors = ["dkhokhlov <dkhokhlov@gmail.com>"]
```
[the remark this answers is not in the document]
dkhokhlov: That is my Github email.

api/src/tailer/error.rs
```diff
@@ -0,0 +1,23 @@
+use std::fmt::{self, Display, Formatter};
+
+#[allow(dead_code)] // TODO
```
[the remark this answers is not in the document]
dkhokhlov: just to start using it.

api/src/tailer/mod.rs
dkhokhlov: correct. this tailer api is using etf8 encoding, same as JSON payload.

bin/src/_main.rs
dkhokhlov: the source will be None if these params are not configured. meaning - the source is optional. by default it is not enabled on non-windows platforms (as windows installation package comes with bundled agent config file that has these options enabled).
It is standard approach in agent, see line 512.

api/src/tailer/mod.rs
dkhokhlov: We do lossy string ending in agent on data path to backend (through http etc). Meaning if encoding is wrong it will be mitigated.

bin/src/_main.rs
james-t-bloom: Gunna approve I guess my one ask would be a comment saying where these are set, but i'm not gunna hold up the PR for that.

james-t-bloom (approved):
lgtm one nit

PR #423, opened 2022-09-30
fix: timeout in AppVeyor CI build

dkhokhlov:
- removed release build - duplicates Jenkins
- left only tests


dkhokhlov:
> Just one small question: Why the empty file `raw.rs` showing up in the PR?

file flags fixed - removed x.

stsantilena (approved):

PR #436, opened 2022-10-27
feat(image-tag): Add image name and tag to meta object

james-t-bloom:
Summary:
Parsing out container image and tag from appropriate container, sending via _meta object

Test Plan:

Manually test on web application
Unit tests

Ref: Log-10828

### DIFF
--- a/common/k8s/src/middleware/metadata.rs
+++ b/common/k8s/src/middleware/metadata.rs
@@ -32,7 +32,7 @@ pub enum Error {
     K8s(#[from] kube::Error),
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct MetaObject {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename(serialize = "Image Name"))]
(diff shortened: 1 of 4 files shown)


stsantilena (approved):
This LGTM.

common/k8s/src/middleware/metadata.rs
```diff
@@ -112,6 +124,12 @@ impl Middleware for K8sMetadata {
                 let obj_ref =
                     ObjectRef::new(&parse_result.pod_name).within(&parse_result.pod_namespace);
                 if let Some(pod) = self.store.get(&obj_ref) {
+                    let meta_object =
+                        extract_image_name_and_tag(parse_result.container_name, pod.as_ref());
+                    if line.set_meta(json!(meta_object)).is_err() {
+                        return Status::Skip;
```
dkhokhlov, line 130: just skip?

james-t-bloom: following the pattern of how the rest of the fields are set.

dkhokhlov: how about logging the error? if meta got lost then the line that gets sent is invalid.

james-t-bloom: I would normally agree but the error that is returned from this is actually nothing. It would be just as helpful to not have a log for it. 

However - in the spirit of this i am going to add 

> This looks great, can we get an integration test for it? Or maybe just add the check to one of the existing kubernetes ones?

added

james-t-bloom: @dkhokhlov added some more logic around your suggestion.

PR #439, opened 2022-10-29
Log 14607 windows add rotation by size for agent service log file

dkhokhlov:
windows: added file rotation by size against log flood


stsantilena (approved):
This LGTM.

PR #440, opened 2022-10-31
Tweak ci so that cargo vendor is honoured by the release builds

c-nixon:
This changes the Makefile to use the newish `--offline` flag for cargo to force it to use the vendored deps and then overrides all of the build step's target directories to cut down on contention for the target tree locks.

It also reduces how clean the environment the release builds run in as we don't blow away the registry etc, so we're less sandboxed from a misbehaving build.rs...


dkhokhlov (approved):
lgtm

PR #442, opened 2022-11-04
Windows File Handle Leak

stsantilena:
This updates the OS Watcher so that all the proper events are sent from the OS. It also specifies a number of tests as being for unix only. These tests are very event specific. 

LOG-14882

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -867,6 +867,12 @@ impl FileSystem {
         if !self.passes(path, _entries) {
             // Do not continuously log ignored files
             if let false = self.ignored_dirs.contains(path) {
+                #[cfg(windows)]
+                self.ignored_dirs.insert(path.to_path_buf());
+                info!("ignoring {:?}", path);
+                _entries.clear();
+
+                #[cfg(unix)]
                 self.ignored_dirs.insert(path.to_path_buf());
                 info!("ignoring {:?}", path);
             }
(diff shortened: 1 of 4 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -1833,6 +1834,7 @@ mod tests {
 
     // Deletes a file
     #[tokio::test]
+    #[cfg(unix)]
     async fn filesystem_delete_file() -> io::Result<()> {
```
dkhokhlov, line 1838: this is basic test to verify on windows for this fix. shall pass.

common/notify_stream/src/lib.rs
```diff
@@ -247,6 +247,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(unix)]
     async fn test_unwatch_if_exists() {
```
dkhokhlov, line 251: this is basic test to verify on windows for this fix. shall pass.

dkhokhlov (changes requested):

common/notify_stream/src/lib.rs
```diff
@@ -317,6 +318,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(unix)]
     async fn test_create_write_delete() -> io::Result<()> {
```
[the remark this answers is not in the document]
stsantilena: Added additional, Windows specific test for this.

stsantilena:
The fixes for this were included in #453. Closing this out.

PR #443, opened 2022-11-07
mock ingester

dkhokhlov:
pip install flask
python mock_ingester.py

change these params in agent config:
  host: localhost
  use_ssl: false


jiangao1 (approved):
Nice little script. So. I can direct ingestion url to 0.0.0.0:8080/logs/agent to see if certain log is sent. Looks good. I'll try it out soon.

PR #444, opened 2022-11-08
disable failing tests on Windows to unblock CI builds

dkhokhlov:
Tracked by LOG-15013

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -1700,6 +1700,7 @@ mod tests {
     }
 
     /// Creates a symlink directory
+    #[cfg(unix)]
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -1700,6 +1700,7 @@ mod tests {
     }
 
     /// Creates a symlink directory
+    #[cfg(unix)]
```
stsantilena, line 1703: Nit: put unix tag underneath test tag for consistency.

stsantilena (approved):

common/fs/src/cache/mod.rs
dkhokhlov: this is temp, all tests must be fixed and all these ifdefs removed.

PR #445, opened 2022-11-09
Fixup Makefile TargetDir logic and new lints for rust 1.65

c-nixon:
This fixes up some lines that the TargetDir change caused and also applies `cargo clippy --fix`


dkhokhlov (approved):
lgtm

PR #446, opened 2022-11-09
tweak resolve_valid_paths not to return directories

c-nixon:
This adds some additional logic to the fs cache in order to make sure the resolve_valid_paths function never returns a directory. This fix exposed a bug in the FS initialisation logic which caused files not to be added

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -395,7 +395,7 @@ impl FileSystem {
                 }
             }
 
-            if !(path_cpy.components().count() > 1) {
+            if path_cpy.components().count() <= 1 {
                 debug!("Not recursively walking paths under {:?}", path_cpy);
                 continue;
             }


common/fs/src/cache/mod.rs
```diff
@@ -390,17 +390,35 @@ impl FileSystem {
...
                     path.ok().and_then(|path| {
                         fs.is_initial_dir_target(path.path())
                             .then(|| path.path().to_path_buf())
                     })
                 })
-                .collect::<Vec<_>>()
-                .iter()
-            {
+                .filter_map(|path: PathBuf| {
+                    fs::read_dir(&path).ok().map(|entries| {
```
stsantilena, line 403: Is `entries` referring to filename, or SlopMap `entries`?

stsantilena (approved):
One small question. LGTM! 🚢

PR #447, opened 2022-11-10
Log 14502 missing lines

james-t-bloom:
bug(lookback-tail):  Add new loopback option to address missing line bug

Summary: If a .log file is created after the agent begins running, that file is read from the beginning with new loopback option 

Test plan:

Manually testing on web app
unit tests

Ref: LOG-14502


common/fs/src/cache/mod.rs
```diff
@@ -799,6 +804,30 @@ impl FileSystem {
                 .metadata()
                 .map(|m| [Span::new(0, m.len()).unwrap()].iter().collect())
                 .unwrap_or_default(),
+            Lookback::Tail => {
+                let mut should_lookback = false;
+                let file_len = path.metadata().map(|m| m.len()).unwrap_or(0);
+
+                if file_len > TAIL_WARN_THRESHOLD_B {
+                    log::warn!("lookback ocurred on larger file {:?}", path);
```
dkhokhlov, line 812: also print the threshold?

common/fs/src/cache/mod.rs
```diff
@@ -299,6 +301,8 @@ impl FileSystem {
             }
         });
 
+        let fs_start_time = SystemTime::now();
```
dkhokhlov, line 304: not monotonic

docs/README.md
```diff
@@ -261,6 +261,9 @@ The valid values for this option are:
    * When set to **`start`**:
       * If there is information in the “state file”, use the last recorded state. 
       * If the file is not present in the “state file”, start at the beginning. 
+   * When set to **`tail`**:
+      * Will read .log files starting at the beginning if created after the agent has started. This can result in large .log files being processed if copied into log directory. A WARN log will be printed if this happens.  
+
```
dkhokhlov, line 266: what is the default? is it fixed in default case?

dkhokhlov (changes requested):

docs/README.md
james-t-bloom: the default case will still behave the same (none) 

this was the change that was discussed to address the bug.

common/fs/src/cache/mod.rs
james-t-bloom: i'm not sure what you're getting at - but this is just the system time that is captured to compare when a file was created for tail

common/fs/src/cache/mod.rs
```diff
@@ -49,6 +49,7 @@ type FsResult<T> = Result<T, Error>;
 pub const EVENT_STREAM_BUFFER_COUNT: usize = 1000;
 const EVENT_RETRY_DELAY_MS: u64 = 1000;
 const EVENT_RETRY_COUNT: u32 = 5;
+const TAIL_WARN_THRESHOLD_B: u64 = 3000000;
```
stsantilena, line 52: Two questions: One, was this value chosen arbitrarily? Two, what does the `_B` represent? Bytes?

common/fs/src/cache/mod.rs
james-t-bloom: i'm not sure it's necessary - as it's not configurable - could just be extra noise - and is just a heads up to the user that we are doing a lookback on a not insignificant file.

common/fs/src/cache/mod.rs
james-t-bloom: 1 - basically, it doesn't affect functionality in anyway. It's just if we warn the user that large files are being lookback'd on
2 - yes

docs/README.md
```diff
@@ -261,6 +261,9 @@ The valid values for this option are:
    * When set to **`start`**:
       * If there is information in the “state file”, use the last recorded state. 
       * If the file is not present in the “state file”, start at the beginning. 
+   * When set to **`tail`**:
+      * Will read .log files starting at the beginning if created after the agent has started. This can result in large .log files being processed if copied into log directory. A WARN log will be printed if this happens.
```
stsantilena, line 265: Maybe reference the default behavior to refresh the reader's mind. It starts from the beginning of the file, but as opposed to what? Just a suggestion.

james-t-bloom: 👍

stsantilena (approved):
Couple minor comments. Other than than, LGTM. :shipit:

PR #448, opened 2022-11-10
Fix Windows Compiler Warnings

stsantilena:
These two functions were added for the missing initial directory bug (#385). This fix currently only applies to Linux. As a result, when compiling on Windows, the compiler throws a warning about the functions not being used.

### DIFF
--- a/common/fs/src/cache/dir_path.rs
+++ b/common/fs/src/cache/dir_path.rs
@@ -147,6 +147,7 @@ mod tests {
     use super::*;
 
     #[test]
+    #[cfg(unix)]
     fn test_level_up() {
         let mut expe_pathbuf = PathBuf::new();
         expe_pathbuf.push("/test-directory");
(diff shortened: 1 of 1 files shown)


dkhokhlov (approved):
lgtm

common/fs/src/cache/dir_path.rs
```diff
@@ -89,6 +89,7 @@ impl From<DirPathBuf> for PathBuf {
 ///
 /// When a valid directory is found, it adds the missing piece of the directory
 /// path to `postfix` and returns the `DirPathBuf` result.
+#[cfg(unix)]
```
dkhokhlov, line 92: @stsantilena 
i missed it - we want this feature (non-existing directories) tested on windows too.
possible?

dkhokhlov (changes requested):

common/fs/src/cache/dir_path.rs
stsantilena: Not at this time. The fix for missing directory only applies to Linux at the moment. There is a ticket to extend to Windows.

dkhokhlov (approved):
lgtm

PR #452, opened 2022-11-30
Cache and set target dir for integration tests

c-nixon:
This adds a timeout to the cmd_line_env integration tests, makes them use the global agent binary cache and separates the integration test agent builds by their feature sets

### DIFF
--- a/bin/tests/it/cmd_line_env.rs
+++ b/bin/tests/it/cmd_line_env.rs
@@ -850,7 +850,7 @@ where
         panic!("timed out waiting for agent to start: {}", *data);
     }
 
-    log::debug!("{}", data);
+    log::debug!("Agent STDERR output:\n{}", data);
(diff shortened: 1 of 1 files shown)


bin/tests/it/cmd_line_env.rs
```diff
@@ -806,39 +808,53 @@ where
...
+    handle.kill().unwrap();
+    handle.wait().unwrap();
+
+    // Kill the agent
+    if result.timed_out() || *pending {
+        panic!("timed out waiting for agent to start: {}", *data);
     }
 
-    let data = data.lock().unwrap();
+    log::debug!("{}", data);
```
stsantilena, line 853: Maybe add a prefix on the debug to make it more clear what data this is?

stsantilena (approved):
LGTM.

PR #456, opened 2022-12-14
feature: fs cache refresh

dkhokhlov:
- added config.log.clear_cache_interval, default is 6 hours
- fs tailer gets restarted periodically using clear_cache_interval to clear fs cache 

LOG_15140

### DIFF
--- a/common/config/src/raw.rs
+++ b/common/config/src/raw.rs
@@ -300,6 +300,8 @@ pub struct LogConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub k8s_metadata_exclude: Option<Vec<String>>,
     pub log_metric_server_stats: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub clear_cache_interval: Option<u32>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Clone, Default)]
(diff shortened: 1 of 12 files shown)


common/config/src/raw.rs
```diff
@@ -445,6 +447,7 @@ impl Default for LogConfig {
             k8s_metadata_include: None,
             k8s_metadata_exclude: None,
             log_metric_server_stats: None,
+            clear_cache_interval: Some(3600 * 3), // 3 hours
```
stsantilena, line 450: Was 3 hours chosen at random or is there a significant build up over that period?

dkhokhlov: build up is unconstrained. depends on file event rate. by consensus.

common/fs/src/tail.rs
```diff
@@ -319,6 +324,20 @@ where
...
+            if let Some(start_time) = this.start_time.as_mut().as_pin_mut() {
+                if this.restart_interval.as_secs() > 0
+                    && start_time.elapsed().as_secs() > this.restart_interval.as_secs()
+                {
+                    this.start_time.set(Some(Instant::now()));
+                    let stream_fut = (this.f)(this.params);
+                    this.pending.set(Some(stream_fut));
+                    info!(
+                        "restarting stream, interval={}",
+                        this.restart_interval.as_secs()
```
stsantilena, line 337: Is that just resetting the interval which resets the loop?

stsantilena (approved):
Just had a couple small questions, other than that, this LGTM.

dkhokhlov:
rebased to master, changed interval default to 6 hours.

PR #457, opened 2022-12-15
Log 11792 journal toggle

james-t-bloom:
bug(journal_toggle):  Wrap task spawning in more conditions including new env variable 

Summary: added new environment variable MZ_SYSTEMD_JOURNAL_TAILER
If user explicitly sets that to false, no matter if there is a journal path no tasks will spawn

Test plan:

Manually testing 
unit tests 
Int tests

Ref: LOG-11792

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,20 +65,6 @@ pipeline {
         }
         stage('Lint and Unit Test') {
             parallel {
-                stage('Lint'){
-                    steps {
-                        withCredentials([[
-                                           $class: 'AmazonWebServicesCredentialsBinding',
-                                           credentialsId: 'aws',
-                                           accessKeyVariable: 'AWS_ACCESS_KEY_ID',
-                                           secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'
-                                         ]]){
-                            sh """
-                              make lint
-                            """
-                        }
-                    }
-                }
                 stage('Unit Tests'){
                     steps {
                         withCredentials([[
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,6 +65,20 @@ pipeline {
         }
         stage('Lint and Unit Test') {
             parallel {
+                stage('Lint'){
+                    steps {
+                        withCredentials([[
+                                           $class: 'AmazonWebServicesCredentialsBinding',
+                                           credentialsId: 'aws',
+                                           accessKeyVariable: 'AWS_ACCESS_KEY_ID',
+                                           secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'
+                                         ]]){
+                            sh """
+                              make lint
+                            """
+                        }
+                    }
+                }
                 stage('Unit Tests'){
                     steps {
                         withCredentials([[
--- a/bin/src/_main.rs
+++ b/bin/src/_main.rs
@@ -282,7 +282,7 @@ pub async fn _main(
     executor.init();
 
     // Use an internal env var to support running integration test w/o additional delays
-    let event_delay = std::env::var(config::env_vars::INTERNAL_FS_DELAY)
+    let _event_delay = std::env::var(config::env_vars::INTERNAL_FS_DELAY)
         .map(|s| Duration::from_millis(s.parse().unwrap()))
         .unwrap_or(FS_EVENT_DELAY);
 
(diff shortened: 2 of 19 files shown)


docs/README.md
```diff
@@ -294,7 +295,7 @@ file in docs directory of this repository.
 
 ### Configuring Journald
 
-If the agent pods have access to journald log files or directories, monitoring can be enabled on them with the `LOGDNA_JOURNALD_PATHS`. Common values include `/var/log/journal` and `/run/systemd/journal`. To specify both, use a comma separated list: `/var/log/journal,/run/systemd/journal`.
+If the agent pods have access to journald log files or directories, monitoring can be enabled on them with the `LOGDNA_JOURNALD_PATHS` as well as setting `MZ_SYSTEMD_JOURNAL_TAILER` to true, or leaving blank which defaults to true. Common values include `/var/log/journal` and `/run/systemd/journal`. To specify both, use a comma separated list: `/var/log/journal,/run/systemd/journal`.
```
dkhokhlov, line 298: How many vars need to be defined to enable journal log? 

How many vars need to be defined to have journal log disabled?

dkhokhlov: table needs to be updated:
https://github.com/logdna/logdna-agent-v2/blob/master/docs/README.md#configuration

james-t-bloom: updated a bit

Jenkinsfile
```diff
@@ -65,20 +65,6 @@ pipeline {
         }
         stage('Lint and Unit Test') {
             parallel {
-                stage('Lint'){
```
dkhokhlov, line 68: why removed?

bin/src/_main.rs
```diff
@@ -286,29 +286,44 @@ pub async fn _main(
...
+                        info!("Journalctl source was not initialized");
+                        debug!("Journalctl source initialization error: {}", e);
+                    })
+                    .ok()
+            } else {
+                None
+            }
+        }
+        false => None,
     };
```
dkhokhlov, line 317: How do we know journlad is started/initialized?

james-t-bloom: there are two logs in the code one for journalCTL and for ID a little deeper.

common/config/src/properties.rs
```diff
@@ -256,6 +257,11 @@ fn from_property_map(map: HashMap<String, String>) -> Result<Config, ConfigError
             .for_each(|v| paths.push(PathBuf::from(v)));
     }
 
+    if let Some(value) = map.get(&SYSTEMD_JOURNAL_TAILER) {
+        log::error!("HEYYY {}", value);
```
dkhokhlov, line 261: Error message?

dkhokhlov (changes requested):

common/config/src/properties.rs
james-t-bloom: lol 🙄

bin/tests/it/cli.rs
```diff
@@ -610,8 +610,9 @@ async fn test_z_journald_support() {
     settings.journald_dirs = Some(dir);
     settings.features = Some("libjournald");
     settings.exclusion_regex = Some(r"^(?!/var/log/journal).*$");
-    assert_eq!(systemd::journal::print(6, "Sample info"), 0);
+    settings.log_journal_d = Some("true");
```
dkhokhlov, line 613: everywhere:
log_journal_d -> log_journald_enabled or log_journald
?

docs/README.md
```diff
@@ -294,7 +295,7 @@ file in docs directory of this repository.
 
 ### Configuring Journald
 
-If the agent pods have access to journald log files or directories, monitoring can be enabled on them with the `LOGDNA_JOURNALD_PATHS`. Common values include `/var/log/journal` and `/run/systemd/journal`. To specify both, use a comma separated list: `/var/log/journal,/run/systemd/journal`.
+If the agent pods have access to journald log files or directories, monitoring can be enabled on them with the `LOGDNA_JOURNALD_PATHS` as well as setting `MZ_SYSTEMD_JOURNAL_TAILER` to true, or leaving blank which defaults to true. Both variables are required to be configured. Common values include for `LOGDNA_JOURNALD_PATHS` - `/var/log/journal` and `/run/systemd/journal`. To specify both, use a comma separated list: `/var/log/journal,/run/systemd/journal`.
```
stsantilena, line 298: Nit: the wording is a bit confusing. Do they both default to true? So, to enable, you need to either set them BOTH to to true, or leave them BOTH blank?

bin/src/_main.rs
```diff
@@ -286,29 +286,44 @@ pub async fn _main(
...
+            if !config.journald.paths.is_empty() {
+                Some(create_source(&config.journald.paths).map(StrictOrLazyLineBuilder::Strict))
+            } else {
+                None
+            }
+        }
+        false => None,
+    };
+
+    #[cfg(all(feature = "libjournald", target_os = "linux"))]
```
stsantilena, line 301: Nit: These annotations can get confusing, especially when the naming within this section so is similar. Perhaps add a small comment just to clarify?

stsantilena (approved):
Just a couple minor things. LGTM.

PR #459, opened 2022-12-21
Fix Shellcheck Lint Error

stsantilena:
Adds specific ignore attributes for `SC2317` and updates our hadolint/hadolint image.


dkhokhlov (approved):
lgtm

PR #462, opened 2023-01-06
Log 15367 reopen missing files

c-nixon:
This is cut down version of the logic to reopen a Tailed File if we receive a write event for a file that has been deleted that cleanly backports to other branches


james-t-bloom (approved):
lgtm

stsantilena (approved):

PR #467, opened 2023-01-18
Add tracing-rs support

stsantilena:
This removes the `log` logging crate and instead uses`tracing-rs`. The goal is to have log parity using the `tracing-rs` framework instead of the `log` framework.

TODO:
- [x]  Migrate from `log` to `tracing-rs`
- [x] Add `RUST_LOG` capability


bin/tests/it/cli.rs
```diff
@@ -11,14 +11,16 @@ use crate::common::{self, consume_output, AgentSettings};
...
 use logdna_mock_ingester::FileInfo;
 use predicates::prelude::*;
 use proptest::prelude::*;
 use tempfile::tempdir;
 use test_types::strategies::random_line_string_vec;
 use tokio::io::BufWriter;
 use tokio::task;
+use tracing::{debug, info};
+
+use test_log::test;
```
dkhokhlov, line 23: separate logging framework for tests?

PR #468, opened 2023-01-20
fix: panic at handling Rename notify-rs event with empty path

dkhokhlov:
- fixed panic during handling Rename-From notify-rs event with empty path
- it happens when From event is generated separately from To event due to some racing in notify-rs
- the event is ignored and rectified by explicit  FS rescan.

### DIFF
--- a/common/notify_stream/src/lib.rs
+++ b/common/notify_stream/src/lib.rs
@@ -176,9 +176,10 @@ impl Watcher {
                     EventKind::Modify(ModifyKind::Data(DataChange::Any)) => {
                         Some(Event::Write(received.paths.first().unwrap().clone()))
                     }
-                    EventKind::Modify(ModifyKind::Name(RenameMode::From)) => {
-                        Some(Event::Remove(received.paths.first().unwrap().clone()))
-                    }
+                    EventKind::Modify(ModifyKind::Name(RenameMode::From)) => received
+                        .paths
+                        .first()
+                        .map(|path| Event::Remove(path.clone())),
                     EventKind::Modify(ModifyKind::Name(RenameMode::To)) => {
                         Some(Event::Create(received.paths.first().unwrap().clone()))
                     }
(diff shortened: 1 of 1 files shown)


common/notify_stream/src/lib.rs
```diff
@@ -90,27 +92,29 @@ pub enum RecursiveMode {
 pub struct Watcher {
     watcher: OsWatcher,
     rx: Rc<async_channel::Receiver<Result<notify::Event, notify::Error>>>,
+    tx: Rc<async_channel::Sender<Result<notify::Event, notify::Error>>>,
```
stsantilena, line 95: Was the purpose for adding Sender to the Watcher struct for testing purposes?

common/notify_stream/src/lib.rs
```diff
@@ -176,9 +184,14 @@ impl Watcher {
                     EventKind::Modify(ModifyKind::Data(DataChange::Any)) => {
                         Some(Event::Write(received.paths.first().unwrap().clone()))
                     }
-                    EventKind::Modify(ModifyKind::Name(RenameMode::From)) => {
-                        Some(Event::Remove(received.paths.first().unwrap().clone()))
-                    }
+                    EventKind::Modify(ModifyKind::Name(RenameMode::From)) => received
```
stsantilena, line 187: Do you think there's value in adding this type of catch to the other EventKinds? Or, is this type of thing just specific to the notify-rs behavior on this type of event?

dkhokhlov: It is abnormal behavior of notify-rs. This workaround is about "Rename From" event only. It is part of event pair: "From" + "To". Bad case happens when notify-rs did not find tracked path in its own map at the time it composes "From" event. We should not ignore empty path cases in other events.

common/notify_stream/src/lib.rs
dkhokhlov: for testing.

common/notify_stream/src/lib.rs
dkhokhlov: implemented

PR #470, opened 2023-01-25
feat(macos): Build for macOS [LOG-13025]

james-t-bloom:
feat(macos): Build for macOS

Summary: Ended up being mostly cargo make, some cargo.toml changes and
a lot of test ignores
Makes it so agent builds on Mac with a test suite

Test plan:

Manually testing
unit tests
Int tests

Ref: LOG-13025


dkhokhlov:
very big PR - is it possible to split it?

james-t-bloom:
> very big PR - is it possible to split it?

Not really - but it's not as big as it seems. Mostly ignoring tests with cfg - and the ignoring has a cascading effect for lint reasons.

common/fs/src/cache/mod.rs
```diff
@@ -1896,7 +1901,7 @@ mod tests {
 
     // Deletes a directory
     #[tokio::test]
-    #[cfg(unix)]
+    #[cfg(target_os = "linux")]
     async fn filesystem_delete_filled_dir() -> io::Result<()> {
         let tempdir = TempDir::new()?;
```
dkhokhlov, line 1906: why disabled on mac?

common/notify_stream/src/lib.rs
```diff
@@ -191,6 +191,9 @@ impl Watcher {
                     EventKind::Modify(ModifyKind::Data(DataChange::Any)) => {
                         Some(Event::Write(received.paths.first().unwrap().clone()))
                     }
+                    EventKind::Modify(ModifyKind::Data(DataChange::Content)) => {
+                        Some(Event::Write(received.paths.first().unwrap().clone()))
```
dkhokhlov, line 195: did we miss it from the beginning?

common/notify_stream/src/lib.rs
```diff
@@ -586,50 +593,6 @@ mod tests {
         Ok(())
     }
 
-    /// macOS will follow symlink files
-    #[tokio::test]
-    #[cfg(target_os = "macos")]
-    async fn test_watch_symlink_write_after_create_macos() -> io::Result<()> {
```
dkhokhlov, line 592: removed for mac?

bin/src/main.rs
```diff
@@ -1,13 +1,9 @@
...
 use std::sync::Arc;
 use tokio::sync::Mutex;
-
-#[cfg(unix)]
-use tracing::{debug, info, trace, warn};
-
-#[cfg(windows)]
 use tracing::info;
-
+#[cfg(any(target_os = "windows", target_os = "linux"))]
```
dkhokhlov, line 5: cfg(not(target_os = "macos")) ?

common/notify_stream/src/lib.rs
james-t-bloom: no - as far as I can tell it's a mac thing.

common/fs/src/cache/mod.rs
james-t-bloom: There was kind of a cut bait and run strategy on some of these tests - Timing and directory creation issues that were found to be on the test side not code side.

common/notify_stream/src/lib.rs
james-t-bloom: I don't know where this test came from - but it didn't work on macos

dkhokhlov: Lets track all Mac problems in JIRA (create separate tickets). Disabled tests, symptoms, err cause if any.

james-t-bloom: yup will do

bin/tests/it/cmd_line_env.rs
```diff
@@ -10,6 +10,8 @@ use std::path::Path;
 use std::process::{Command, Output, Stdio};
 use std::str::from_utf8;
 use std::sync::{Arc, Condvar, Mutex};
+#[cfg(any(target_os = "windows", target_os = "linux"))]
```
stsantilena, line 13: Is there a "not mac" option?

Makefile.toml
```diff
@@ -0,0 +1,40 @@
+[env]
```
stsantilena, line 1: Can you explain a bit why the need for the this new `Makefile.toml` file? How is it different from a traditional Makefile?

stsantilena (approved):
Just a couple small questions. Other than that, LGTM.

bin/tests/it/cmd_line_env.rs
james-t-bloom: I honestly just did what the linter suggested, but probably would have been the way to go

Makefile.toml
james-t-bloom: so its a part of this cargo make plug in, it kind of augments the make file. Use it mostly for convenience cargo make run, cargo make int-tests etc

PR #473, opened 2023-02-02
Fix Trace Crate Windows Compiler Warnings

stsantilena:
Some of that paths in Windows didn't include some of the crate modules and were throwing warnings.


dkhokhlov (approved):
lgtm

PR #480, opened 2023-02-10
Add Execution Context to Errors

stsantilena:
This adds an error wrapper around our existing error type in the `notify-stream` library. It uses `std::backtrace` to capture the backtrace at the time of the error and returns it with the error:

Testing:
* Have a unit test that covers the capture.
* Windows compiler is throwing an E0658 error saying that backtrace is unstable. Also, looking through the rust git history, it looks as thought it's [stable](https://github.com/rust-lang/rust/pull/99573). I don't see this error in Linux or in the Jenkins pipeline.


common/notify_stream/src/lib.rs
```diff
@@ -83,6 +84,22 @@ pub enum Error {
...
+}
+
+impl ContextError {
+    pub fn new(error: Error) -> Self {
+        Self {
+            error,
+            context: Backtrace::capture(),
+        }
+    }
+}
```
stsantilena, line 101: Had to create custom error type and constructor to pass the backtrace information along.

PR #481, opened 2023-02-11
fix: non-existing log directory crash at startup on windows

dkhokhlov:
- agent is not able to start with logging directory that is missing on windows  - LOG-9806 regression on Windows
- typos

Ref: LOG-14549
Ref: LOG-9806

### DIFF
--- a/common/fs/src/cache/dir_path.rs
+++ b/common/fs/src/cache/dir_path.rs
@@ -1,9 +1,17 @@
+use regex::Regex;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use thiserror::Error;
-use tracing::error;
+use tracing::{error, warn};
+
 #[cfg(not(windows))]
-use tracing::warn;
+lazy_static! {
+    static ref REGEX_FS_ROOT: Regex = Regex::new(r"^/$").unwrap();
+}
+#[cfg(windows)]
+lazy_static! {
+    static ref REGEX_FS_ROOT: Regex = Regex::new(r"^ [a-zA-Z]:[\\/]$").unwrap();
+}
 
 #[derive(Error, std::fmt::Debug)]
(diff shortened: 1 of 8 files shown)


common/fs/src/cache/dir_path.rs
```diff
@@ -1,9 +1,17 @@
...
+use tracing::{error, warn};
+
 #[cfg(not(windows))]
-use tracing::warn;
+lazy_static! {
+    static ref REGEX_FS_ROOT: Regex = Regex::new(r"^/$").unwrap();
+}
+#[cfg(windows)]
+lazy_static! {
+    static ref REGEX_FS_ROOT: Regex = Regex::new(r"^ [a-zA-Z]:[\\/]$").unwrap();
```
stsantilena, line 13: Just curious, would a root path for Windows ever contain a lowercase char?

dkhokhlov: yes

common/fs/src/cache/dir_path.rs
```diff
@@ -88,50 +78,50 @@ impl From<DirPathBuf> for PathBuf {
...
                 }
-                let parent = match level_up(&p) {
-                    Some(p) => {
-                        if p == root_pathbuf {
+                let parent = match level_up(&some_path) {
+                    Some(some_parent) => {
+                        if REGEX_FS_ROOT
+                            .is_match(some_parent.to_str().expect("invalid unicode path").as_ref())
+                        {
                             warn!("root level directory was missing, as a result configured directory recursed to the root level!");
```
stsantilena, line 108: Wondering if this could be re-worded? The idea is that the configured directory doesn't even exist at the root level and therefore the Agent recursed all the way back to root.

stsantilena (approved):
This LGTM. Nice work!

PR #487, opened 2023-03-09
ci: publish x86 darwin build to correct s3 path

jakedipity:
We were publishing the x86 binary to the aarch64-darwin-apple path when it should have been x86_64-darwin-apple path.

Ref: LOG-13026


dkhokhlov (approved):
lgtm

PR #488, opened 2023-03-10
chore: add more logging to k8s watcher stream

jakedipity:
Adds logging that captures the name, namespace, labels, and annotations of pods that are captured by the k8s watcher stream. This is to help debug why some people are experiencing an edge case where the agent stops picking up labels.

Ref: LOG-16194

### DIFF
--- a/common/k8s/src/middleware/metadata.rs
+++ b/common/k8s/src/middleware/metadata.rs
@@ -14,6 +14,7 @@ use kube::{
     Api, Client,
 };
 use serde_json::json;
+use std::fmt;
 use std::pin::Pin;
 
 use backoff::ExponentialBackoff;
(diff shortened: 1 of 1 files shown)


common/k8s/src/middleware/metadata.rs
```diff
@@ -100,13 +100,41 @@ impl K8sMetadata {
                 let obj_ref = ObjectRef::from_obj(&pod);
                 if reader.get(&obj_ref).is_none() {
                     Metrics::k8s().increment_creates();
+                    info!(
```
dkhokhlov, line 103: info might be too chatty. agent sends these logs to backend too. affects the bill.

jakedipity: I modified the code to track generic pod changes with the info level, and more detailed annotation/label information with the trace level.

I thought the metadata cache tracking might be good to track in general. Let me know if you think I should just make it all trace.

jakedipity:
@dkhokhlov Updated the code to (hopefully) be more rustic? rusty?. Removed duplicate code and considered the organization chat we had about being more smart with types.

common/k8s/src/middleware/metadata.rs
```diff
@@ -191,6 +215,25 @@ fn extract_image_name_and_tag(container_name: String, pod: &Pod) -> Option<MetaO
     None
 }
 
+fn log_watcher_pod(log_event: LogEvent, pod: &Pod) -> () {
+    info!(
```
dkhokhlov, line 219: when you ran it - this "info" did not print many messages?

jakedipity:
@dkhokhlov I was able to fix the linting and integration test issues in this PR which is on main now: https://github.com/logdna/logdna-agent-v2/pull/487

I'm going to remove your lint commit and rebase onto master.

Also I decided to change all logs to `trace`, I didn't see much noise on my cluster, but that's with only a single pod... probably better to err on the side of caution.

PR #491, opened 2023-03-29
fix: refresh kubernetes metadata watcher on stream error

jakedipity:
In some cases, for unknown reasons, the k8s metadata watcher stream can
become inactive after receiving an error. This causes the cache to
become stale over time and for labels and annotations to be stale or
completely missing.

This PR fixes the issue by taking direct control of error handling on
stream errors and recreates the metadata watcher to start fresh.

Ref: LOG-16194

### DIFF
--- a/bin/src/_main.rs
+++ b/bin/src/_main.rs
@@ -24,7 +24,7 @@ use k8s::event_source::K8sEventStream;
 use k8s::lease::{get_available_lease, K8S_STARTUP_LEASE_LABEL, K8S_STARTUP_LEASE_RETRY_ATTEMPTS};
 
 use k8s::create_k8s_client_default_from_env;
-use k8s::middleware::K8sMetadata;
+use k8s::middleware::metadata_runner;
 use kube::Client as Kube_Client;
 use metrics::Metrics;
 use middleware::k8s_line_rules::K8sLineFilter;
@@ -154,21 +154,7 @@ pub async fn _main(
                     && std::env::var_os("KUBERNETES_SERVICE_HOST").is_some()
                 {
                     let node_name = std::env::var("NODE_NAME").ok();
-                    match K8sMetadata::new(k8s_client.clone(), node_name.as_deref()).await {
-                        Ok((driver, v)) => {
-                            tokio::spawn(driver);
-                            executor.register(v);
-                            info!("Registered k8s metadata middleware");
-                        }
-                        Err(e) => {
-                            let message = format!(
-                                "The agent could not access k8s api after several attempts: {}",
-                                e
-                            );
-                            error!("{}", message);
-                            panic!("{}", message);
-                        }
-                    };
+                    metadata_runner(user_agent, node_name, &mut executor);
                 }
 
                 let pod_name = std::env::var("POD_NAME").ok();
--- a/bin/tests/it/k8s.rs
+++ b/bin/tests/it/k8s.rs
@@ -11,7 +11,7 @@ use k8s_openapi::api::coordination::v1::Lease;
 use k8s_openapi::api::core::v1::{Endpoints, Namespace, Pod, Service, ServiceAccount};
 use k8s_openapi::api::rbac::v1::{ClusterRole, ClusterRoleBinding, Role, RoleBinding};
 use kube::api::{Api, ListParams, LogParams, PostParams, WatchEvent};
-use kube::{Client, ResourceExt};
+use kube::Client;
 use tracing::{debug, info};
 
 use test_log::test;
@@ -28,7 +28,7 @@ async fn print_pod_logs(client: Client, namespace: &str, label: &str) {
             async move {
                 let mut logs = pods
                     .log_stream(
-                        &p.name(),
+                        &p.metadata.name.clone().unwrap(),
                         &LogParams {
                             follow: true,
                             tail_lines: None,
@@ -39,11 +39,11 @@ async fn print_pod_logs(client: Client, namespace: &str, label: &str) {
                     .unwrap()
                     .boxed();
 
-                debug!("Logging agent pod {}", p.name());
+                debug!("Logging agent pod {:?}", p.metadata.name);
                 while let Some(line) = logs.next().await {
                     debug!(
                         "LOG [{:?}] {:?}",
-                        p.name(),
+                        p.metadata.name,
                         String::from_utf8_lossy(&line.unwrap())
                     );
                 }
--- a/common/k8s/src/lib.rs
+++ b/common/k8s/src/lib.rs
@@ -27,7 +27,6 @@ fn create_k8s_client(
     user_agent: hyper::http::header::HeaderValue,
     config: Config,
 ) -> Result<Client, kube::Error> {
-    let timeout = config.timeout;
     let default_ns = config.default_namespace.clone();
 
     let client: hyper::Client<_, Body> = {
@@ -40,8 +39,8 @@ fn create_k8s_client(
         ));
 
         let mut connector = TimeoutConnector::new(connector);
-        connector.set_connect_timeout(timeout);
-        connector.set_read_timeout(timeout);
+        connector.set_connect_timeout(config.connect_timeout);
+        connector.set_read_timeout(config.read_timeout);
 
         hyper::Client::builder().build(connector)
     };
@@ -76,7 +75,8 @@ pub fn create_k8s_client_default_from_env(
         return Err(K8sError::K8sNotInClusterError());
     }
 
-    let config = Config::from_cluster_env()?;
+    let config = Config::incluster_dns()?;
+
     Ok(create_k8s_client(user_agent, config)?)
 }
 
--- a/common/k8s/src/middleware/metadata.rs
+++ b/common/k8s/src/middleware/metadata.rs
@@ -1,6 +1,7 @@
 use crate::errors::K8sError;
 use crate::middleware::parse_container_path;
 use futures::StreamExt;
+use futures::TryStreamExt;
 use http::types::body::{KeyValueMap, LineBufferMut};
 use k8s_openapi::api::core::v1::Pod;
 use kube::{
@@ -11,18 +12,17 @@ use kube::{
         utils::StreamBackoff,
         watcher::{watcher, Event as WatcherEvent},
     },
-    Api, Client,
+    Api, Client, ResourceExt,
 };
 use serde_json::json;
-use std::fmt;
-use std::pin::Pin;
+use std::{fmt, sync::Mutex};
 
 use backoff::ExponentialBackoff;
 use metrics::Metrics;
 use middleware::{Middleware, Status};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
-use tracing::{trace, warn};
+use tracing::{error, trace, warn};
 
 #[derive(Error, Debug)]
 pub enum Error {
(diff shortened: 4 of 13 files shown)


bin/src/_main.rs
```diff
@@ -162,22 +163,48 @@ pub async fn _main(
...
-                        }
-                        Err(e) => {
-                            let message = format!(
-                                "The agent could not access k8s api after several attempts: {}",
-                                e
-                            );
-                            error!("{}", message);
-                            panic!("{}", message);
+                    let k8s_client_clone = k8s_client.clone();
+                    tokio::spawn(async move {
```
dkhokhlov, line 171: client is reused? connection is not recreated?
and we reuse the storage?

bin/src/_main.rs
```diff
@@ -162,22 +163,48 @@ pub async fn _main(
...
+                                    let message = format!(
+                                    "The agent could not access k8s api after several attempts: {}",
+                                    e
+                                );
+                                    error!("{}", message);
+                                    panic!("{}", message);
+                                }
+                            };
+
+                            tokio::time::sleep(tokio::time::Duration::from_secs(10 * 60)).await;
```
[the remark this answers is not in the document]
jakedipity: Yeah I can move this to a constant. I picked 10 minutes arbitrarily, let me know if you think we should change it.

bin/src/_main.rs
jakedipity: Ah I didn't think about the client, it makes sense to recreate it every time in case our issue is arising there. The reflector store is dropped with the thread each refresh; so a new store and watcher is created each time.

bin/src/_main.rs
```diff
@@ -51,6 +52,7 @@ use crate::stream_adapter::{StrictOrLazyLineBuilder, StrictOrLazyLines};
 pub static REPORTER_LEASE_NAME: &str = "logdna-agent-reporter-lease";
 pub static K8S_EVENTS_LEASE_NAME: &str = "logdna-agent-k8-events-lease";
 pub static DEFAULT_CHECK_FOR_LEADER_S: i32 = 300;
+pub static DEFAULT_K8S_WATCHER_REFRESH: Duration = Duration::from_secs(10 * 60);
```
dkhokhlov, line 55: DEFAULT_K8S_WATCHER_REFRESH_S
?

bin/src/_main.rs
dkhokhlov: I assume while we restart watcher we will not lose notifications. could you check API spec that we do get some backlog of notifications after new connect?

bin/src/_main.rs
jakedipity: Renamed to hopefully be clearer.

jakedipity:
I moved the `Mutex` to only surround the reflector inside of `K8sMetadata` instead of around the entire middleware itself. This simplifies things some and helps limit the critical section.

The `kick_over` function now returns a new `StoreSwapIntent` instead of the reflector store itself; this is to make it clear what the object is supposed to be used for.

I added a `VecDeque` to track more than one running  watcher thread. This collection will only ever have at most 2 watcher threads and allows us to overlap the running threads, so a new watcher can populate it's store before being utilized by processing lines in the middleware.

jakedipity:
@dkhokhlov Should be ready for re-review

Also @c-nixon would be good to have some extra eyes on this

jakedipity:
@c-nixon Bumped kube to latest, `0.80.0`. Needed to update our minimum kubernetes version to `v1.20`; however, and needed to make some small tweaks because of some api changes.

bin/src/_main.rs
```diff
@@ -139,110 +142,170 @@ pub async fn _main(
...
-                        .await;
+                        let (stream_handle, abort_registration) = AbortHandle::new_pair();
+                        let (thread_handle, swap_intent) = match middleware
+                            .kick_over(middleware_k8s_client.unwrap(), node_name.as_deref())
+                        {
+                            Ok((stream, swap_intent)) => {
+                                debug!("k8s metadata watcher kicked over!");
+                                let abortable_stream = Abortable::new(stream, abort_registration);
+                                (
+                                    tokio::spawn(async move {
```
dkhokhlov, line 186: I do not feel good about this additional spawn inside loop.  how about using a channel instead of abortable stream?

bin/src/_main.rs
```diff
@@ -139,110 +142,170 @@ pub async fn _main(
...
-                    K8sTrackingConf::Always => {
-                        let created_metrics_stream_feature_meta = set_up_leader(
-                            &namespace,
-                            &k8s_client,
-                            &pod_name,
-                            REPORTER_LEASE_NAME.to_string(),
-                        )
-                        .await;
+                        let (stream_handle, abort_registration) = AbortHandle::new_pair();
+                        let (thread_handle, swap_intent) = match middleware
```
dkhokhlov, line 179: first loop run starts with kick over right away?

bin/src/_main.rs
```diff
@@ -139,110 +142,170 @@ pub async fn _main(
...
+                                    "The agent could not access k8s api after several attempts: {}",
+                                    e
+                                );
+                                error!("{}", message);
+                                panic!("{}", message);
+                            }
+                        };
+
+                        handle_queue.push_back((thread_handle, stream_handle, Some(swap_intent)));
+                        if handle_queue.len() == 2 {
```
dkhokhlov, line 205: hard coded 2. is it when queue is full?

bin/src/_main.rs
```diff
@@ -139,110 +142,170 @@ pub async fn _main(
...
+                            debug!("delaying k8s metadata watcher rotation so new store can populate...");
+                            // when rotating watchers, introduce a small delay so the new
+                            // watcher has a chance to populate its cache
+                            sleep(K8S_WATCHER_ROTATION_DELAY).await;
```
dkhokhlov, line 209: how about to use a channel here? for explicit handshake.

bin/tests/it/k8s.rs
```diff
@@ -28,7 +28,7 @@ async fn print_pod_logs(client: Client, namespace: &str, label: &str) {
             async move {
                 let mut logs = pods
                     .log_stream(
-                        &p.name(),
+                        &p.metadata.name.clone().unwrap(),
```
dkhokhlov, line 31: always Some()?

common/k8s/src/lib.rs
```diff
@@ -76,7 +75,8 @@ pub fn create_k8s_client_default_from_env(
         return Err(K8sError::K8sNotInClusterError());
     }
 
-    let config = Config::from_cluster_env()?;
+    let config = Config::incluster_dns()?;
```
dkhokhlov, line 78: why renamed to dns?

common/k8s/src/middleware/metadata.rs
```diff
@@ -82,56 +105,51 @@ impl K8sMetadata {
...
-                    .await
-            }
-        };
-
-        Ok((Box::pin(driver), K8sMetadata { store }))
+                    }
+                    Err(e) => warn!("k8s watch stream error: {}", e),
+                })
+                .take_while(|x| std::future::ready(x.is_ok()))
+                .map(Result::unwrap),
```
dkhokhlov, line 125: safe unwrap?

common/k8s/src/middleware/metadata.rs
```diff
@@ -148,38 +166,40 @@ impl Middleware for K8sMetadata {
...
-                    if let Some(ref annotations) = pod.metadata.annotations {
-                        if line
-                            .set_annotations(
-                                annotations
-                                    .iter()
-                                    .fold(KeyValueMap::new(), |acc, (k, v)| acc.add(k, v)),
-                            )
-                            .is_err()
-                        {
+                if let Some(ref store) = *self.store.lock().unwrap() {
```
dkhokhlov, line 169: always success unwrap?

common/k8s/src/middleware/runner.rs
```diff
@@ -0,0 +1,111 @@
...
+    match state {
+        State::Init => State::Creating(true),
+        State::Creating(first) => {
+            let middleware_k8s_client = create_k8s_client_default_from_env(user_agent.clone());
+            if let Err(e) = middleware_k8s_client {
+                let message = format!("unable to create client for k8s metadata watcher: {:?}", e);
+                error!("{}", message);
+                panic!("{}", message);
+            }
+            State::Kicking(first, middleware_k8s_client.unwrap())
```
dkhokhlov, line 40: middleware_k8s_client always Some()? Err or Option?
may be match...

common/middleware/src/lib.rs
```diff
@@ -17,6 +17,41 @@ pub trait Middleware: Send + Sync + 'static {
     fn process<'a>(&self, lines: &'a mut dyn LineBufferMut) -> Status<&'a mut dyn LineBufferMut>;
 }
 
+pub enum MixedMiddleware<T: Middleware> {
```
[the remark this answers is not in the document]
jakedipity: I implemented this suggestion.

common/k8s/src/lib.rs
jakedipity: Because we are using rustls-tls, we have to use the dns instead of from the env: https://docs.rs/kube/latest/kube/struct.Config.html#method.incluster_dns

common/k8s/src/middleware/metadata.rs
jakedipity: Yes, `take_while` will terminate the stream if an error is encountered, so the unwrap will never get an error.

bin/src/_main.rs
jakedipity: outdated

bin/src/_main.rs
jakedipity: outdated

bin/src/_main.rs
jakedipity: outdated - the new code now uses a state machine of sorts

bin/src/_main.rs
jakedipity: outdated

common/k8s/src/middleware/metadata.rs
jakedipity: Unwrap could panic if a previous holder of the lock panicked while holding the lock. There's not much else we can do in this situation, so panicking seems like an acceptable result.

common/k8s/src/middleware/runner.rs
jakedipity: The error case is handled in the lines before (it'll panic), so the unwrap should always have a `Some`. I simplified the code to process the error at the same time it unwraps it.

bin/tests/it/k8s.rs
jakedipity: Not sure, with the update to our rust dependencies, these values are now returned as an Option. IMO given that it's a test I think it's okay to just unwrap without trying to handle the error - panicking will be seen as a test failure.

common/k8s/src/middleware/runner.rs
```diff
@@ -0,0 +1,111 @@
...
+
+enum State<'a> {
+    Init,
+    Creating(bool),
+    Kicking(bool, kube::Client),
+    Swapping(bool, JoinHandle<()>, StoreSwapIntent<'a>),
+    Running(JoinHandle<()>),
+}
+
+async fn step_trampoline<'a>(
```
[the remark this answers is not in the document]
jakedipity: I added exponential backoff, it will pause after the Running state. It starts at 2 seconds and goes to a maximum of 30 seconds. If the running state is able to run successfully for longer than 30 seconds, the attempts counter will get reset and go back to a 2 second delay.

PR #493, opened 2023-04-13
Reduce k8s metadata memory use

c-nixon:
This reduces the k8s metadata watcher's memory footprint by only tracking the pod metadata rather than the entire pod spec

### DIFF
--- a/common/k8s/src/event_source.rs
+++ b/common/k8s/src/event_source.rs
@@ -12,8 +12,8 @@ use k8s_openapi::apimachinery::pkg::apis::meta::v1::Time;
 use kube::api::ListParams;
 use kube::{
     core::PartialObjectMeta,
-    runtime::{watcher, metadata_watcher, WatchStreamExt},
-    Api, Client,ResourceExt,
+    runtime::{metadata_watcher, watcher, WatchStreamExt},
+    Api, Client, ResourceExt,
 };
 use metrics::Metrics;
 use pin_utils::pin_mut;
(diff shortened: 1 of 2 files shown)


common/k8s/src/middleware/metadata.rs
```diff
@@ -103,7 +104,32 @@ impl K8sMetadata {
...
+                                .collect()
+                        ,
+                        ..Default::default()
+                    };
+                    *pod_spec = filtered_pod_spec;
+                }
+            })
+        });
+
+        let watched = StreamBackoff::new(watcher, ExponentialBackoff::default());
```
dkhokhlov, line 132: @jakedipity 

This StreamBackoff may be our suspect in the missing k8s meta issue. The stream processing has silent errors (no logging).

> After any Err is emitted, the stream is paused for Backoff::next_backoff. The Backoff is reset on any Ok value. If Backoff::next_backoff returns None then the backing stream is given up on, and closed.

dkhokhlov (approved):
lgtm

common/k8s/src/middleware/metadata.rs
```diff
@@ -231,10 +231,31 @@ impl K8sMetadata {
             ListParams::default()
         };
 
-        let watcher = Box::pin(StreamBackoff::new(
-            watcher(api, params),
-            ExponentialBackoff::default(),
-        ));
+        let watcher = watcher(api, params).map_ok(|ev| {
+            ev.modify(|pod| {
+                pod.managed_fields_mut().clear();
```
jakedipity, line 236: Clearing managed_fields happens twice, once here and again on L384. You can remove the `map_ok` on L384

jakedipity (changes requested):
Just one instance of duplicate code

jakedipity (approved):

PR #495, opened 2023-04-13
fix: delay k8s metadata cleanup until file is deleted

c-nixon:
In some cases, logs ingested closed to when a pod is deleted might cause
the logs to not be able to collect metadata information since the info
has been removed from our cache. This fix makes sure that metadata info
lives as long as the file exists to prevent this from happening.

### DIFF
--- a/common/k8s/src/middleware/metadata.rs
+++ b/common/k8s/src/middleware/metadata.rs
@@ -1,7 +1,19 @@
 use crate::errors::K8sError;
-use crate::middleware::parse_container_path;
-use futures::StreamExt;
-use futures::TryStreamExt;
+use crate::middleware::{parse_container_path, ParseResult};
+
+use std::{
+    collections::HashMap,
+    fmt,
+    ops::DerefMut,
+    sync::{Arc, Mutex},
+};
+
+use arc_interner::ArcIntern;
+use async_channel::Receiver;
+use futures::{
+    stream::{self, select},
+    StreamExt, TryStreamExt,
+};
 use http::types::body::{KeyValueMap, LineBufferMut};
 use k8s_openapi::api::core::v1::Pod;
 use kube::{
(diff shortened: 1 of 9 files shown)


common/k8s/src/middleware/metadata.rs
```diff
@@ -103,7 +228,107 @@ impl K8sMetadata {
...
+        let (deletion_ack_recv, deletion_state) = {
+            let state = self.state.lock().map_err(|_| Error::K8sMiddlewareState)?;
+            // Take the deletion ack stream, copy the deletion state stream
+            (self.deletion_ack_recv.clone(), state.deletion_state.clone())
+        };
+
+        let watched = stream::unfold(
+            (watcher, deletion_state, deletion_ack_recv),
+            move |(mut watcher, deletion_state, mut deletion_ack_recv)| async move {
+                // Check if we have a delete acknowledgement
```
jakedipity, line 245: Trying to accomplish both deletion ack handling and watcher event intercepting in the same stream will cause them to be in a weird lockstep e.g. if the stream blocks down below waiting for a watcher event, no delete events can be processed.

Could we instead turn these into two separate streams and then merge them together? That should be easier to understand and work quite a bit better.

Finally, in my own testing it didn't seem like this section of code was ever being run; I didn't see `handle_pod` called for any delete events despite the log file being deleted.

dkhokhlov:
How about to add a deleted node flag with timestamp? And delete node entry later on schedule.

jakedipity:
I made some changes to get this into working order:
1. Replaced the multi sender single receiver channel with an async compatible drop in that allows multiple receivers. This is required because the receiver is consumed each time the middleware is kicked over. Before it was using an `Arc` around the single receive to support cloning, and mutability through `Arc::get_mut`. This unfortunately never works because there's multiple references.
2. Break up the stream going into the reflector into multiple streams which are combined for a final stream.
3. Fix the output of the `stream::unfold` function to not return a None (because that causes the stream to end).

jakedipity:
I was able to get the filesystem to send deletion acks by instead triggering them when the watcher unregisters a path. This works around the still existing racing condition between the Tailer and FileSystem trying to delete internal state.

PR #496, opened 2023-04-14
fix: don't duplicate on tailer restart

jakedipity:
The agent reuses the same initial offsets created on agent startup each time the tailer is restarted. This causes the agent to send duplicate logs each time the tailer restarts. This PR fixes that issue by generating the offsets each time the tailer restarts.

Ref: LOG-16194


bin/src/_main.rs
```diff
@@ -390,12 +381,23 @@ pub async fn _main(
...
+                    Err(e) => warn!("couldn't retrieve offsets from agent state, {:?}", e),
+                }
+            }
+
             let tailer = tail::Tailer::new(
                 watched_dirs.clone(),
                 rules.clone(),
                 lookback.clone(),
-                offsets.clone(),
+                initial_offsets,
```
dkhokhlov, line 400: @jakedipity 
does this mean that instead of reusing existing in-memory state we are querying fresh "initial_offsets" from database now?
Is it possible that some offset in DB may get updated after new Tailer created?

jakedipity: From my understanding of how the restarting tailer stream works, it only polls the current active tailer. So if the tailer is replaced, no more events are polled from the old tailer.

I think there's still a chance lines could be somewhere in the middleware or client or retries, but that seems like a very minimal chance.

Initial offsets were never pulled from memory. They've always been calculated from the DB, but were only being done once on initial start.

PR #502, opened 2023-05-11
Bump Rocksdb and tweak state memory usage

c-nixon:
This bumps rocksdb to a more recent version which fixes compilation with newer versions of GCC

I also took the opportunity to bump the SpanVec's internal SmallVec stack allocated size to 12, this eliminates ~10% of the agent's allocator activity while keeping the default SpanVec less than a cache line, so there shouldn't be much if any appreciable memory pressure impact.


dkhokhlov (approved):
lgtm

PR #509, opened 2023-05-31
fix: clear pending batches when processed

c-nixon:
Looks like we aren't currently popping batches off the pending queue in the state tracking module

Ref: LOG-14646


dkhokhlov (commented):
lgtm

dkhokhlov (approved):
lgtm

PR #510, opened 2023-06-07
chore: add better fs error messages to fs::cache::try_from

jakedipity:
We have noticed some mystic error messages from the results of try_from which don't tell us where the issue is happening. This PR uses a drop in replacement of std::fs to extract better error messages that include the path and action

Ref: LOG-14646

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1076,6 +1076,7 @@ dependencies = [
  "mz-http",
  "notify_stream",
  "os_str_bytes",
+ "path_abs",
  "pcre2",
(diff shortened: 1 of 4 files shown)


dkhokhlov:
would be nice to drop that into tailed_file.rs:
https://github.com/logdna/logdna-agent-v2/blob/df491cd8cca74d98cc6d8ca7ce25d559073c602c/common/fs/src/cache/tailed_file.rs#L702

dkhokhlov (approved):
lgtm

PR #513, opened 2023-06-19
Updated hyper-rustls to native-tls

SivakaviKavitha:
To avoid the ring dependency updated hyper-rustls to native-tls

Fixes: #521


dkhokhlov:
@SivakaviKavitha 
I tried to build your PR with
```
make test
```
and got error:
```
error: failed to select a version for the requirement `native-tls = "^0.23"` (locked to 0.23.2)
candidate versions found which didn't match: 0.2.11, 0.2.10, 0.2.9, ...
``` 
Do you see the same?

SivakaviKavitha:
Hi @dkhokhlov ,
yeah, i am also facing the same issue.

Could you please suggest us on how to proceed ?

PR #516, opened 2023-06-20
fix: delay k8s metadata cleanup until file is deleted

jakedipity:
In some cases, logs ingested closed to when a pod is deleted might cause the logs to not be able to collect metadata information since the info has been removed from our cache. This fix makes sure that metadata info lives as long as the file exists to prevent this from happening.

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -274,6 +274,8 @@ pub struct FileSystem {
 
     fo_state_handles: Option<(FileOffsetWriteHandle, FileOffsetFlushHandle)>,
 
+    deletion_ack_sender: async_channel::Sender<Vec<PathBuf>>,
+
     _c: countme::Count<Self>,
 }
 
@@ -315,6 +317,7 @@ impl FileSystem {
         initial_offsets: HashMap<FileId, SpanVec>,
         rules: Rules,
         fo_state_handles: Option<(FileOffsetWriteHandle, FileOffsetFlushHandle)>,
+        deletion_ack_sender: async_channel::Sender<Vec<PathBuf>>,
     ) -> Self {
         let (resume_events_send, resume_events_recv) = async_channel::unbounded();
         let (retry_events_send, retry_events_recv) = async_channel::unbounded();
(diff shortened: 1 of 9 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -1153,6 +1157,7 @@ impl FileSystem {
         };
 
         let path = entry.path().to_path_buf();
+        let _ = self.deletion_ack_sender.try_send(vec![path.clone()]);
```
dkhokhlov, line 1172: can it fail?

common/k8s/src/middleware/metadata.rs
```diff
@@ -104,11 +231,130 @@ impl K8sMetadata {
...
+            let state = self.state.lock().map_err(|_| Error::K8sMiddlewareState)?;
+            // Take the deletion ack stream, copy the deletion state stream
+            (self.deletion_ack_recv.clone(), state.deletion_state.clone())
+        };
+
+        let delete_stream = stream::unfold(
+            (deletion_state.clone(), deletion_ack_recv),
+            move |(deletion_state, deletion_ack_recv)| async move {
+                let next_event = loop {
+                    if let Ok(Some((deleted_container, deleted_container_pod))) = deletion_ack_recv
```
dkhokhlov, line 249: Err possible?

common/fs/src/cache/mod.rs
jakedipity: Since the channel is unbounded, it will only fail if the receiver is dropped. The only time that might happen is if the agent is in process of exiting. I added a warn log to catch any errors just to be safe.

common/k8s/src/middleware/metadata.rs
jakedipity: Same comment as above. An error condition is unlikely. I added so warning logs in case an error does occur.

common/fs/src/cache/mod.rs
```diff
@@ -1165,6 +1169,9 @@ impl FileSystem {
         };
 
         let path = entry.path().to_path_buf();
+        if let Err(e) = self.deletion_ack_sender.try_send(vec![path.clone()]) {
```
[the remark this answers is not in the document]
jakedipity: Moved. Because the path is referenced in the error messaging, I can't remove the clone.

PR #518, opened 2023-06-21
refactor: features allowing compiling sans rocksdb

c-nixon:
This adds features allowing us to compile interdependent common modules sans rocksdb to make their builds quicker

There should be no functionality changes for this refactor


dkhokhlov (approved):
no functional changes, just extract refactoring?

lgtm

PR #520, opened 2023-06-28
fix: allow k8s deletion stream to exit on closed delete channel

jakedipity:
Before the k8s deletion stream would continuously poll the delete channel after it had closed. This change reworks the logic flow so that the deletion stream will instead close in this case given that it is non-recoverable.

Ref: LOG-17430


stsantilena (approved):
This LGTM.

PR #525, opened 2023-07-31
Bump rocks db to support newer compilers

c-nixon:
This bumps rocksdb to a more recent version which fixes compilation with newer versions of GCC


dkhokhlov (approved):
lgtm

PR #527, opened 2023-08-01
chore: remove rust-toolchain.toml now that clippy is fixed

c-nixon:
This removes the rust-toolchain.toml as the bug in clippy that it was working around has been fixed.

I'd prefer to leave it and change channel to "stable", but having the additional targets (which are required for it to work at all) also means a lot of unnecessary installations happen inside docker build steps even though the targets aren't necessary for that step


dkhokhlov (approved):
lgtm

PR #529, opened 2023-08-03
Coalesce fs events

c-nixon:
Add an event coalescer/debouncer for the inotify event stream. This serves a couple of purposes.

One: It coalesces long streams of redundant writes into a single write event, which the TailedFile object already handles correctly (by trying to read the whole way from it's current location to the end of the file when it gets a write event).

Two: It ensures that the notify stream is kept clear where possible, rather than having a limited, fixed, buffer of events  that can overflow if it backs up the debouncer maintains a compressed history of all of the events it has seen at a particular file path.

### DIFF
--- a/bin/tests/it/k8s.rs
+++ b/bin/tests/it/k8s.rs
@@ -8,6 +8,7 @@ use k8s_openapi::api::core::v1::{Endpoints, Namespace, Pod, Service, ServiceAcco
 use k8s_openapi::api::rbac::v1::{ClusterRole, ClusterRoleBinding, Role, RoleBinding};
 use kube::api::{Api, ListParams, LogParams, PostParams, WatchEvent};
 use kube::Client;
+
 use test_log::test;
 use tokio::io::AsyncWriteExt;
 use tokio::net::TcpStream;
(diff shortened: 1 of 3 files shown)


common/fs/src/cache/event_debouncer.rs
```diff
@@ -0,0 +1,407 @@
...
+        Some(ret)
+    }
+}
+
+#[derive(Debug)]
+struct PrioritisedEvents<T> {
+    key: std::path::PathBuf, // What filepath is this talking about
+    // Coalescing FIFO queue of events. end is head, start is tail
+    pub events: smallvec::SmallVec<[T; 4]>, // Debounced events
+}
```
dkhokhlov, line 98: does it have any delay to keep some backlog of events ready for coalescing?

dkhokhlov: > full history, in case we get a long repeating stream of "Create, Write, Delete"" 

how is this history synchronized with actual file state/content on the disk?

dkhokhlov (approved):
lgtm

jakedipity (approved):

PR #530, opened 2023-08-04
Harden k8s metadata middleware

c-nixon:
This updates the kube-rs dependency and configures our watchers to be more resilient and hopefully have lower impact on the api server.

### DIFF
--- a/common/k8s/src/middleware/metadata.rs
+++ b/common/k8s/src/middleware/metadata.rs
@@ -1,5 +1,6 @@
 use crate::errors::K8sError;
 use crate::middleware::{parse_container_path, ParseResult};
+use crate::K8S_WATCHER_TIMEOUT;
 use middleware::MiddlewareError;
 use rate_limit_macro::rate_limit;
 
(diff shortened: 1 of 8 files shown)


common/k8s/src/middleware/metadata.rs
```diff
@@ -225,13 +225,15 @@ impl K8sMetadata {
         let store_writer = reflector::store::Writer::default();
         let store = store_writer.as_reader();
 
-        let params = if let Some(node) = node_name {
-            ListParams::default().fields(&format!("spec.nodeName={}", node))
+        let wc = WatcherConfig::default().timeout(30).any_semantic();
```
jakedipity, line 228: Could this potentially be a fix for the issue we've seen where in some cases the watcher becomes un-responsive after the API returns an error? https://github.com/logdna/logdna-agent-v2/pull/491

jakedipity: Right now our fix is a bit ugly, requiring us to recreate the watcher on errors

dkhokhlov: also 30 - make it configurable and/or name this default val as const?

dkhokhlov (approved):
lgtm

assuming any_semantic() will not cause inconsistent pod meta we care about.

jakedipity (approved):

PR #531, opened 2023-08-08
fix: "missing" initial log dir is not watched after create + delete + create

dkhokhlov:
- added FS Rescan (restart) when an initial log dir gets created or deleted
- fixed old bug when Error::Rscan as event processing result was ignored
- fixed case when folders on the same level with missing initial log dir are watched (shall not)

Ref: LOG-17339, LOG-17338


common/fs/src/cache/watch.rs
```diff
@@ -1,417 +0,0 @@
-use inotify::{EventMask, Inotify, WatchDescriptor, WatchMask};
```
jakedipity, line 1: Is this file just leftover dead code?

jakedipity (approved):
Approved, but had 1 question regarding the deletion of the watch.rs file

common/fs/src/cache/watch.rs
dkhokhlov: yes, left over after porting from inode to cross platform notify-rs.

PR #532, opened 2023-08-14
fix: accurately track offsets between filesystem restarts

jakedipity:
Adds a memory cache of offsets to accurately manage the current state of offsets between filesystem restarts. This cache is completely separate from the db-persisted cache used by the client.

Ref: LOG-17217

### DIFF
--- a/bin/src/_main.rs
+++ b/bin/src/_main.rs
@@ -33,8 +33,9 @@ use middleware::Executor;
 
 use pin_utils::pin_mut;
 use rand::Rng;
-use state::{AgentState, FileId, SpanVec};
+use state::{AgentState, FileId, GetOffset, Span, SpanVec};
 use std::collections::HashMap;
+use std::convert::TryFrom;
 use std::path::Path;
 use std::sync::Arc;
 use tokio::signal::*;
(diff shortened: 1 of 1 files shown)


bin/src/_main.rs
```diff
@@ -367,28 +383,42 @@ pub async fn _main(
...
+                        let mut pair = (None, None);
+                        if let Ok(line) = line {
+                            pair = (line.get_key(), line.get_offset());
+                        }
+
+                        let fs_offsets = fs_offsets.clone();
+                        async move {
+                            if let (Some(key), Some(offsets)) = pair {
+                                let mut span_vec = SpanVec::new();
+                                span_vec.insert(Span::try_from(offsets).unwrap());
```
dkhokhlov, line 415: this is the only place where we read offsets from DB?

jakedipity: Offsets from the DB are only used once on agent startup as seen in L364. All offsets thereafter are updated as lines come through.

dkhokhlov: fs_offsets.clone() gets updated?

jakedipity: `fs_offsets` gets updated from each line that passes through. Each time the tailer is restarted, the new tailer gets a clone of fs_offsets so it starts with the the most up-to-date offsets.

dkhokhlov: i mean - why do you need clone() in the first place?

dkhokhlov (approved):
lgtm

PR #544, opened 2023-09-11
feat: rate-limit-macro: optional parameter for fallback code block

dkhokhlov:
- added new optional parameter for fallback code block


dkhokhlov:
published at crates.io:
https://crates.io/crates/rate-limit-macro

PR #546, opened 2023-09-12
enable it test for delayed lines

dkhokhlov:
- enabled k8s_integration_test_for_dalyed_lines test


common/k8s/src/middleware/metadata.rs
```diff
@@ -489,7 +490,9 @@ impl Middleware for K8sMetadata {
     ) -> Result<&'a dyn LineBufferMut, MiddlewareError> {
         // here we retry only pod log lines that do not have k8s metadata
         let file_name = line.get_file().unwrap_or("");
-        debug!("validate line from file: '{:?}'", file_name);
+        rate_limit!(rate = 1, interval = 5, {
```
jakedipity, line 493: Do we need a rate limit for a debug log?

common/config/src/lib.rs
```diff
@@ -190,6 +190,18 @@ impl Config {
...
+            // Check if the key starts with "LOGDNA_" or "MZ_"
+            if key.starts_with("LOGDNA_") || key.starts_with("MZ_") {
+                // Print the key and its value
+                if key.contains("KEY") || key.contains("PIN") {
+                    info!("{}: [REDACTED]", key);
+                } else {
+                    info!("{}: {}", key, value);
+                }
+            }
+        }
```
jakedipity, line 203: Printing all the env vars might leak some secrets besides the ones we recognize, and the agent already prints out its config values. Is this just a left over from testing?

bin/tests/it/k8s.rs
```diff
@@ -1338,20 +1421,20 @@ async fn test_k8s_enrichment() {
         }
 
         // Wait for the data to be received by the mock ingester
-        tokio::time::sleep(tokio::time::Duration::from_millis(10_000)).await;
+        tokio::time::sleep(tokio::time::Duration::from_millis(3_000)).await;
```
jakedipity, line 1424: Could this lower time cause integration tests to be unreliable?

bin/tests/it/k8s.rs
```diff
@@ -1000,20 +1086,18 @@ fn get_agent_ds_yaml(
...
-                        existing_env_vars.push(serde_json::Value::Object(map));
-                    }
-                } else {
-                    panic!("unexpected");
-                }
+        let existing_env_vars = &mut cfgmap["spec"]["template"]["spec"]["containers"][0]["env"]
+            .as_array_mut()
+            .unwrap();
+        for env_var in env_vars {
+            if let Some((k, v)) = env_var.iter().next() {
```
jakedipity, line 1093: nit: I find the previous method of converting `env_var` into a map and then collecting it to be a bit more ergonomic, but by no means is this bad/wrong/blocker.

bin/tests/it/k8s.rs
```diff
@@ -200,6 +200,92 @@ async fn delete_service(client: Client, service_name: &str, namespace: &str) {
     }
 }
 
+async fn delete_daemonset(client: Client, daemonset_name: &str, namespace: &str) {
```
jakedipity, line 203: I think we can do a better job to ensure RAII with Rust's language features. Maybe we instead have the `create_agent_ds` function return a `AgentGuard` which is comparable to a `MutexGuard` which we get after locking a mutex - this resource can implement these delete operations with the drop trait

jakedipity (commented):
Mostly nit comments, but I had 1 big suggestion for ensuring RAII through a structured lifetime e.g. like a `MutexGuard`

common/config/src/lib.rs
dkhokhlov: Requested from IBM support.

dkhokhlov: I am going do it in separate PR. that will not use open list.

bin/tests/it/k8s.rs
dkhokhlov: test code. explicit easy to debug.

bin/tests/it/k8s.rs
dkhokhlov: no,, it is enough. legacy attempts to fight with flakiness of test. now we properly wait for pod state.

bin/tests/it/k8s.rs
dkhokhlov: same as before - this is test code, explicit order etc, easy to debug. the Drop trait's drop method isn't asynchronous. so have to spawn new task. and then deal with borrow/ownership rules of cleanup params. complexity is increasing and new code will need to be tested itself.

jakedipity (approved):

PR #555, opened 2023-09-21
Remove no-longer-necessary Arc<Mutex<T>>

c-nixon:
With rust 1.72 the old, no longer necessary, `Arc<Mutex<T>>` in fs is now causing a lint error. This PR replaces it with `RC<RefCell<T>>` where relevant. There are no functional changes.

This also fixes a couple of other lints, like using [T] instead of vec![] where possible and using try_fold where applicable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3001,16 +3001,15 @@ dependencies = [
 
 [[package]]
 name = "proptest"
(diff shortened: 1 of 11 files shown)


dkhokhlov (approved):
lgtm

PR #560, opened 2023-09-29
feat: feature flag for integration test hooks

dkhokhlov:
Ref: LOG_17820


dkhokhlov:
@c-nixon 
feature trick doe snot work - test is running image that does not have MOCK active.

PR #574, opened 2023-10-20
fix: respect filesystem offsets when lookback is disabled

jakedipity:
Normally offsets are ignored when lookback is disabled, but filesystem offsets should still be respected. Filesystem offsets track state across internal filesystem offsets and differ from global offsets which track state across application restarts.

Before this change global offsets were always loaded and the filesystem would optionally respect them based on the lookback configuration. Now the filesystem always respect offsets regardless of lookback configuration, and global offsets are only loaded if the current lookback configuration respects global state.

Ref: LOG-17217


dkhokhlov (approved):
stress tested - looks good.

PR #587, opened 2023-11-28
fix: handle truncation while assigning initial offsets

jakedipity:
Initial offsets never considered the current state of the file, instead just preferring to use the state saved offsets or defaulting to the value configured by the `MZ_LOOKBACK` option. This means truncation is handled downstream where the agent doesn't have enough contextual information to appropriately handle the case and potentially dropping logs.

This change causes initial offsets to be ignored if they exceed the current length of the file, and instead use the default values.

Ref: LOG-17217

### DIFF
--- a/common/fs/src/cache/mod.rs
+++ b/common/fs/src/cache/mod.rs
@@ -867,11 +867,11 @@ impl FileSystem {
                 None
             }
         }
+        let file_len = path.metadata().map(|m| m.len()).unwrap_or(0);
         let default_offset = match self.lookback_config {
             Lookback::Start => SpanVec::new(),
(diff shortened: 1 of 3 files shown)


common/fs/src/cache/mod.rs
```diff
@@ -867,11 +867,11 @@ impl FileSystem {
                 None
             }
         }
+        let file_len = path.metadata().map(|m| m.len()).unwrap_or(0);
         let default_offset = match self.lookback_config {
```
dkhokhlov, line 871: should we call it lookback_offset?

dkhokhlov (approved):
lgtm!

PR #592, opened 2023-12-12
chore: update kube dependency to 0.87

jakedipity:
Version 0.86 of `kube` has been yanked, so updating to the current latest.


dkhokhlov (approved):
lgtm

PR #602, opened 2024-03-19
doc: giving better instructions for helm

gjanco:
showing how to set extraEnv when using helm to manage Agent

Ref: LOG-11023


jmoses (approved):
lgtm

PR #603, opened 2024-03-28
feat(config): Make ingestion flush duration configurable

dhable:
Adds a configuration param to control the duration of time to wait before sending captured logs to the ingestion API. Changes the default value from 250 ms to 5000 ms.

Ref: LOG-19388


dhable:
The check failures seem to be from some odd timing on the journald tests
```
[2024-03-29T14:42:44.257Z]         FAIL [   0.738s] journald journalctl::test::stream_gets_new_logs
[2024-03-29T14:42:44.257Z]         FAIL [   0.774s] journald libjournald::stream::tests::stream_gets_new_logs
```

bin/src/_main.rs
```diff
@@ -765,8 +765,7 @@ pub async fn _main(
             Metrics::middleware().increment_lines_egress();
             l
         })
-        // TODO: parameterize the flush frequency
```
jakedipity, line 768: nice

jakedipity (approved):

aholmberg (approved):
🥇 
Now if we can just get users to upgrade... :)

PR #621, opened 2024-06-05
fix: trim hostname whitespace

jakedipity:
Trims whitespace from the read contents of the hostname. Adds additional warnings for reading errors instead of failing silently.

Ref: LOG-19260


jakedipity:
@c-nixon Incorporated your suggestion and modified the test code to fit with the new argument type

PR #635, opened 2024-11-14
fix(fs): ensure dangling symlinks are monitored correctly

jakedipity:
Track the targets of dangling symlinks for a watched directory to make sure it's not unwatched if it runs out of children. A couple of changed were made to achieve this:

1. Move `Symlinks` type to it's own struct and add helper functions.
2. Add new `dependencies` field to dir entries to track symlinks that target a child entry regardless if whether the symlink is broken or not.
3. When inserting paths from within the `insert` and `insert_helper` functions, don't run an unncessary rules check.

Additionally, the following changes and additions were made:

1. Split `Entry` enum into `Entry` struct and `EntryKind` enum.
2. Add new `GuardedOption<T>` to enable a toggleable option; used by symlink entries for their target to differentiate dangling symlinks without losing the underlying path.
3. Use `target` variable name more consistently in regards to symlinks.
4. Remove directory walking during init; inserting the main logging directory does the same thing and this causes extra warnings about watching the same files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,7 +225,7 @@ integration-test: ## Run integration tests using image with additional tools
 .PHONY:k8s-test
 k8s-test: build-image-debian ## Run integration tests using k8s kind
 	$(DOCKER) tag $(REPO):$(IMAGE_TAG) $(REPO):local
-	IMAGE_TAG=$(IMAGE_TAG) $(DOCKER_KIND_DISPATCH) $(K8S_TEST_CREATE_CLUSTER) $(RUST_IMAGE) "--env RUST_LOG=$(RUST_LOG)" "cargo nextest run --no-fail-fast --nocapture $(TARGET_DOCKER_ARG) --manifest-path bin/Cargo.toml --features k8s_tests $(TESTS)"
+	IMAGE_TAG=$(IMAGE_TAG) $(DOCKER_KIND_DISPATCH) $(K8S_TEST_CREATE_CLUSTER) $(RUST_IMAGE) "--env RUST_LOG=$(RUST_LOG)" "cargo nextest run --no-fail-fast --retries=2 --nocapture $(TARGET_DOCKER_ARG) --manifest-path bin/Cargo.toml --features k8s_tests $(TESTS)"
 
 .PHONY:test-journald
 test-journald: ## Run journald unit tests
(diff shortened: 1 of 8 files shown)


bin/tests/it/cli.rs
```diff
@@ -961,7 +962,7 @@ fn lookback_tail_lines_file_created_after_agent_start_at_beg() {
         let (line_count, server) = tokio::join!(
             async {
                 let mut handle = common::spawn_agent(AgentSettings {
-                    log_dirs: &dir_path,
+                    log_dirs: &log_path,
```
jakedipity, line 965: Changed this so the tempdir does not get moved into this block. Previously this was causing the tempdir to be dropped immediately after we were done checking the agents log output. The agent would see it's logging directory delete, force a rescan, and panic because the logging directory isn't a directory.

This all happens before the lines get sent to the mock ingester which are checked afterwards.

Not sure why my changes is causing this problem to present itself now - maybe fs operations are faster now.

Makefile
```diff
@@ -225,7 +225,7 @@ integration-test: ## Run integration tests using image with additional tools
 .PHONY:k8s-test
 k8s-test: build-image-debian ## Run integration tests using k8s kind
 	$(DOCKER) tag $(REPO):$(IMAGE_TAG) $(REPO):local
-	IMAGE_TAG=$(IMAGE_TAG) $(DOCKER_KIND_DISPATCH) $(K8S_TEST_CREATE_CLUSTER) $(RUST_IMAGE) "--env RUST_LOG=$(RUST_LOG)" "cargo nextest run --no-fail-fast --nocapture $(TARGET_DOCKER_ARG) --manifest-path bin/Cargo.toml --features k8s_tests $(TESTS)"
+	IMAGE_TAG=$(IMAGE_TAG) $(DOCKER_KIND_DISPATCH) $(K8S_TEST_CREATE_CLUSTER) $(RUST_IMAGE) "--env RUST_LOG=$(RUST_LOG)" "cargo nextest run --no-fail-fast --retries=2 --nocapture $(TARGET_DOCKER_ARG) --manifest-path bin/Cargo.toml --features k8s_tests $(TESTS)"
```
jakedipity, line 228: Added more retries because k8s tests are likely to be flaky.

common/fs/src/cache/mod.rs
```diff
@@ -1414,17 +1452,8 @@ impl FileSystem {
...
                     Metrics::fs().decrement_tracked_files();
-                    if let Some(entries) = self.symlinks.get(path) {
-                        for entry in entries {
-                            if let Some(entry) = _entries.get(*entry) {
-                                if !self.passes(entry.path(), _entries) {
-                                    _links.push(entry.path().to_path_buf())
-                                }
-                            }
-                        }
-                    }
```
jakedipity, line 1427: Removed this logic that apparently wasn't doing anything

PR #640, opened 2024-12-09
chore: ignore RUSTSEC-2024-0421

jakedipity:
Ignore a vulnerability for one of our dependencies, `trust-dns-resolver` until a patch fix is available.

Also removes some outdated rustsec ignore rules which are no longer an issue.


gjanco (approved):

jakedipity:
Bypassing branch protections because Jenkins is borked and these changes does not have any code changes

PR #646, opened 2025-06-12
chore: upgrade to build-image and tooling to bookworm

jakedipity:
Upgrades our build-image and tooling to bookworm. Bookworm also uses a newer version of systemd which is more modular - this enables us to set up a journald testing image without all systemd requirements which was broken on some hosts.

Additionally refactors the docker dispatch scripts to not utilize detached containers and instead run the commands directly. Additionally drops -j from make test call in Jenkins to make output easier to parse.

Finally disables some tests which try and access to `/etc/logdna` which doesn't seem to work anymore in our Jenkins environment.

Ref: LOG-21788

### DIFF
--- a/common/config/src/lib.rs
+++ b/common/config/src/lib.rs
@@ -322,8 +322,7 @@ impl TryFrom<RawConfig> for Config {
             retry_dir: raw.http.retry_dir.unwrap_or_else(|| {
                 if cfg!(windows) {
                     if let Ok(temp) = std::env::var("temp") {
-                        let dir = format!("{}/{}", temp, "logdna");
-                        return PathBuf::from(dir);
+                        return Path::new(&temp).join("logdna");
                     }
                 }
                 PathBuf::from("/tmp/logdna")
@@ -465,7 +464,7 @@ pub fn get_hostname(alt: Option<&[&dyn AsRef<OsStr>]>) -> Option<String> {
             };
             match File::open(path).and_then(closure) {
                 Ok(s) => return Some(s.trim().to_string()),
-                Err(e) => warn!("Unable to read hostname from {:?}: {}", path, e),
+                Err(e) => warn!("Unable to read hostname from {path:?}: {e}"),
             }
         }
     }
--- a/common/test/mock-ingester/src/lib.rs
+++ b/common/test/mock-ingester/src/lib.rs
@@ -363,7 +363,7 @@ pub fn https_ingester_with_processors(
             let tcp = TcpListener::bind(&addr)
                 .await
                 .unwrap_or_else(|_| panic!("Couldn't bind to {:?}", addr));
-            info!("ingester listening at {:?}", addr);
+            info!("ingester listening at {addr:?}");
             let tls_acceptor = TlsAcceptor::from(tls_cfg);
             // Prepare a long-running future stream to accept and serve cients.
 
--- a/docker/dispatch.sh
+++ b/docker/dispatch.sh
@@ -2,40 +2,19 @@
 
 curpath=$(dirname "$0")
 
-# shellcheck disable=SC2317
-_term() {
-  docker kill "$child"
-  status=$(docker inspect "$child" --format='{{.State.ExitCode}}')
-  docker rm "$child"
-  exit "$status"
-}
-
 # shellcheck source=/dev/null
 . "$curpath/lib.sh"
 
 extra_args="$(get_volume_mounts "$1" "$3") $(get_sccache_args)"
 
-trap _term TERM
-trap _term INT
-
-start_sccache="if [ ! -z \${RUSTC_WRAPPER+x} ]; then while sccache --start-server > /dev/null 2>&1; do echo 'starting sccache server'; done; fi"
-
 if [ "$HOST_MACHINE" = "Mac" ]; then
-	# shellcheck disable=SC2086,SC2317
-	child=$(docker run -dit -w "$1" $extra_args -v "$2" $4 "$3" /bin/sh -ic "$5")
-	#docker run --rm -it -w "$1" $extra_args -v "$2" $4 "$3" /bin/bash -ic "$5"
+  # shellcheck disable=SC2086
+  docker run --rm --tty --workdir "$1" $extra_args --volume "$2" $4 "$3" /bin/sh -ic "$5"
 elif [ "$HOST_MACHINE" = "Linux" ]; then
-	# shellcheck disable=SC2086,SC2317,SC2154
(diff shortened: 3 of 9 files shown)


common/config/src/lib.rs
```diff
@@ -465,7 +465,7 @@ pub fn get_hostname(alt: Option<&[&dyn AsRef<OsStr>]>) -> Option<String> {
             };
             match File::open(path).and_then(closure) {
                 Ok(s) => return Some(s.trim().to_string()),
-                Err(e) => warn!("Unable to read hostname from {:?}: {}", path, e),
+                Err(e) => warn!("Unable to read hostname from {path:?}: {e}"),
```
kristof-mattei, line 468: This is personal, but I prefer to keep the variables outside of the message. I know there's a clippy lint that complains about this.

I disable the clippy lint:

```
[lints.clippy]
uninlined-format-args = { level = "allow", priority = 127 }
```

Why? `{foo:?}` works, but `{bar.quz:?}` does not work, so when you're dotting you get `warn!("something {foo:?} something else {:?}", bar.quz);` which is confusing. 

Ideally we move the parameters outside of the message, like so:

```
warn!(?err = e, path = %path.display(), "Unable to read hostname");
```

common/test/mock-ingester/src/lib.rs
```diff
@@ -363,7 +363,7 @@ pub fn https_ingester_with_processors(
             let tcp = TcpListener::bind(&addr)
                 .await
                 .unwrap_or_else(|_| panic!("Couldn't bind to {:?}", addr));
-            info!("ingester listening at {:?}", addr);
+            info!("ingester listening at {addr:?}");
```
kristof-mattei, line 366: As above

docker/journald/Dockerfile
```diff
@@ -4,27 +4,27 @@ FROM ${BUILD_IMAGE}
 
 ARG UID
 
-STOPSIGNAL SIGRTMIN+3
+RUN <<-EOF
```
kristof-mattei, line 7: First of all, applause of using the `<<-` marker, then we don't see leading spaces!

docker/journald/Dockerfile
```diff
@@ -4,27 +4,27 @@ FROM ${BUILD_IMAGE}
...
+  apt-get update
```
kristof-mattei, line 8: I would strongly recommend using `&&` here.

Without it the lines keep on executing even if previous ones fails. Example:

```
FROM ubuntu

RUN <<-EOF
    echo "Hello"
    cat /i-really-don-t-exist
    echo "World"
EOF

RUN echo "Done"
```

Output: 


```
❯ docker build . --progress plain --no-cache
#0 building with "desktop-linux" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 146B done
#1 DONE 0.0s

#2 [internal] load metadata for docker.io/library/ubuntu:latest
#2 DONE 0.2s

#3 [internal] load .dockerignore
#3 transferring context: 2B done
#3 DONE 0.0s

#4 [1/3] FROM docker.io/library/ubuntu:latest@sha256:7c06e91f61fa88c08cc74f7e1b7c69ae24910d745357e0dfe1d2c0322aaf20f9
#4 resolve docker.io/library/ubuntu:latest@sha256:7c06e91f61fa88c08cc74f7e1b7c69ae24910d745357e0dfe1d2c0322aaf20f9 done
#4 CACHED

#5 [2/3] RUN <<-EOF (echo "Hello"...)
#5 0.088 Hello
#5 0.088 cat: /i-really-don-t-exist: No such file or directory
#5 0.088 World
#5 DONE 0.1s

#6 [3/3] RUN echo "Done"
#6 0.109 Done
#6 DONE 0.1s

#7 exporting to image
#7 exporting layers 0.0s done
#7 exporting manifest sha256:8ff65f04719141e59b04655a826177ad923dc16029c9e1b3cb59cf3f00810380 done
#7 exporting config sha256:3bc6f43524e4e3ed0b547262d0b3af405243315a0c436497834e276695970a64 done
#7 exporting attestation manifest sha256:e988e50cd5222f8ed540a50f48baaca6088a12c6c819dcc28d9a525bc5e46b24 done
#7 exporting manifest list sha256:6cc01386fd2851af4dbd768915e1a6288c8d2eac0778f4d7673e170a31f9f47e done
#7 naming to moby-dangling@sha256:6cc01386fd2851af4dbd768915e1a6288c8d2eac0778f4d7673e170a31f9f47e done
#7 unpacking to moby-dangling@sha256:6cc01386fd2851af4dbd768915e1a6288c8d2eac0778f4d7673e170a31f9f47e done
#7 DONE 0.0s
```

Notice the 5th block where `World` is printed, which is unexpected. Same with the 6th block, and overall, we don't expect this build to succeed.

docker/dispatch.sh
```diff
@@ -2,40 +2,19 @@
...
-trap _term INT
-
-start_sccache="if [ ! -z \${RUSTC_WRAPPER+x} ]; then while sccache --start-server > /dev/null 2>&1; do echo 'starting sccache server'; done; fi"
-
 if [ "$HOST_MACHINE" = "Mac" ]; then
-	# shellcheck disable=SC2086,SC2317
-	child=$(docker run -dit -w "$1" $extra_args -v "$2" $4 "$3" /bin/sh -ic "$5")
-	#docker run --rm -it -w "$1" $extra_args -v "$2" $4 "$3" /bin/bash -ic "$5"
+  # shellcheck disable=SC2086
+  docker run --rm -t -w "$1" $extra_args -v "$2" $4 "$3" /bin/sh -ic "$5"
```
kristof-mattei, line 12: Nitpick: use long options in scripts: 

https://changelog.com/posts/use-long-flags-when-scripting

https://matklad.github.io/2025/03/21/use-long-options-in-scripts.html

docker/journald_dispatch.sh
```diff
@@ -3,54 +3,19 @@
...
 
 extra_args="$(get_volume_mounts "$1" "$image") $(get_sccache_args) $journald_args"
 
-# shellcheck disable=SC2086,SC2317
-child=$(docker run -d -w "$1" $extra_args -v "$2" $3 "$image")
-
 if [ "$HOST_MACHINE" = "Mac" ]; then
-	docker exec "$child" /bin/sh -c "$4"
+  # shellcheck disable=SC2086
+  docker run --rm -t -w "$1" $extra_args -v "$2" $3 "$image" /bin/sh -c "$4"
```
kristof-mattei, line 17: As above ^

docker/journald/Dockerfile
jakedipity: Ah yeah good catch. Fixed with `&&`!

common/config/src/lib.rs
kristof-mattei: Totally fair. Resolving.

PR #649, opened 2025-08-21
chore: update to ubi9

jakedipity:
Updates the base image from from ubi8 -> ubi9. Additionally tracks down remaining references to bullseye and replace with bookworm.

Ref: LOG-22220

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:1.0-experimental
 
-ARG UBI_MAJOR_VERSION=8
-ARG UBI_MINOR_VERSION=8
+ARG UBI_MAJOR_VERSION=9
+ARG UBI_MINOR_VERSION=6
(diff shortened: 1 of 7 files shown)


jmoses (approved):

Dockerfile
```diff
@@ -109,11 +109,12 @@ WORKDIR /work/
...
-    && microdnf clean all \
-    && rm -rf /var/cache/yum \
     && chmod -R 777 . \
     && setcap "cap_dac_read_search+p" /work/logdna-agent \
     && groupadd -g 5000 logdna \
-    && useradd -u 5000 -g logdna logdna
+    && useradd -u 5000 -g logdna logdna \
+    && microdnf remove shadow-utils -y \
+    && microdnf clean all \
+    && rm -rf /var/cache/yum
```
jakedipity, line 118: `shadow-utils` isn't required after `useadd` and `groupadd`, so save a little space here.

jakedipity:
@kristof-mattei Our build image isn't updated to use trixie - we were already partially over to bookworm so this just finishes it off. One thing at a time heh

